PR #336, opened 2016-05-04
Geo-distribution

kklin:
This PR adds most of the functionality needed for the geodistributed wordpress example.

There seems to be an issue with logging in as the wordpress admin, but that's not placement-related.


akshayknarayan:
re: ```network: Exclusively place containers that listen on the same port``` (commit 09b4cbf)[https://github.com/NetSys/quilt/commit/09b4cbf267f497bb5c12f102ad48ad0d09ed4d4d], it would be nice to have this at a higher level rather than as a special case. 
The problem is that given this structure, if you call dsl.QueryPlacements() the result is actually incorrect and you have to know about this and deal with it separately.

ejj:
I pushed a branch containing the code for this PR and made  a Trello card for figuring out what to do with it here: https://trello.com/c/e0g0p36O

Closing the PR for now.

PR #394, opened 2016-05-25
Hash Join with lonely lambdas

mmussomele:
Seems to be pretty clean, trades iteration after join for lambda declarations before it.

Known issues:
- store_test.go currently causes a panic because of duplicate nil keys being produced. Should we ignore nil keys when joining or is this a bug in store?


quilt.go
```diff
@@ -77,3 +88,27 @@ func updateConfig(conn db.Conn, configPath string) error {
 
 	return engine.UpdatePolicy(conn, spec)
 }
+
+// parseLogLevel returns the log.Level type corresponding to the given string (case insensitive).
```
ejj: This should be wrapped to 89 characters.

quilt.go
```diff
@@ -77,3 +88,27 @@ func updateConfig(conn db.Conn, configPath string) error {
...
+	case "warn":
+		return log.WarnLevel, nil
+	case "error":
+		return log.ErrorLevel, nil
+	case "fatal":
+		return log.FatalLevel, nil
+	case "panic":
+		return log.PanicLevel, nil
+	}
+	badLevel := "Invalid log level given. " +
```
ejj: Hmm, I think I commented on this last time around.  Are you sure this is the latest version of the commit?

ejj:
Just to make things are clean, let's repush a fresh commit

PR #397, opened 2016-05-25
minion: Add golang profiler support to the minion

nlsun:
There is currently only a placeholder method to invoking the profiler,
which is to compile the minion to run with the profiler.


ejj:
Merged

PR #399, opened 2016-05-25
Log levels, HashJoin and lonely lambdas

mmussomele:
Log levels is finished, HashJoin with lonely lambdas seems to be pretty clean, trades iteration after join for lambda declarations before it.

Known issues with HashJoin:
- store_test.go currently causes a panic because of duplicate nil keys being produced. Should we ignore nil keys when joining or is this a bug in store?


mmussomele:
Also, just let me know if I should squash the hashjoin commit's together

ejj:
So, the first patch introduces some trailing whitespace.  Should avoid that in future.  Btw, go format fixes that, make sure you're running make lint.

Anyways, I fixed it and merged.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
 	slice[l-1] = nil // Allow garbage collection.
 	return slice[:l-1]
 }
+
+// `ListLike` is an interface to let us get around using reflection, by simply requiring
```
ejj: I'd ditch the mention of reflection here.  Using something like this would be the most go-ish way of doing this so there's no need to justify not using it.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+// implementing types to allow access to their contained values by integer index.
+type ListLike interface {
```
ejj: I'd just call this `List`  cleaner.  Also importers will see it as `join.List`, with that added context it's a lot more clear.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+	Len() int
+	Get(int) interface{}
+}
+
+// The identity function
+var identity = func(val interface{}) interface{} {
```
ejj: Just define this within HashJoin.  Looks kinda weird out here.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+	return val
+}
+
+// Contains a hashtable entry and a flag for finding duplicates
+type joinTableEntry struct {
```
ejj: I might consider defining this within HashJoin too.  It's only meaningful within that function.  That said, I'm not totally sure.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+}
+
+// HashJoin attempts to match each element in `lSlice` with an element in `rSlice` by
+// performing a hash join. If such a match is found for a given element of `lSlice`,
+// it is returned as an element of `pairs`, while leftover elements from `lSlice` and
+// `rSlice` that couldn`t be matched are returned as elements of `lonelyLefts` and
+// `lonelyRights` respectively. The join keys for `lSlice` and `rSlice` are defined by
+// the passed in `lKey` and `rKey` functions, respectively.
+//
+// If `lKey` or `rKey` are nil, the identity is used instead.
```
ejj: `identity` is an implementation detail.  Just say "the keys are compared directly" or something in plain english.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+//
+// If `lKey` or `rKey` produce duplicate keys or produce unhashable keys, HashJoin 
+// will panic.
+func HashJoin(lSlice, rSlice ListLike, lKey, rKey func(interface{}) interface{}) (
+	pairs []Pair, lonelyLefts, lonelyRights []interface{}) {
+
+	// Use the identity if a nil function pointer is given
```
ejj: This comment adds no value, obvious from the code below.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+	if lKey == nil {
+		lKey = identity
+	}
+	if rKey == nil {
+		rKey = identity
+	}
+
+	// Initialize lonely left and rights set (need to be hash-sets because hash join
```
ejj: I find this comment confusing.

Why can't `llefts` and `lrights` be slices again?  Might be worth slacking me to explain before you make the comment clearer.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+		rKey = identity
+	}
+
+	// Initialize lonely left and rights set (need to be hash-sets because hash join 
+	// can't remove by index). Must map to interface{} instead of bool since the 
+	// elements of lSlice and rSlice might not be hashable, so the original elements
+	// are the values, allowing for reconstruction of the lonely sets. Use pointers
+	// to save some memory.
+	llefts, lrights := make(map[interface{}]*interface{}), make(map[interface{}]*interface{})
+	lLen, rLen := lSlice.Len(), rSlice.Len()
```
ejj: you only use lLen and rLen once, I'd just call them in their respective for loops.  Clearer.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+
+	// Initialize join table for hash join
```
ejj: This comment adds no information.  I'd ditch it.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+	joinTable := make(map[interface{}]*joinTableEntry)
+
+	// Build the join table and lonely lefts set using lSlice
+	for ii := 0; ii < lLen; ii++ {
+		lElem := lSlice.Get(ii)
+		lElemKey := lKey(lElem)
+
+		// check for duplicate keys produced by lKey on lSlice
```
ejj: Don't think this comment adds value either, pretty obvious what the iff statement is doing.

Also, it's idiomatic (I think it is at least), to use `ok` instead of `present`.  We've been doing it in this code base at least.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+		if _, present := joinTable[lElemKey]; present {
+			panic("duplicate keys produced by lKey function")
```
ejj: Could be slightly clearer.  Perhaps "HashJoin found duplicate keys".  The `lKey` reference is going to seem mysterious when looking at a stack trace.

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+			// check for duplicate keys produced by rKey on rSlice
+			if entry.matched {
+				panic("duplicate keys produced by rKey function")
+			} else {
+				entry.matched = true
+				pairs = append(pairs, Pair{*entry.val, rElem})
+				delete(llefts, rElemKey) // ok since rElemKey == lElemKey here
+			}
+		} else {
+			lrights[rElemKey] = &rElem
```
ejj: why do we even need `lrights`?  Couldn't we just append to the `lonelyRights` slice directly?

join/join.go
```diff
@@ -73,3 +73,99 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+// it is returned as an element of `pairs`, while leftover elements from `lSlice` and
+// `rSlice` that couldn`t be matched are returned as elements of `lonelyLefts` and
+// `lonelyRights` respectively. The join keys for `lSlice` and `rSlice` are defined by
+// the passed in `lKey` and `rKey` functions, respectively.
+//
+// If `lKey` or `rKey` are nil, the identity is used instead.
+//
+// If `lKey` or `rKey` produce duplicate keys or produce unhashable keys, HashJoin 
+// will panic.
+func HashJoin(lSlice, rSlice ListLike, lKey, rKey func(interface{}) interface{}) (
```
ejj: So reading this over now.  It seems way more complicated than it needs to be.  I think the reason for the added complexity is the requirement that we detect duplicate keys and panic.  If we ditched that requirement, we could likely ditch the joinTableEntry all together (wouldn't need the matched flag), plus make a host of over simplifications.

I know we discussed this before, but seeing the code now, I've changed my mind.  I think we should just deal with duplicate keys in whatever way is most natural.  Likely that means saying the behavior is "undefined", but I haven't thought about it carefully.

minion/engine.go
```diff
@@ -124,26 +120,42 @@ func updatePlacements(view db.Database, spec dsl.Dsl) {
 	}
 }
 
+func makeConnectionSlice(cm map[dsl.Connection]struct{}) dsl.ConnectionSlice {
```
ejj: Meh.  Trying too hard to make line 132 clean ;)

minion/network/worker.go
```diff
@@ -182,13 +194,11 @@ func updateNamespaces(containers []db.Container) {
...
 	}
 
-	_, lefts, rights := join.Join(currentNamespaces, targetNamespaces, func(
-		left, right interface{}) int {
-		if left.(nsInfo).ns == right.(nsInfo).ns {
-			return 0
-		}
-		return -1
-	})
+	key := func(val interface{}) interface{} {
```
ejj: Btw, is there any way to get more specific with the types on this?

db/connection.go
```diff
@@ -23,15 +23,15 @@ func (db Database) InsertConnection() Connection {
 }
```
ejj: Haven't looked at the whole patch super carefully beyond the actual hashjoin function, but overall this looks good!

join/join.go
```diff
@@ -73,3 +73,64 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+		lKey = identity
+	}
+	if rKey == nil {
+		rKey = identity
+	}
+
+	// lonely lefts are tracked implicity by remaining elements in joinTable
+	joinTable := make(map[interface{}]*interface{})
+
+	// Build the join table and lonely lefts set using lSlice
```
ejj: Comment is out of date.  Not sure it's useful either, would probably just ditch it.

join/join.go
```diff
@@ -73,3 +73,64 @@ func sliceDel(slice []interface{}, i int) []interface{} {
...
+	for ii := 0; ii < lSlice.Len(); ii++ {
```
ejj: Btw, why do you do this `ii` thing?  Never seen it before

mmussomele: I've found that when I need to refactor or search for loop variables, just 'i' (or j or k) gets tagged a lot as a letter in other variables/comments, but nothing else usually has 'ii' in it. There are tools and options you can use in searches to mitigate that but this has always been simplest for me.

join/join.go
mmussomele: fixed!

join/join.go
ejj: It's not a bad idea, but it's not in line with the rest of the code.  For consistency sake, just `i` is better I think.

ejj:
So finally had a chance to look at this.  I don't think the lonely lambdas add much value.  Simpler to just stick with the generic hash join patch.  I'll do a careful review of it now.

For that patch btw, the commit message should be tweaked.  "adding" should be "Add".  Also, this patch is significant enough to be worthy of more explanation in the commit message.  Something along the lines of "we do joins a lot, they're expensive, worth linear time."

Also did you migrate the hand-rolled hash joins in the patch to this new algo?

db/connection.go
```diff
@@ -23,15 +23,15 @@ func (db Database) InsertConnection() Connection {
...
 
 // SelectFromConnection gets all connections in the database that satisfy 'check'.
-func (db Database) SelectFromConnection(check func(Connection) bool) []Connection {
+func (db Database) SelectFromConnection(check func(Connection) bool) ConnectionSlice {
```
ejj: So I'm somewhat conflicted on this, but I think we should just return `[]Connection` (same for all the other db tables).  Basically, we should use the native slice, i.e `[]Connection` throughout most of the code, as it's the most natural representation.  Then, those callers that need to actually do a join can cast to `db.ConnectionSlice` when they call the join.

minion/engine.go
```diff
@@ -124,26 +120,42 @@ func updatePlacements(view db.Database, spec dsl.Dsl) {
...
-	dslcs := spec.QueryConnections()
-	for _, dbc := range view.SelectFromConnection(nil) {
-		key := dsl.Connection{
-			From:    dbc.From,
-			To:      dbc.To,
-			MinPort: dbc.MinPort,
-			MaxPort: dbc.MaxPort,
+	scs, vcs := makeConnectionSlice(spec.QueryConnections()), view.SelectFromConnection(nil)
+
+	// the key for a db.Connection is all fields excluding it's ID
```
ejj: This comment adds no value.

minion/engine.go
ejj: Actually more I think about it.  Cleanest thing would be to, in a separate patch, change dsl.QueryConnections() to return a slice.

mmussomele:
Fixed what I could find, let me know if I missed anything

ejj:
Merged

PR #416, opened 2016-05-29
config: Update the default stitch

nlsun:


config.spec
```diff
@@ -1,28 +1,42 @@
...
-(label "purple" (makeList 1 (docker "nginx")))
-
-(connect (list 1024 65535) "red" "blue")
-(connect (list 1024 65535) "blue" "red")
-
-(place (labelRule "exclusive" "yellow") "purple")
+(define AdminACL (list "local" "127.0.0.1/32"))
+
+
+// This is a global constant, which is discouraged in favor of local
```
kklin: Now that we have `set`, `define`s are no longer constants.

config.spec
```diff
@@ -1,28 +1,42 @@
...
+  (connect (list 1024 65535) "red" "blue")
+  (connect (list 1024 65535) "blue" "red")
+
+  (place (labelRule "exclusive" "yellow") "purple"))
+
+
+// This is a function that launches machines.
+(define (createMachines nMaster nWorker)
+  (machineAttribute
+    (list (machineAttribute (makeList nMaster (machine)) (role "Master"))
```
kklin: Don't think we need a `machineAttribute` here, could just do `(machine (role "Master"))`.

config.spec
nlsun: oh, does set mutate a defined variable?

ejj:
Merged

PR #420, opened 2016-06-01

### LINKED ISSUE #318: azure: Fix virtual networking setup
Right now, Azure boots each machine in a completely new virtual network. This means private IP's aren't really good for anything, and all communication must happen over public IP's.

The fix for this is to create a single virtual network as part of the provider initialization, and then boot each machine inside that virtual network.

I spent some time playing around with this in the azurecli, and it seems that virtual networking may not be properly supported in the "classic manager" -- instead, we'll have to use the "azure resource manager".

### LINKED ISSUE #368: Azure: Switch to resource manager from classic manager
Using classic manager has, apparently, caused all sorts of delightful problems.  Should switch to the resource manager

ZhongtianWang:
Before this patch, each vm was not booted in the same virtual network, and we are not taking adventage of private IPs. In this patch, we migrated from classic manager to Azure Resource Manager, as classic manager does not support the current model well.

This should close #318, close #368


provider/azure.go
```diff
@@ -1,44 +1,52 @@
...
+	storageName             string
+	storageType             string
+	subnet1Name             string
+	vnetName                string
+	vnetAddressPrefix       string
+	vnetSubnetAddressPrefix string
+	nsgName                 string
+	imagePublisher          string
+	imageOffer              string
+	imageSku                string
```
ejj: A lot of these things should be consts outside of the struct I think. Basically, if you set the value at struct initialization time, and it never changes, it should be const.

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
 		return errors.New("namespace cannot be empty")
 	}
 
-	keyfile := filepath.Join(os.Getenv("HOME"), ".azure", "azure.publishsettings")
+	var err error
```
ejj: No need to declare `err` here.  Line below can just be:
```
if err := clst.loadCredentials(); err != nil {
```

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+
+	err = clst.loadCredentials()
+	if err != nil {
+		return errors.New("failed loading credentials")
+	}
 
-	azureClient, err := management.ClientFromPublishSettingsFile(keyfile, "")
+	clst.spt, err = clst.newServicePrincipalTokenFromCredentials()
```
ejj: What a silly name for a function ... "newSericePrincipalTokenFromCredentials" ... Microsoft =(

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
 		return errors.New("namespace cannot be empty")
 	}
 
-	keyfile := filepath.Join(os.Getenv("HOME"), ".azure", "azure.publishsettings")
+	var err error
+
+	err = clst.loadCredentials()
+	if err != nil {
+		return errors.New("failed loading credentials")
```
ejj: "failed to load azure credentials"

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+	}
 
-	azureClient, err := management.ClientFromPublishSettingsFile(keyfile, "")
+	clst.spt, err = clst.newServicePrincipalTokenFromCredentials()
 	if err != nil {
-		return errors.New("error retrieving azure client from publishsettings")
+		return errors.New("failed generating service principal token")
```
ejj: "failed to generate service principal token"

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+	clst.vnetSubnetAddressPrefix = "10.0.0.0/24"
+	clst.nsgName = "quiltnsg"
+	clst.imagePublisher = "Canonical"
+	clst.imageOffer = "UbuntuServer"
+	clst.imageSku = "15.10"
+	clst.imageVersion = "15.10.201511161"
+	clst.adminUsername = "ubuntu"
+	clst.adminPassword = uuid.NewV4().String()
+
+	err = clst.createOrUpdateResourceGroup()
```
ejj: combine this and the next line

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
 			continue
 		}
-		id := hs.ServiceName
 
-		// Will return empty string if the hostedservice does not have a deployment.
-		// e.g. some hosted services contains only a storage account, but no deployment.
-		deploymentName, err := clst.vmClient.GetDeploymentName(id)
-		if err != nil {
-			return nil, err
+		tags := *vm.Tags
```
ejj:
```
tag := *vm.Tags[clst.tagKey]
if tag == nil || tag != ...
```

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+		if tags[clst.tagKey] == nil || *tags[clst.tagKey] != clst.namespace {
+			continue
 		}
 
-		if deploymentName == "" {
-			clst.instanceDel(id)
-			continue
+		nicname := *vm.Name + "-nic"
+
+		iResult, err := ic.Get(clst.rgName, nicname, "")
```
ejj: This line is a bit mysterious.  Could benefit from a short comment.

ejj: also, iResult isn't a really good name for this variable.  Not particularly descriptive.

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
-			return nil, err
+		tags := *vm.Tags
+		if tags[clst.tagKey] == nil || *tags[clst.tagKey] != clst.namespace {
+			continue
 		}
 
-		if deploymentName == "" {
-			clst.instanceDel(id)
-			continue
+		nicname := *vm.Name + "-nic"
```
ejj: Btw, why not just call `nicname` `nic`?  At the very least `nicName` is easier to read.

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
-	clst.hsClient = hostedservice.NewClient(azureClient)
-	clst.vmClient = virtualmachine.NewClient(azureClient)
 	clst.namespace = namespace
-	clst.storageAccount = storageAccount
-	clst.location = clusterLocation
-	clst.vmImage = vmImage
-	clst.username = username
-	clst.userPassword = uuid.NewV4().String() // Randomly generate pwd
+	clst.rgName = "quilt"
+	clst.rgLocation = "centralus"
```
ejj: This should be set-able btw.  Fine to do it in another patch

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+	clst.tagKey = "quilt-namespace"
+	clst.storageName = "quiltstorageaccnt"
+	clst.storageType = "Standard_GRS"
+	clst.subnet1Name = "quiltsubnet1"
```
ejj: why is this subnet1Name? Why not just subnetName? Or just subnet?

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+	clst.vnetName = "quiltvnet"
+	clst.vnetAddressPrefix = "10.0.0.0/16"
```
ejj: These two things could use comments.  Why are they not both /16?  Why not /8?

Also, if we have some choice the 172 prefix (look it up precisely), is slightly better.  Less confusing as we use 10/8 internally within Quilt.

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+	clst.vnetSubnetAddressPrefix = "10.0.0.0/24"
+	clst.nsgName = "quiltnsg"
+	clst.imagePublisher = "Canonical"
+	clst.imageOffer = "UbuntuServer"
+	clst.imageSku = "15.10"
```
ejj: btw, if Ubuntu 16.04 is available, that's slightly better.

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+		nicname := *vm.Name + "-nic"
+
+		iResult, err := ic.Get(clst.rgName, nicname, "")
+		if err != nil {
+			return nil, err
 		}
+		privateIP := *(*iResult.Properties.IPConfigurations)[0].Properties.PrivateIPAddress
 
-		deploymentResponse, err := clst.vmClient.GetDeployment(id, deploymentName)
+		pipResult, err := pipac.Get(clst.rgName, nicname, "")
```
ejj: pipResult could just be pip

provider/azure.go
```diff
@@ -47,69 +55,94 @@ func (clst *azureCluster) Connect(namespace string) error {
...
-	clst.azureClient = azureClient
-	clst.hsClient = hostedservice.NewClient(azureClient)
-	clst.vmClient = virtualmachine.NewClient(azureClient)
 	clst.namespace = namespace
-	clst.storageAccount = storageAccount
-	clst.location = clusterLocation
-	clst.vmImage = vmImage
-	clst.username = username
-	clst.userPassword = uuid.NewV4().String() // Randomly generate pwd
+	clst.rgName = "quilt"
```
ejj: Maybe "resourceGroup"

provider/azure.go
```diff
@@ -122,20 +155,62 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
 		panic("boot count cannot be negative")
 	}
 
+	var err error
```
ejj: Why?

provider/azure.go
```diff
@@ -122,20 +155,62 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
 			return err
 		}
-	}
 
+		spf := network.SubnetPropertiesFormat{
+			AddressPrefix: &clst.vnetSubnetAddressPrefix,
+		}
+
+		subnetID := fmt.Sprintf(
+			"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/virtualNetworks/%s/subnets/%s",
```
ejj: This probably spills over length.  Could use a `+` on the format string.

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
 
 // Disconnect.
 func (clst *azureCluster) Disconnect() {
-	// nothing
+	gc := resources.NewGroupsClient(clst.subscriptionID)
+	gc.Authorizer = clst.spt
+	cancel := make(chan struct{})
+	gc.Delete(clst.rgName, cancel)
```
ejj: Can delete fail?  What if it fails? Does it block? How long does it block?  The fact that it takes a cancel chan is odd . . . makes me thing we aren't dealing with this function properly.

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
 
-// Create one Azure instance (blocking).
-func (clst *azureCluster) instanceNew(name string, vmSize string, region string,
-	cloudConfig string) error {
-	if region != clst.location {
-		return fmt.Errorf("cannot create instance in %s, only %s is supported", region, clusterLocation)
+func (clst *azureCluster) loadCredentials() error {
+	u, err := user.Current()
+	if err != nil {
+		return errors.New("unable to determine current OS user")
```
ejj: "current user"

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
-			ServiceName: name,
-			Description: clst.namespace,
-			Location:    clst.location,
-			Label:       base64.StdEncoding.EncodeToString([]byte(name)),
-		})
+
+	n := u.HomeDir + credentialsPath
+	f, err := os.Open(n)
+	if err != nil {
+		return errors.New("unable to open credentials at " + n)
```
ejj: "azure credentials"

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
+	}
+
+	b, err := ioutil.ReadAll(f)
+	if err != nil {
+		return errors.New("unable to read " + n)
+	}
+
+	c := map[string]interface{}{}
```
ejj: Will a map[string][string] work?

ejj: Btw this whole process is a bit odd.  Are you sure that, now that we've updated the azure SDK, they haven't come up with an officially sanctioned way to pull the credentials in?  Seems like every user will have this problem ...  What do they suggest?

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
+	err = json.Unmarshal(b, &c)
+	if err != nil {
+		return errors.New(n + " contained invalid JSON")
+	}
+
+	var ok bool
```
ejj: Why?

Perhaps you aren't aware of `:=`?

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
-	// create hostedservice
-	err := clst.hsClient.CreateHostedService(
-		hostedservice.CreateHostedServiceParameters{
-			ServiceName: name,
-			Description: clst.namespace,
-			Location:    clst.location,
-			Label:       base64.StdEncoding.EncodeToString([]byte(name)),
-		})
+
+	n := u.HomeDir + credentialsPath
```
ejj: `n` isn't a great name for this variable.  Sounds like a counter.

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
+	}
+
+	if clst.subscriptionID, ok = c["subscriptionID"].(string); !ok {
+		return errors.New(n + " contained invalid subscriptionID")
+	}
+
+	return nil
+}
+
+func (clst *azureCluster) newServicePrincipalTokenFromCredentials() (*azure.ServicePrincipalToken, error) {
```
ejj: Thus function name is *way to long*.

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
+	oauthConfig, err := azure.PublicCloud.OAuthConfigForTenant(clst.tenantID)
+	if err != nil {
+		return nil, err
+	}
+	return azure.NewServicePrincipalToken(*oauthConfig, clst.clientID, clst.clientSecret, azure.PublicCloud.ResourceManagerEndpoint)
+}
+
+func (clst *azureCluster) createOrUpdateResourceGroup() error {
```
ejj: also this one

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
+		Name:     &clst.rgName,
+		Location: &clst.rgLocation,
+	}
+
+	_, err := gc.CreateOrUpdate(clst.rgName, model)
+
+	return err
+}
+
+func (clst *azureCluster) createOrUpdateStorageAccount(location string) error {
```
ejj: `configureStorageAccount`

provider/azure.go
```diff
@@ -144,83 +219,395 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
+
+	return nil
+}
+
+func (clst *azureCluster) newServicePrincipalTokenFromCredentials() (*azure.ServicePrincipalToken, error) {
+	oauthConfig, err := azure.PublicCloud.OAuthConfigForTenant(clst.tenantID)
+	if err != nil {
+		return nil, err
+	}
+	return azure.NewServicePrincipalToken(*oauthConfig, clst.clientID, clst.clientSecret, azure.PublicCloud.ResourceManagerEndpoint)
```
ejj: Line is too long.

provider/azure.go
ZhongtianWang: Yeah. By the way, we should also let users specify their subscription ID, as well as OS image, right?

provider/azure.go
ZhongtianWang: 😅😅😅😅😅

provider/azure.go
ZhongtianWang: I thought about it too, but in the end I was thinking that the name of a resource group doesn't really matter as much, since it's just a group name that contains all kinds of resources under it (nic, vm, network interface). However, the location does impact as little bit when reading what kinds of resources does this group contain, so I definitely shouldn't just set it to central US. Should we just include all these fields in dsl? It definitely gives users more control over everything, but it might be annoying for many people, too.

provider/azure.go
ZhongtianWang: I was playing with different subnets. Now I'm guessing we probably would never need more than 1 subnet, so lemme change it :P

provider/azure.go
ZhongtianWang: This is so far their official way to pull in the credentials. They showed some code in an example folder, and what we currently have is very similar to what they have. The difference is, I'm checking 4 fields manually, while they use code to check all fields in a loop, and they try to convert everything to string.

provider/azure.go
ZhongtianWang: haha I thought this way looks prettier in the beginning...  🙃  But `:=` should be fine

provider/azure.go
ZhongtianWang: Actually if we don't declare it here, we would still need to declare it right after this line. Because we can't have something like `clst.spt, err := clst.newServicePrincipalToken()` since clst.spt is not an identifier.

provider/azure.go
ZhongtianWang: Actually, we can't have line like this `if clst.clientID, ok := cred["clientID"].(string); !ok {`since clst.clientID cannot be at the left side of `:=`

ejj:
Btw, could you rebase?  I updated the vendor directory, so I think the first patch is no longer necessary.

provider/azure.go
```diff
@@ -1,44 +1,55 @@
...
 	uuid "github.com/satori/go.uuid"
 )
 
-const storageAccount string = "netsysstorage"
-const clusterLocation string = "Central US"
-const vmImage string = "b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-15_10-amd64-server-20151116.1-en-us-30GB"
-const username string = "ubuntu"
+const credentialsPath = "/.azure/credentials.json"
+
+var (
```
ejj: All these should be const

provider/azure.go
```diff
@@ -1,44 +1,55 @@
...
+	"github.com/Azure/go-autorest/autorest/to"
+
 	uuid "github.com/satori/go.uuid"
 )
 
-const storageAccount string = "netsysstorage"
-const clusterLocation string = "Central US"
-const vmImage string = "b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-15_10-amd64-server-20151116.1-en-us-30GB"
-const username string = "ubuntu"
+const credentialsPath = "/.azure/credentials.json"
```
ejj: Why is this treated differently than below?

provider/azure.go
```diff
@@ -1,44 +1,55 @@
...
+
+var (
+	tagKey                  = "quilt-namespace"
+	resourceGroup           = "quilt"
+	rgLocation              = "centralus"
+	storageName             = "quiltstorageaccnt"
```
ejj: Camel Case  Also perhaps the variable should be calle storageAccount

provider/azure.go
```diff
@@ -1,44 +1,55 @@
...
+	storageType             = "Standard_GRS"
+	subnetName              = "quiltsubnet"
```
ejj: Just call the variable "subnet" and the value quiltSubnet  basically I don't like "Name" in variable names, it's redundant.

provider/azure.go
```diff
@@ -1,44 +1,55 @@
...
+	vnetName                = "quiltvnet"
+	vnetAddressPrefix       = "10.0.0.0/8"
+	vnetSubnetAddressPrefix = "10.0.0.0/8"
+	nsgName                 = "quiltnsg"
```
ejj: This is pretty mysterious.  What's an nsg?  Actually come to think of it, most of these constants deserve a comment.

provider/azure.go
```diff
@@ -1,44 +1,55 @@
...
+	clientSecret   string
+	tenantID       string
+	subscriptionID string
 	namespace      string
-	storageAccount string
-	location       string
-	vmImage        string
-	username       string
-	userPassword   string // Required password is a randomly generated UUID.
+	adminUsername  string
```
ejj: Does this need to be part of azureCluster, or can it just be generated as needed?

provider/azure.go
ejj: Ah if that's the case, still cleaner to do the following 
```
err := clst.LoadCredentials
if err != nil {
    ....
}
```

provider/azure.go
```diff
@@ -47,69 +58,79 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+	if err = clst.configureResourceGroup(); err != nil {
+		return err
+	}
 
 	return nil
 }
 
 // Retrieve list of instances.
 func (clst *azureCluster) List() ([]Machine, error) {
+	vmc := compute.NewVirtualMachinesClient(clst.subscriptionID)
```
ejj: Does it make sense to be generating this on each function call, or instead, should they be part of azureCluster?

provider/azure.go
```diff
@@ -47,69 +58,79 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+	vmc.Authorizer = clst.spt
+
+	ic := network.NewInterfacesClient(clst.subscriptionID)
+	ic.Authorizer = clst.spt
+
+	pipac := network.NewPublicIPAddressesClient(clst.subscriptionID)
+	pipac.Authorizer = clst.spt
+
 	var mList []Machine
```
ejj: I'd define this above the for loop as that's closest to it first use.

provider/azure.go
```diff
@@ -47,69 +58,79 @@ func (clst *azureCluster) Connect(namespace string) error {
...
-		if err != nil {
-			return nil, err
+		if tag := (*vm.Tags)[tagKey]; tag == nil || *tag != clst.namespace {
+			continue
 		}
 
-		if deploymentName == "" {
-			clst.instanceDel(id)
-			continue
+		nicName := *vm.Name + "-nic"
```
ejj: just `nic` instead of `nicName`.  Also can vm.Name be nil?  If not, why is it a *?

provider/azure.go
```diff
@@ -47,69 +58,79 @@ func (clst *azureCluster) Connect(namespace string) error {
...
+
+		// retreive details about the specified network interface
+		ni, err := ic.Get(resourceGroup, nicName, "")
+		if err != nil {
+			return nil, err
 		}
+		privateIP := *(*ni.Properties.IPConfigurations)[0].Properties.PrivateIPAddress
```
ejj: You're doing a lot of dereferencing and slice element access.  Are you sure these things can't be nil or be empty?  Probably should check and return errors.

provider/azure.go
```diff
@@ -47,69 +58,79 @@ func (clst *azureCluster) Connect(namespace string) error {
...
 
-		deploymentResponse, err := clst.vmClient.GetDeployment(id, deploymentName)
+		pip, err := pipac.Get(resourceGroup, nicName, "")
 		if err != nil {
 			return nil, err
 		}
+		publicIP := *pip.Properties.IPAddress
```
ejj: Could this be nil?

provider/azure.go
```diff
@@ -123,16 +144,50 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
 	}
```
ejj: Btw, above there's some weirdness.  This panic check is useless and the function comment has a cryptic statement about instanceNew that isn't called anywhere.  Probably should fix in a separate patch.

provider/azure.go
```diff
@@ -123,16 +144,50 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
 
 	for _, m := range bootSet {
-		name := "quilt-" + uuid.NewV4().String()
-		if err := clst.instanceNew(name, m.Size, m.Region, cloudConfigUbuntu(m.SSHKeys, "wily")); err != nil {
+		// One storage account is needed per region
+		if err := clst.configureStorageAccount(m.Region); err != nil {
```
ejj: Should we really be re-creating the storage account for every region _every single time_ we boot a machine?  Seems to me, this could be done once for all regions when the cluster is create.

Btw, I'm really surprised this works ... based on my reading of configureStorageAccount(), my impression is this should fail because we keep re-trying to create the storage account with the same name.

provider/azure.go
```diff
@@ -123,16 +144,50 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
 			return err
 		}
-	}
 
+		spf := network.SubnetPropertiesFormat{
+			AddressPrefix: &vnetSubnetAddressPrefix,
+		}
+
+		subnetID := fmt.Sprintf(
+			"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/"+
```
ejj: This is over 89 characters.

provider/azure.go
```diff
@@ -123,16 +144,50 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+				"virtualNetworks/%s/subnets/%s",
+			clst.subscriptionID,
```
ejj: No need for a newline after each function argument.

provider/azure.go
```diff
@@ -123,16 +144,50 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+			vnetName,
+			subnetName)
+
+		subnet := network.Subnet{
+			Name:       &subnetName,
+			Properties: &spf,
+			ID:         &subnetID,
+		}
+
+		if err := clst.configureVirtualNetwork(subnet, m.Region); err != nil {
```
ejj: Again, does this really have to be done for *every* machine, or can it be done once per region?

provider/azure.go
```diff
@@ -123,16 +144,50 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+			return err
+		}
+
+		vmName := "quilt-" + uuid.NewV4().String()
+		osDiskName := vmName + "-osdisk"
+		nicName := vmName + "-nic"
+
+		if err := clst.configureNetworkInterface(nicName, subnet, m.Region); err != nil {
```
ejj: Line is too long.

ejj: btw, you don't always have to do the err check in a single if statment.  Just if the line is short.  In this case I'd just break this into two lines, one makes the function call, the other does the if check

provider/azure.go
```diff
@@ -123,16 +144,50 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+			return err
+		}
+
+		cloudConfig := cloudConfigUbuntu(m.SSHKeys, "wily")
```
ejj: Pull this outside of the loop.

provider/azure.go
```diff
@@ -144,83 +199,399 @@ func (clst *azureCluster) Stop(machines []Machine) error {
 
 // Disconnect.
 func (clst *azureCluster) Disconnect() {
-	// nothing
+	gc := resources.NewGroupsClient(clst.subscriptionID)
+	gc.Authorizer = clst.spt
+	cancel := make(chan struct{})
+	gc.Delete(resourceGroup, cancel)
```
ejj: Does this delete all the storage accounts and virtual networks and subnets and everything we've created?

provider/azure.go
```diff
@@ -144,83 +199,399 @@ func (clst *azureCluster) Stop(machines []Machine) error {
 
 // Disconnect.
```
ejj: This comment adds nothing btw.

provider/azure.go
ejj: Actually, it doesn't really matter, we never actually call Disconnect apparently.

provider/azure.go
```diff
@@ -144,83 +199,399 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
+	if err != nil {
+		return errors.New("unable to open Azure credentials at " + dir)
+	}
+
+	b, err := ioutil.ReadAll(f)
+	if err != nil {
+		return errors.New("unable to read " + dir)
+	}
+
+	cred := map[string]interface{}{}
```
ejj: Why not map[string]string?

provider/azure.go
```diff
@@ -144,83 +199,399 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
-	err := clst.hsClient.CreateHostedService(
-		hostedservice.CreateHostedServiceParameters{
-			ServiceName: name,
-			Description: clst.namespace,
-			Location:    clst.location,
-			Label:       base64.StdEncoding.EncodeToString([]byte(name)),
-		})
+
+	dir := u.HomeDir + credentialsPath
+	f, err := os.Open(dir)
```
ejj: Should you defer closing the file?

provider/azure.go
```diff
@@ -144,83 +199,399 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
+	}
+
+	if clst.subscriptionID, ok = cred["subscriptionID"].(string); !ok {
+		return errors.New(dir + " contains invalid subscriptionID")
+	}
+
+	return nil
+}
+
+func (clst *azureCluster) newServicePrincipalToken() (*azure.ServicePrincipalToken,
```
ejj: I think the name `newToken` is a tad cleaner.

provider/azure.go
```diff
@@ -1,44 +1,55 @@
...
-const storageAccount string = "netsysstorage"
-const clusterLocation string = "Central US"
-const vmImage string = "b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-15_10-amd64-server-20151116.1-en-us-30GB"
-const username string = "ubuntu"
+const credentialsPath = "/.azure/credentials.json"
+
+var (
+	tagKey                  = "quilt-namespace"
+	resourceGroup           = "quilt"
+	rgLocation              = "centralus"
```
ejj: Is this going to be an issue?  How will this work with Azure VMs in other regions?

provider/azure.go
```diff
@@ -144,83 +199,399 @@ func (clst *azureCluster) Stop(machines []Machine) error {
...
+	// Create public IP address
+	pipac := network.NewPublicIPAddressesClient(clst.subscriptionID)
+	pipac.Authorizer = clst.spt
+
+	pipapf := network.PublicIPAddressPropertiesFormat{
+		PublicIPAllocationMethod: network.Dynamic,
+	}
+
+	pipaID := fmt.Sprintf(
+		"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/"+
```
ejj: You have some very similar code strewn about to generate these IDs.  Perhaps you could pull it into a function?

ejj:
So I'm going to stop reviewing this for now.  There are some high level issues I think need to be resolved.

1) We're doing a lot of work per-VM ... creating various subnets and what not.  I think this is likely what's causing a lot of the performance issues.

2) the SetACLs situation isn't good enough at the moment.  Basically, you're creating all of the security groups at NIC creation time, but that really should be handled dynamically in the SetACLs function.

3) Really probably should do the ChooseBestSize thingy.

At this point, this seems like a promising prototype, but until those three issues are resolved, I don't think we should merge.  Doesn't have to be in the same commit, but probably should be in PR.  I'm going to close the PR for now when that stuffs done, go ahead and resubmit.

provider/azure.go
ZhongtianWang: But eventually we would need pointers to there vars. If those are const, we would still need to create temp variables and use their addresses.

provider/azure.go
ZhongtianWang: Because it's a pointer

provider/azure.go
ZhongtianWang: Camel case is not allowed for storage account

Issue #407, opened 2016-05-28
System should warn when masters but no workers are specified


ejj:
If the user specifies master vms, but doesn't specify any worker vms, it's almost certainly a bug.  The system should warn at us in that case.

ejj:
Done

PR #425, opened 2016-06-02
engine: Warning message for no workers

secant:
Added warning message for users that specify a master machine but
no worker machines. Addresses issue #407.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -149,26 +149,32 @@ func toDBMachine(machines []dsl.Machine, maxPrice float64) []db.Machine {
 	return dbMachines
 }
 
-// Quilt can only boot if there is at least one master and one worker.
-func canBoot(machines []db.Machine) bool {
-	hasMaster := false
-	hasWorker := false
-	for _, machine := range machines {
-		hasMaster = hasMaster || machine.Role == db.Master
-		hasWorker = hasWorker || machine.Role == db.Worker
-	}
-	return hasMaster && hasWorker
-}
-
 func machineTxn(view db.Database, dsl dsl.Dsl, clusterID int) error {
 	// XXX: How best to deal with machines that don't specify enough information?
 	dslMachinesRaw := dsl.QueryMachines()
 	maxPrice, _ := dsl.QueryFloat("MaxPrice")
 
 	var dslMachines = toDBMachine(dslMachinesRaw, maxPrice)
+	hasMaster := false
+	hasWorker := false
 
-	if !canBoot(dslMachines) {
+	for _, machine := range dslMachines {
+		hasMaster = hasMaster || machine.Role == db.Master
+		hasWorker = hasWorker || machine.Role == db.Worker
+	}
+
+	if !hasMaster || !hasWorker {
 		dslMachines = []db.Machine{}
+
+		if hasMaster != hasWorker {
+
+			if hasMaster {
+				log.Warning("A master machine was specified but worker machines were not.")
+			} else {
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -157,6 +157,11 @@ func canBoot(machines []db.Machine) bool {
 		hasMaster = hasMaster || machine.Role == db.Master
 		hasWorker = hasWorker || machine.Role == db.Worker
 	}
+
+	if hasMaster && !hasWorker {
+		log.Warning("No worker machines were specified but a master machine was.")
```
kklin: Maybe it'd be better to log based on the return value of `canBoot` in `machineTxn`? Not sure.

secant:
Not sure if I did it in the prettiest way 😅

engine/engine.go
```diff
@@ -167,7 +167,13 @@ func machineTxn(view db.Database, dsl dsl.Dsl, clusterID int) error {
 
 	var dslMachines = toDBMachine(dslMachinesRaw, maxPrice)
 
-	if !canBoot(dslMachines) {
+	canBoot, hasWorker := canBoot(dslMachines)
+	if !canBoot {
+
+		if !hasWorker {
```
ejj: So I think the issue was a bit too specific.  It's not so much the fact that a master was specified without workers that's the problem.  Really, it was just that if canBoot() returns false, we do nothing without warning.  Therefore, I don't think we really need to differentiate between whether or not has Worker is true.

As a result, I'd just leave canBoot as it is, log if it's false saying something along the lines of: you're missing masters and/or workers, so we aren't booting.

engine/engine.go
```diff
@@ -168,6 +168,7 @@ func machineTxn(view db.Database, dsl dsl.Dsl, clusterID int) error {
 	var dslMachines = toDBMachine(dslMachinesRaw, maxPrice)
 
 	if !canBoot(dslMachines) {
+		log.Warning("No worker machines and/or no master machines were specified.")
```
ejj: Sorry, I'm being super flip-floppy on this.  So the problem with this is we warn when there are neither masters nor workers.  That actually shouldn't warn as that's how you indicate that the cluster should be deleted.  What we really want to do is warn if workers are specified and masters aren't or, if masters are specified and workers aren't.

Btw, the canBoot function is useless.  Probably the easiest way to achieve this is to ditch canBoot, pull it's code into machineTxn, and then you can implement the logic.

Sorry to be so wishy-washy about this . . .

engine/engine.go
```diff
@@ -149,26 +149,32 @@ func toDBMachine(machines []dsl.Machine, maxPrice float64) []db.Machine {
...
-	if !canBoot(dslMachines) {
+	for _, machine := range dslMachines {
+		hasMaster = hasMaster || machine.Role == db.Master
+		hasWorker = hasWorker || machine.Role == db.Worker
+	}
+
+	if !hasMaster || !hasWorker {
 		dslMachines = []db.Machine{}
+
+		if (hasMaster && !hasWorker) || (hasWorker && !hasMaster) {
```
ejj: `hasMaster != hasWorker`

ejj:
One tiny nit.  Looks good otherwise.  Ping me when you've update and I'll merge.

ejj:
Oh two minor nits about the commit message for future reference.

1) Please word wrap to 70 characters.  Why 70 characters for commit messages?  I have no idea ... it's the standard for email which commits are occasionally sent in.

2) Don't mention issue numbers in the commit message.  A lot of projects do this, I don't particularly like it.  No good reason.

PR #426, opened 2016-06-02
documentation: Add getting started guide for Azure

ZhongtianWang:


Azure.md
```diff
@@ -0,0 +1,27 @@
+# Azure Getting Started Guide for Quilt
```
ejj: Word Wrap to 80

Azure.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Azure Resource Manager
+Azure introduced the Azure Resource Manager (ARM) deployment model as an enhancement of the previous Classic deployment model. Read [here](https://azure.microsoft.com/en-us/documentation/articles/resource-manager-deployment-model/) for more details.
```
ejj: Hmm, I'd  just say something like "Microsoft Azure recommends using the Azure Resource Manager (ARM) model ... which Quilt uses ... " Something like that.  No need to mention Classic, or call out the distinction between them.

Azure.md
```diff
@@ -0,0 +1,27 @@
+# Azure Getting Started Guide for Quilt
+
+## Azure Resource Manager
+Azure introduced the Azure Resource Manager (ARM) deployment model as an enhancement of the previous Classic deployment model. Read [here](https://azure.microsoft.com/en-us/documentation/articles/resource-manager-deployment-model/) for more details. 
+
+
+## Authenticate ARM
+To access and modify resources, you must set up an AD (Active Directory) application, so that Quilt's identity is granted direct access to your Azure resources.
```
ejj: You mention `identity` without defining it which is a bit confusing.  Maybe just `so that Quilt is granted`?

Azure.md
```diff
@@ -0,0 +1,27 @@
...
+
+Before going into more details, it is worth noting the following two points:
+
+1. Permissions around AD accounts don't fall in line exactly with Azure subscription administrators. You might run into issues such as unable to view or create AD applications. So before continuing, make sure that you are a global admin the directory to which you are part of. This happens often if you are working on someone else's subscription and the subscription owner added you as co-administrator in his/her subscription.
```
ejj: `directory` is sued before defined.  also `global admin of` the directory.  Also might be worth noting how to check if you're global admin, and (roughly) where you can change the setting.  Also, is there a link to any azure documentation worth pointing to here?

Azure.md
```diff
@@ -0,0 +1,27 @@
...
+
+2. You don't need multiple AD accounts per subscription. If there already exists an AD application with correct permissions, you can just generate a new key (client secret) for yourself.
```
ejj: This sounds a bit mysterious to the un-initiated.  What's the different between an AD account and an AD application?  Is an AD account basically a user?

Azure.md
```diff
@@ -0,0 +1,27 @@
...
+
+See the MSDN article on how to create an AD application: [Use portal to create Active Directory application that can access resources](https://azure.microsoft.com/en-us/documentation/articles/resource-group-create-service-principal-portal/).
```
ejj: Nice, maybe link this earlier?

ejj:
This basically looks good.  Couple of nits.  At a high level, it needs a bit more description of terms.  Most readers of this aren't going to be familiar with Azure . . .  I'd hand hold them through the terminology a bit more.

Azure.md
ZhongtianWang: Sounds good!

Azure.md
ZhongtianWang: AD account acts as an user, and it creates AD application, which is the service principal, that provides api access to let user access resources. Yeah... It is very confusing for something that should be much simpler. Let me make sure to explain this in more details

ejj:
This really depends on the ARM patch being merged.  Overall looks good, but I think it should be folded into that PR when you resubmit it.  Doesn't make sense to merge until that's in.

PR #432, opened 2016-06-07
Hashjoin lint fix

mmussomele:
Added doc comments where they needed to go as relevant to hashjoin. I also went ahead and changed that tick channel thing I asked about earlier just because (separate commit in case you wanna leave it alone)


quilt.go
```diff
@@ -58,16 +57,11 @@ func main() {
 
 	conn := db.New()
 	go func() {
-		tick := time.Tick(5 * time.Second)
-		for {
+		for range time.Tick(5 * time.Second) {
```
ejj: Ah so I realized after we talked on slack, the problem with this is Quilt hangs until the first tick happen.  Technically correct, but it's pretty annoing.

Anyways, I wrote up a patch that deals with this but I haven't merged it yet.  I'd just drop this for now.

mmussomele: Gotcha, threw away that commit

ejj:
Merged

Issue #253, opened 2016-03-29
Container RAM/CPU targets


ejj:
We should be able to specify the RAM/CPU requirements/limits a particular container needs.  Syntax presumably should be similar to the machine RAM/CPU specifications.

This isn't strictly required for the paper, but would be nice to have.

ejj:
This will be super useful for two things.  1) we want the ability to say "the cluster you're attempting to run this on is not big enough"  or alternatively make a recommendation o the number of VMs you'll need to pull of deploying a particular system. 2) Given a particular cluster we want the ability to say "Fill out excess capacity in the cluster with spark workers"

Both of these issues require knowledge of container sizes.

akshayknarayan:
Additionally, 3) the ability to schedule containers on to VMs given their relative resource demands (Dominant Resource Fairness, etc)

akshayknarayan:
Note: disk targets too.

ejj:
So what we really need, is just a way to specify this stuff in the DSL.  For now, we shouldn't actually respect it in the dataplane as that would mean teaching updating our docker swarm code.  Once we've ripped out docker swarm and replaced it with our own scheduler, then we would implement in the dataplane.

secant:
Migrated to Trello.
https://trello.com/c/8rjT2yCN

PR #433, opened 2016-06-07
inspect: Make assertions about reachability, incorporate container pl…

akshayknarayan:
…acement constraints

- de-capitalize types and funcs
- add way to define invariants on the quilt container graph
- add way to query inspect for whether invariants will hold given changes in the container graph
- basic algorithm to place containers on "availability sets" (machines, regions, or providers) and display placement in visualization


Makefile
```diff
@@ -32,6 +32,11 @@ lint: format
 		golint -min_confidence .25 $$package ; \
 	done
 
+inspect:
```
ejj: If you're going to have this target, then I think we should make `all` depend on it, and remove building inspect from line 11.

inspect/graph.go
```diff
@@ -1,95 +1,161 @@
 package main
 
-type Node struct {
-	Name        Label
-	Connections map[string]*Node
+import (
+	"fmt"
+)
+
+type node struct {
```
ejj: The lower casing should probably be in its own patch ...

inspect/placement.go
```diff
@@ -0,0 +1,127 @@
+package main
+
+import (
+	"fmt"
+	"github.com/NetSys/quilt/dsl"
+)
+
+type availabilitySet map[string]struct{}
```
ejj: availabilitySet needs a comment explaining what the concept is.

inspect/graph.go
```diff
@@ -1,95 +1,161 @@
...
+	nodes        map[string]node
+	connections  []connection
+	availability []availabilitySet
 }
 
-type Graph struct {
-	Nodes       map[string]Node
-	Connections []Connection
+func makeGraph() graph {
+	av := make([]availabilitySet, 0)
```
ejj: Don't really need to use make for this.  Something like
```
av := []availabilitySet{availabilitySet{}}
```

ejj: Also, it's probably worth adding a comment why this is initialized with a single availability set instead of none.  It's a bit surprising first time you read it.

inspect/graph.go
```diff
@@ -1,95 +1,161 @@
...
+	av = append(av, make(availabilitySet))
+	return graph{
+		nodes:        make(map[string]node),
```
ejj: In general you can avoid using make for initialization `map[string]node{}` would be better go style.  You do this a lot below, I'm not going to comment on it from here on.

inspect/graph.go
```diff
@@ -1,95 +1,161 @@
...
+	newAvail := make([]availabilitySet, len(g.availability))
+	copy(newAvail, g.availability)
+
+	return graph{nodes: newNodes, connections: newConnections, availability: newAvail}
 }
 
-func (g Graph) getNode(label string) *Node {
-	node, have := g.Nodes[label]
+func (g graph) hasNode(label string) *node {
+	node, have := g.nodes[label]
```
ejj: We've been using `ok` instead of `have` for this purpose.  Doesn't really matter, but better to be consistent.

ejj: Also, i don't think this function really adds anything. Why not just ditch it and have the callers check the map directly?

akshayknarayan:
Commit "insepct: compute availability sets": 
- add ability to output quilt graph as ascii, not just pdf
- refactoring graph data structure to incorporate placement constraints.
- rewrote placement code, algorithm attempts to use as few availability sets as possible
    - placing containers on a minimal number of VMs given placement exclusivity constraints is (NP-complete)[https://en.wikipedia.org/wiki/Clique_cover], the implementation in placement.go:placeNodes() is a naive algorithm and should be considered a first step.
    - after #253 is resolved, this would anyway be replaced by something like DRF.

ejj:
Tian will do the first round review of this

inspect/main.go
```diff
@@ -11,14 +12,21 @@ import (
...
+
 func main() {
-	var configPath string
-	switch len(os.Args) {
-	case 2:
-		configPath = os.Args[1]
-	default:
-		panic("Usage: dslinspect <path to spec file>")
-	}
+	configPath := os.Args[1]
```
ZhongtianWang: We should probably check the length of args as well, and prompt the user with usage if there is no args

inspect/main.go
```diff
@@ -38,30 +46,62 @@ func main() {
...
+
+	for _, m := range spec.QueryMachines() {
+		graph.machines = append(graph.machines, getMachine(m))
 	}
 
-	Graphviz(slug, graph, containerLabels)
+	// actions
+	ignoreNext := 0
+	foundFlags := map[string]argOption{}
+	func() {
```
ZhongtianWang: I think using golang's build-in `flag` package would simplify the argument parsing code a lot.

inspect/placement.go
```diff
@@ -0,0 +1,175 @@
...
+import (
+	"fmt"
+	"github.com/NetSys/quilt/dsl"
+)
+
+// a set of containers which can be placed together on a VM
+type availabilitySet map[string]struct{}
+
+func (avSet availabilitySet) nodes() []string {
+	labels := make([]string, 0, len(avSet))
```
ZhongtianWang: I'd go with just `labels := []string{}` as it might be a better go style. But it's nice to set a cap too... I'm not sure 😛

inspect/placement.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	return labels
+}
+
+func (avSet availabilitySet) str() string {
+	return fmt.Sprint(avSet.nodes())
+}
+
+func (avSet availabilitySet) check(label string) bool {
+	_, ok := avSet[label]
```
ZhongtianWang: No need to declare `ok` here. Line can just be:

if _, ok := avSet[label]; !ok ...

inspect/placement.go
```diff
@@ -0,0 +1,175 @@
...
+func (avSet *availabilitySet) remove(labels ...string) {
+	for _, label := range labels {
+		delete(*avSet, label)
+	}
+}
+
+func (avSet *availabilitySet) removeAndCheck(labels ...string) []string {
+	var toMove []string
+	for _, label := range labels {
+		if there := avSet.check(label); there {
```
ZhongtianWang: `check()` is already returning bool, so we can just use `if avSet.check(label) {`

inspect/placement.go
```diff
@@ -0,0 +1,175 @@
...
+		if existing, ok := g.placement[sep]; ok {
+			g.placement[sep] = append(existing, target)
+		} else {
+			g.placement[sep] = []string{target}
+		}
+	}
+
+	// add extra rule separating "public" into its own availability set
+	// remove once "implicit placement rules" are incorporated into dsl
+	publicLabel := "public"
```
ZhongtianWang: Use `dsl.PublicInternetLabel` instead would be better in my opinion

inspect/query.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+type query struct {
+	form   queryType
+	target string // node to remove
+	str    string // original query text
+}
+
+var queryFormKeywords map[string]queryType
+var queryFormImpls map[queryType]func(graph graph, invs []invariant, query query) *invariant
```
ZhongtianWang: This line (along with a couple ones after) is over 89 characters 😳

inspect/query.go
```diff
@@ -0,0 +1,109 @@
...
+	removeSetQuery
+)
+
+type query struct {
+	form   queryType
+	target string // node to remove
+	str    string // original query text
+}
+
+var queryFormKeywords map[string]queryType
```
ZhongtianWang: What is `queryFormKeywords` for? I don't really see other usages for this

inspect/query.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return nil, nil, nil
+}
+
+func parseQueries(graph graph, path string) ([]query, error) {
+	var queries []query
+
+	var parse func(string)
+	parse = func(line string) {
```
ZhongtianWang: maybe just `parse := func(line string) {`

inspect/query.go
```diff
@@ -0,0 +1,109 @@
...
+		sp := strings.Split(line, " ")
+		_, ok := graph.nodes[sp[1]]
+		if !ok {
```
ZhongtianWang: As before, merge these two to one line could be nice 🎃

inspect/query.go
```diff
@@ -0,0 +1,109 @@
...
+	node := query.target
+	avSet := graphCopy.findAvailabilitySet(node)
+	if avSet == nil {
+		panic(fmt.Errorf("could not find availability set: %s", node))
+	}
+
+	graphCopy.removeAvailabiltySet(avSet)
+
+	_, failer, err := checkInvariants(graphCopy, invs)
+	if err != nil {
```
ZhongtianWang: those could be same line too~  🤓

inspect/viz.go
```diff
@@ -8,6 +8,31 @@ import (
...
 )
 
+func viz(configPath string, spec dsl.Dsl, graph graph, outputFormat string) {
+	var slug string
+	for i, ch := range configPath {
+		if ch == '.' {
+			slug = configPath[:i]
+		}
+	}
+	if len(slug) == 0 {
```
ZhongtianWang: I don't think slug will ever have length 0, because configPath was first checked by `os.Open(configPath)` and that there will always be a correct slug.

ZhongtianWang: Actually why not just use `configPath` as `slug`? The code would be easier to read and the output of the graph would be just something like `config.spec.pdf`. Lemme know what you think 🐹

inspect/viz.go
```diff
@@ -8,6 +8,31 @@ import (
...
+				containerLabels[label] = make([]*dsl.Container, 0)
+			}
+			containerLabels[label] = append(containerLabels[label], container)
+		}
+	}
+
+	graphviz(outputFormat, slug, graph, containerLabels)
+}
+
 // Write parsed Quilt graph to a graphviz dotfile.
```
ZhongtianWang: There is an extra line here

inspect/viz.go
```diff
@@ -8,6 +8,31 @@ import (
...
 
 func getImageNamesForLabel(containerLabels map[string][]*dsl.Container, label string) (imageNames string) {
```
ZhongtianWang: Line is too long 🕵

inspect/viz.go
```diff
@@ -48,20 +73,50 @@ func Graphviz(slug string, graph Graph, containerLabels map[string][]*dsl.Contai
...
 	dotfile += "}\n"
 
 	f.Write([]byte(dotfile))
 
-	// run "dot" (part of graphviz) on the dotfile to output the image
-	writepng := exec.Command("dot", "-Tpdf", "-o", slug+".pdf", slug+".dot")
-	writepng.Run()
+	// Dependencies:
+	// - easy-graph (install Graph::Easy from cpan)
+	// - graphviz (install from your favorite package manager)
```
ZhongtianWang: Maybe put the dependencies notice inside usage prompt as well? Just so it's easier for people to get started.

inspect/viz.go
```diff
@@ -48,20 +73,50 @@ func Graphviz(slug string, graph Graph, containerLabels map[string][]*dsl.Contai
...
+	case "pdf":
+		writeGraph = exec.Command("dot", "-Tpdf", "-o", slug+".pdf", slug+".dot")
+	}
+	stdout, err := writeGraph.CombinedOutput()
+	if err != nil {
+		fmt.Println(string(stdout))
+		panic(err)
+	}
+
+	if len(stdout) > 0 {
```
ZhongtianWang: for cmd, I always use something like `cmd.Stdout = os.Stdout`, it redirects all the output nicely 😃

inspect/graph.go
```diff
@@ -1,98 +1,147 @@
...
 
-// A Graph represents permission to communicate across a series of Nodes.  Each Node is a
+// A graph represents permission to communicate across a series of nodes.  Each node is a
 // container and each edge is permissions to initiate a connection.
-type Graph struct {
-	Nodes       map[string]Node
-	Connections []Connection
+type graph struct {
+	nodes        map[string]node
+	availability []availabilitySet
```
ZhongtianWang: By calling it `availability`, it get slightly confusing and inconsistent with other variable names, like `av` or `avSet`. I think rename it would be a better idea.  😬

inspect/graph.go
```diff
@@ -1,98 +1,147 @@
...
 	}
 
-	var paths []Path
-	var explore func(t *Node, p Path)
-	explore = func(t *Node, p Path) {
-		if t.Name == end.Name {
+	var paths [][]string
+
+	var explore func(t node, p []string)
+	explore = func(t node, p []string) {
```
ZhongtianWang: could be just `explore :=`

ZhongtianWang:
That should be it for now 😬, there are not many big issues, mostly styling. E.g. line over 89 characters.

inspect/main.go
akshayknarayan: most of this code will go away soon anyway as invariant checking is integrated into the dsl - I'm not too worried about the complexity here.

inspect/graph.go
akshayknarayan: since it's recursive, the ```var explore ...``` is needed

inspect/viz.go
akshayknarayan: 1. a filename without the '.' character is possible (and valid from the perspective of ```os.Open``` - in that case slug would have length 0
2. I'm not a big fan of ```config.spec.pdf``` - having multiple filename extensions seems clunky

inspect/graph.go
akshayknarayan: having ```graph.av``` or ```graph.avSets``` seems more opaque than ```graph.availability``` - while ```av``` is fine for a variable with small scope, the struct field name should describe what it represents.

I'm not too sure about this one - what do you think?

inspect/query.go
akshayknarayan: oops, nice catch. the switch statement in ```parseQueries``` will be replaced with a lookup to this map.

inspect/viz.go
akshayknarayan: done.

inspect/query.go
akshayknarayan: done.

ejj:
First letter in the commit message (after the colon) should be capitalized.

Also the "pass linter" patch should be ditched, just fold the fixes into the commits that introduce the errors

akshayknarayan:
done.

inspect/graph.go
```diff
@@ -1,98 +1,131 @@
...
+func makeGraph() graph {
+	return graph{
+		nodes:       make(map[string]node),
+		connections: make([]connection, 0),
 	}
 }
 
-func (g Graph) getNode(label string) *Node {
-	node, have := g.Nodes[label]
+func (g graph) copyGraph() graph {
```
ejj: These function additions are a bit odd in a patch about renaming.  Should they be in a separate patch?

inspect/graph.go
```diff
@@ -1,98 +1,131 @@
...
+		newNodes[label] = node
+	}
+
+	newConnections := make([]connection, len(g.connections))
+	copy(g.connections, newConnections)
+
+	return graph{nodes: newNodes, connections: newConnections}
+}
+
+func (g graph) hasNode(label string) *node {
```
ejj: I think this hasNode function is a bit odd, it's only caller, `getAddNode` could just do the hash table lookup directly and we could ditch this function all together.

inspect/graph.go
```diff
@@ -1,98 +1,131 @@
...
+	node, have := g.nodes[label]
```
ejj: Also, somewhat of an idiom in this code (in go generally I think) would be to name this `have` variable `ok` instead.

inspect/graph.go
```diff
@@ -1,98 +1,131 @@
...
 	}
 
-	return &node
+	return foundNode
+}
+
+func (g *graph) removeNode(node *node) {
+	delete(g.nodes, node.name)
+
+	newConnections := make([]connection, 0)
```
ejj: `var newConnections []connection` is slightly more idiomatic.

inspect/graph.go
```diff
@@ -1,98 +1,131 @@
...
+			if !explored {
+				reached[label] = struct{}{}
+				explore(node)
 			}
 		}
 	}
 	explore(n)
 
-	var reachable []Label
+	reachable := make([]string, 0)
```
ejj: `var reachable []string`

ejj:
First patch basically looks good, looking at the second next.

inspect/graph.go
```diff
@@ -1,78 +1,103 @@
...
-// A Node in the communiction Graph.
+import (
+	"fmt"
+)
+
+// A node in the communiction graph.
 type node struct {
 	name        string
-	connections map[string]*node
+	connections map[string]node
```
ejj: Ah, I like this a lot better. Was wondering about it in the earlier patch.

Now that I think about it, I might pull some this `*node` to `node` change into the first patch of the series, and then change that patch to be about general style cleanups in the inspect code.  That way this patch can be a lot cleaner.

inspect/graph.go
```diff
@@ -1,78 +1,103 @@
...
 }
 
-// A Connection is an edge in the communication Graph.
+// A connection is an edge in the communication graph.
```
ejj: All these comment changes should be in the first patch.

inspect/graph.go
```diff
@@ -1,78 +1,103 @@
...
 
 	newConnections := make([]connection, len(g.connections))
 	copy(g.connections, newConnections)
 
-	return graph{nodes: newNodes, connections: newConnections}
-}
+	newAvail := make([]availabilitySet, len(g.availability))
+	copy(newAvail, g.availability)
 
-func (g graph) hasNode(label string) *node {
```
ejj: Ah, yeah this should be folded into the first patch, definitely

inspect/graph.go
ZhongtianWang: Ohhhhh ops 😁

akshayknarayan:
Made several changes to account for comments. Ready for next round.

cluster/provider/awsSpot.go
```diff
@@ -322,9 +322,12 @@ func (clst amazonCluster) List() ([]Machine, error) {
 	return machines, nil
 }
 
-func (clst *amazonCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
-	maxPrice float64) string {
-	return pickBestSize(awsDescriptions, ram, cpu, maxPrice)
+func (clst *amazonCluster) ChooseSize(
```
ejj: This change doesn't seem to do anything.  Why not leave it?  Diff is cleaner.

akshayknarayan: The change is "awsDescriptions" -> "AwsDescriptions" (use the exported name), so it's not a functionless patch.

How much do you care about leaving the old style of multiline method declaration?

ejj: Ah, you should only have to change that one character though right?  I.E. the change of the ChooseSize wrapping is purely cosemetic (at least by my read).

As with the wrapping, I don't have anything in particular against this style, that said none of the rest of the code does it that way, and I really like consistency, so my preference would be to leave it.

akshayknarayan: Fair enough. I'll change it.

ejj:
Btw, all the commit messages should be wrapped to 70 characters.  This is somewhat of a historical oddity designed to make it easy to email patches around.

Also, I think several of the commit messages could benefit from having an explanatory paragraph under the title.  Looking back at them, they're going to be a bit mysterious.

akshayknarayan:
fixed commit messages.

Makefile
```diff
@@ -7,15 +7,13 @@ REPO = quilt
 DOCKER = docker
```
ejj: This file has a bunch of trailing white space in it.

inspect/graph.go
```diff
@@ -8,25 +8,65 @@ type node struct {
...
+	availability []availabilitySet
+	placement    map[string][]string
 }
 
 func makeGraph() graph {
 	return graph{
-		nodes:       map[string]node{},
-		connections: []connection{},
+		nodes: map[string]node{},
+		// one global availability set by default
```
ejj: I generally like comments to start with a capital letter and end with a period.

inspect/graph.go
```diff
@@ -8,25 +8,65 @@ type node struct {
...
 type connection struct {
-	from node
-	to   node
+	from string
+	to   string
 }
 
 // A graph represents permission to communicate across a series of nodes.
 // Each node is a container and each edge is permissions to
 // initiate a connection.
```
ejj: I think I might add a sentence or two about what an `availabilitySet` is.  It's not totally obvious and a comment would be helpful

inspect/graph.go
```diff
@@ -35,17 +75,24 @@ func (g *graph) getAddNode(label string) node {
...
-	toNode := g.getAddNode(to)
-	fromNode.connections[to] = toNode
-	g.connections = append(g.connections, connection{from: fromNode, to: toNode})
+func (g *graph) removeNode(label string) {
+	delete(g.nodes, label)
+
+	// delete edges to this node
+	for _, n := range g.getNodes() {
+		if _, ok := n.connections[label]; ok {
+			// copy node, don't include removed
```
ejj: `copy node`?   maybe rewrite this comment to be clearer?

inspect/invariant.go
```diff
@@ -0,0 +1,150 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+)
+
+// invariants are non-hierarchical statements
+type invariantType int
```
ejj: I think all this code would be a lot cleaner if instead of having a single `invariant` struct, you used interfaces instead.  i.e. you had an `invariant` interface that required there be a `form` function, and had two implementin structs, one for reach and one for between.  What do you think?

inspect/invariant.go
```diff
@@ -0,0 +1,150 @@
...
+	formImpls = map[invariantType]func(graph graph, inv invariant) bool{
+		reachInvariant:   reachImpl,
+		betweenInvariant: betweenImpl,
+	}
+}
+
+func check(graph graph, path string) ([]invariant, *invariant, error) {
+	invs, err := parseInvariants(graph, path)
+	if err != nil {
+		panic(err)
```
ejj: panic seems a bit draconian.  I generally like to only use them in cases that could only be caused by bugs in our code. I.e. somewhat like an assert.

Likely better to just log and return the error (or have the caller log).

inspect/invariant.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	return invs, nil, nil
+}
+
+// invariant format: <form> <target value ("true"/"false")> <node labels...>
+func parseInvariants(graph graph, path string) ([]invariant, error) {
+	var invs []invariant
+
+	var parse func(string)
```
ejj: Why not just give this it's own function?  I.E. something like
`func parseLine(graph graph, line string) invariant`.  Should be a bit more readable.  If you still need to wrap it in a little function to do the append, that would be fine I think.

inspect/invariant.go
akshayknarayan: Hm, this could be a cleaner way. I'll look into it.

akshayknarayan: I looked into it - I don't think it would be that much cleaner. Having the interface would save the code from having one map object (`formImpls`), but the boilerplate around that would be needlessly verbose imo.

This may make sense later, though - we should revisit it when integrating invariants into stitch.

Makefile
akshayknarayan: done.

inspect/graph.go
akshayknarayan: done.

akshayknarayan:
Alright, ready for next round.
I addressed many of @ejj 's comments (with the exception of having an invariant interface, let me know what you think there).
I also realized I merged many fixes for tian's comments into the wrong commit - fixed that.

inspect/main.go
```diff
@@ -54,14 +66,52 @@ func main() {
...
-			break
-		}
+	for _, pl := range spec.QueryPlacements() {
+		graph.addPlacementRule(pl)
 	}
 
-	graphviz(slug, graph, containerLabels)
+	ignoreNext := 0
+	foundFlags := map[string]argOption{}
+	func() {
```
ejj: Eek, this is really scary.

I think there's a couple of problems with it:

1) I think you're allowing people to say `inspect viz viz` which complicates the parsing logic quite a bit.  Instead, if you just allowed one command per invocation, this code could be quite a bit simpler.

2) Why is it wrapped in this function?

3) For the optional arguments the flag library should help: https://golang.org/pkg/flag/  I.E. `check` could be a flag instead of a positional argument.

ejj: Hmm, looking at the code, my point on (3) doesn't entirely make sense,

inspect/placement.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	return labels
+}
+
+func (avSet availabilitySet) str() string {
+	return fmt.Sprint(avSet.nodes())
+}
+
+func (avSet availabilitySet) check(label string) bool {
+	if _, ok := avSet[label]; !ok {
```
ejj:
```
_, ok := ...
return ok
```

inspect/placement.go
```diff
@@ -0,0 +1,173 @@
...
+		return false
+	}
+	return true
+}
+
+func (avSet availabilitySet) copyAvSet() availabilitySet {
```
ejj: Why not just call this function `copy`?  Should be clear from the context that the caller is copying an avSet.

inspect/placement.go
```diff
@@ -0,0 +1,173 @@
...
+	newAvSet := map[string]struct{}{}
+	for label := range avSet {
+		newAvSet[label] = struct{}{}
+	}
+	return newAvSet
+}
+
+func (avSet *availabilitySet) insert(labels ...string) {
```
ejj: Since an avSet is a map, which is a reference type, I _think_ you can just use `avSet availabilitySet` as the type for the next couple of functions.  Not entirely sure though.

inspect/placement.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// Merge all placement rules such that each label appears as a target only once
+func (g *graph) addPlacementRule(rule stitch.Placement) {
+	if !rule.Rule.Exclusive {
+		return
+	}
+
+	target, toSeparate := validateRule(rule, *g)
+	if existing, ok := g.placement[target]; ok {
```
ejj: I'm pretty sure you can ditch this if check and just do
```
g.placement[target] = append(toSeparate, g.placement[target]...)
```

If  `g.placement[target]` doesn't exist, then the lookup will return `nil`, which in go is technically the empty list, thus the append will be a noop and you'll end up with equivalent behavior.

inspect/placement.go
```diff
@@ -0,0 +1,173 @@
...
+		g.placement[target] = append(toSeparate, existing...)
+	} else {
+		g.placement[target] = toSeparate
+	}
+
+	for _, sep := range toSeparate {
+		if existing, ok := g.placement[sep]; ok {
```
ejj: Same principle here.

inspect/placement.go
```diff
@@ -0,0 +1,173 @@
...
+			g.placement[sep] = append(existing, target)
+		} else {
+			g.placement[sep] = []string{target}
+		}
+	}
+
+	// Add extra rule separating "public" into its own availability set.
+	// Remove once "implicit placement rules" are incorporated into dsl.
+	if _, found := g.nodes[stitch.PublicInternetLabel]; found {
```
ejj: s/found/ok

inspect/placement.go
```diff
@@ -0,0 +1,173 @@
...
+		if av.check(label) {
+			return av
+		}
+	}
+	return nil
+}
+
+// Finding minimal number of availability sets is NP-complete.
+// Try to pack into as few availability sets as possible.
+func (g *graph) placeNodes() {
```
ejj: Could you add a bit of an expanded comment on how the algorithm actually works?  It's a bit dense to figure out from just the code, will help future readers.

inspect/query.go
```diff
@@ -0,0 +1,126 @@
...
+func ask(graph graph, invs []invariant, path string) (*query, *invariant, error) {
+	queries, err := parseQueries(graph, path)
+	if err != nil {
+		panic(err)
+	}
+
+	return askQueries(graph, invs, queries)
+}
+
+func askQueries(
```
ejj: Meh.  Don't like the indent style, inconsistent with the rest of the code.

inspect/test/placement.spec
```diff
@@ -0,0 +1,16 @@
+(label "a" (docker "ubuntu"))
```
ejj: Thanks for writing unit tests btw.

ejj: Hmm, actually I don't think I see these being invoked anywhere . . .

Would be really nice if they were called as part of the build.  Should be trivial to add an inspect_test.go that does it.

ejj:
Ok, finally read through the first patch.  Had some comments, but nothing huge.  I think it will be ready next round.

inspect/placement.go
akshayknarayan: there's a built-in function called copy, I wasn't sure what the rules are.

inspect/query.go
akshayknarayan: sure, but I like this style better overall :)
Anyway, I'll change it...

inspect/test/placement.spec
akshayknarayan: I'd been running the tests manually. I'll look into expressing them in go unit test format.

ejj: Should be trivial, if it turns out to be hard let me know

akshayknarayan: added inspect_test.go

akshayknarayan: Travis reports failure because of the following:
`--- FAIL: TestSparkInvariants (0.00s)
panic: unable to open import spark [recovered]
	panic: unable to open import spark`

This doesn't happen on my local machine - any ideas?

inspect/main.go
akshayknarayan: I agree this could be cleaner, however, I don't think it's a big deal since this code will go away once invariants get integrated into stitch.

The function wrapping is to allow `query` to be evaluated in a `defer`, since the invariants need to be parsed first.

inspect/test/placement.spec
kklin: probably depends on `QUILT_PATH` being set

akshayknarayan: hm so `stitch/stitch_test.go` doesn't refer to `QUILT_PATH` - this shouldn't need anything more special than that, right?

kklin: None of the tests in `stitch_test.go` actually import anything on the filesystem. Anyways, you should be able to get it to pass by defining `QUILT_PATH` yourself instead of reading it from the environment variable.

That said, maybe it'd be a better test to just remove the dependency on the spark spec. For example, in `stitch_test.go` we define the specs inline. Don't know if you actually need the spark spec though, or would need to tweak your inspect tool.

akshayknarayan: ok, it works now. thanks!

akshayknarayan:
Alright, made several changes to address comments. The biggest thing I didn't address is the structure of `inspect/main.go` - let me know what you think there.

ejj:
Darn, was about to merge the first patch, but the unit tests don't build.  You must fix the error later in the series...

```
inspect/inspect_test.go:38: conn.From undefined (type int has no field or method From)
inspect/inspect_test.go:38: conn.To undefined (type int has no field or method To)
```

inspect/viz.go
```diff
@@ -8,11 +8,12 @@ import (
 	"github.com/NetSys/quilt/stitch"
 )
 
-func viz(configPath string, spec stitch.Stitch, graph graph) {
+func viz(configPath string, spec stitch.Stitch, graph graph, outputFormat string) {
 	var slug string
 	for i, ch := range configPath {
 		if ch == '.' {
 			slug = configPath[:i]
+			break
```
ejj: Shouldn't this be in the previous patch?

inspect/viz.go
```diff
@@ -61,7 +62,12 @@ func getImageNamesForLabel(containerLabels map[string][]*stitch.Container,
 
 // Graphviz generates a specification for the graphviz program that visualizes the
 // communication graph of a stitch.
-func graphviz(slug string, graph graph, containerLabels map[string][]*stitch.Container) {
+func graphviz(
```
ejj: ...

inspect/viz.go
```diff
@@ -91,9 +97,18 @@ func graphviz(slug string, graph graph, containerLabels map[string][]*stitch.Con
...
+	// - graphviz (install from your favorite package manager)
+	var writeGraph *exec.Cmd
+	switch outputFormat {
+	case "ascii":
+		writeGraph = exec.Command("graph-easy", "--input="+slug+".dot", "--as_ascii")
+	case "pdf":
+		writeGraph = exec.Command("dot", "-Tpdf", "-o", slug+".pdf", slug+".dot")
+	}
+	writeGraph.Stdout = os.Stdout
+	writeGraph.CombinedOutput()
```
ejj: Second patch looks good modulo the two nits.

cluster/provider/constraint.go
```diff
@@ -4,27 +4,33 @@ import (
...
+	CPU    int
+	Disk   string
+	Region string
 }
 
 // XXX: cache results?
-func pickBestSize(descriptions []description, ram stitch.Range, cpu stitch.Range,
-	maxPrice float64) string {
-	var best description
+func pickBestSize(
```
ejj: leave it

scripts/gce-descriptions
```diff
@@ -41,14 +41,15 @@ class GCEParser(object):
...
             print("\t{}".format(self.emitLine(row)))
         print("}")
 
     # Returns the line
     def emitLine(self, row):
-        line = "{{size: \"{}\", cpu: {}, ram: {}, price: {}}},".format(
+        line = "{{Size: \"{}\", CPU: {}, RAM: {}, Price: {}}},".format(
                 row[self.machineType], row[self.cpu], row[self.ram],
                 row[self.price])
         return line
```
ejj: Third patch looks good modulo the nit.

inspect/inspect_test.go
```diff
@@ -86,6 +90,15 @@ func TestPlacement(t *testing.T) {
 	}
 }
 
+func TestEnough(t *testing.T) {
+	specPath := "./test/placement.spec"
+	invariantPath := "./test/enough.inv"
+	// correct result: all invariants pass
```
ejj: Capitol `Correct`

inspect/invariant.go
```diff
@@ -12,6 +12,8 @@ const (
 	reachInvariant = iota
 	// On-pathness (between): three arguments, <from> <to> <between>
 	betweenInvariant
+	// schedulability (enough): zero arguments
```
ejj: Same

inspect/invariant.go
```diff
@@ -57,36 +61,49 @@ func checkInvariants(graph graph, invs []invariant) ([]invariant, *invariant, er
 
 func parseLine(graph graph, line string) (invariant, error) {
 	sp := strings.Split(line, " ")
+	var nodes []string
 
-	// Validate target argument.
+	// validate target argument
```
ejj: Just leave it =)

inspect/invariant.go
```diff
@@ -57,36 +61,49 @@ func checkInvariants(graph graph, invs []invariant) ([]invariant, *invariant, er
...
 
-	// Validate label arguments.
-	var nodes []string
-	for _, n := range sp[2:] {
-		if _, ok := graph.nodes[n]; !ok {
-			return invariant{}, fmt.Errorf("malformed assertion (unknown label): %s", n)
+		// validate label arguments
+		for _, n := range sp[2:] {
+			if _, ok := graph.nodes[n]; !ok {
+				panic(fmt.Errorf("malformed assertion (unknown label): %s", n))
```
ejj: I don't see why you'd do the fmt.Errorf here.  Maybe just fmt.Sprintf instead. (If you're going to panic)

akshayknarayan:
ok @ejj, pushed updates.
re: the test failing, the test passes on all commits for me - github is actually showing them in the wrong order, so that might be your issue.
Correct order is:
- inspect: Invariant checking and placement awareness
- inspect: Add option for ascii visualization
- stitch: Make stitch.QueryConnections return value consistent
- provider: Export provider machine types
- inspect: Add placement satisfiability invariant

akshayknarayan:
github is getting the commits wrong. will reopen as a new PR.

PR #444, opened 2016-06-10

ZhongtianWang:


ejj:
Tian could you rebase? This conflicts with master at the moment

ejj:
Also the commit message seems not to be word wrapped.  Should be wrapped to 70 characters.

ejj:
And, don't put issue numbers in the commit messages

provider/azure.go
```diff
@@ -122,111 +206,546 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+			wg.Done()
+		}(location)
+	}
+
+	go func() {
+		wg.Wait()
+		close(errs)
+	}()
+
+	select {
```
ejj: No need for this select statement.  You can just receive from the channel and it will block.

provider/azure.go
```diff
@@ -122,111 +206,546 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+				if _, err := clst.secRuleClient.CreateOrUpdate(resourceGroupName,
+					securityGroupName, *rule.Name, rule, cancel); err != nil {
+					errs <- err
+				}
+			}
+			wg.Done()
+		}(location)
+	}
+
+	go func() {
```
ejj: Why is this in a go routine?  I think I would reorganize this a tad.  Something like

```
wg.Wait()
select {
case err: <-errs:
default:
  return nil
}

File: provider/azure.go [hunk 3]
```diff
@@ -122,111 +206,546 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+
+			cloudRules = *result.Value
+
+			for _, rule := range cloudRules {
+				if *rule.Properties.Priority >= nextAvaiPriority {
+					nextAvaiPriority++
+				}
+			}
+
+			scoreFun := func(left, right interface{}) int {
```
ejj: Use a hash Join.

provider/azure.go
```diff
@@ -122,111 +206,546 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+		}
+		localRules = append(localRules, outboundRule)
+	}
+
+	var wg sync.WaitGroup
+	errs := make(chan error)
+	for location := range clst.secGroups {
+		wg.Add(1)
+
+		go func(location string) {
```
ejj: I might pull this into it's own function, it's a bit messy.

provider/azure.go
```diff
@@ -122,111 +206,546 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+			cloudRules := []network.SecurityRule{}
+			securityGroupName := "quiltnsg-" + location
+			result, err := clst.secRuleClient.List(resourceGroupName, securityGroupName)
+			if err != nil {
+				errs <- err
```
ejj: Two things wrong with this line, can you guess one's obvious, one is subtle.  Can you guess what they are?

provider/azure.go
```diff
@@ -122,111 +206,546 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+			}
+
+			nextAvaiPriority := int32(100)
```
ejj: Why does this start at 100?

provider/azure.go
```diff
@@ -122,111 +206,546 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+
+			cloudRules = *result.Value
+
+			for _, rule := range cloudRules {
+				if *rule.Properties.Priority >= nextAvaiPriority {
+					nextAvaiPriority++
```
ejj: What if this goes above 65k?  This approach feels fragile

provider/azure.go
```diff
@@ -122,111 +206,546 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+		}
+		outboundRule := network.SecurityRule{
+			Name:       &outboundName,
+			Properties: &outboundProperties,
+		}
+		localRules = append(localRules, outboundRule)
+	}
+
+	var wg sync.WaitGroup
+	errs := make(chan error)
```
ejj: You want a buffered channel.  Why?

provider/azure.go
```diff
@@ -122,111 +206,546 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
 
 func (clst *azureCluster) SetACLs(acls []string) error {
-	return nil
+	asterisk := "*"
+	cancel := make(chan struct{})
+
+	localRules := []network.SecurityRule{}
+	for _, acl := range acls {
+		// Azure does not allow `/` as security rule name. So we use `-` instead.
+		inboundName := strings.Replace(acl+"-in", "/", "-", -1)
```
ejj: This would be clearer if you just did the replace on `acl` and then append

provider/azure.go
ZhongtianWang: I added a comment. Azure's security rule priority level starts at 100

provider/azure.go
ZhongtianWang: ohhhh forgot to reply. The obvious one I think is to return after error... The subtle one is that I'm using an unbuffered channel. I'd really need a buffered one, since writing to a buffered channel doesn't block if there is room in the buffer.

ejj: Right!

ejj: Ah a third issue. You really only want to save one error . . . so you really want to do something like 
```
select {
case errs <- err:
default
}
```

i.e. if the channel is only full, don't bother putting stuff in it.

ZhongtianWang: Right! I've already done that, because you mentioned it in another comment as well. Thank you 😺

PR #445, opened 2016-06-10
zookeeper: add runner spec and change docker image source to dockerhub

secant:
Still need to test the docker image - will do when it finishes building.

### DIFF
--- a/specs/zookeeper/zookeeper-runner.spec
+++ b/specs/zookeeper/zookeeper-runner.spec
@@ -1,0 +1,18 @@
+(import "zookeeper")
+
+(define count 3)
+
+(define zoo (zookeeper.New "zookeeper" count))
+
+(define Namespace "CHANGE_ME")
+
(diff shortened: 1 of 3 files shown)


specs/zookeeper/zookeeper-runner.spec
```diff
@@ -0,0 +1,18 @@
+(import "zookeeper")
```
ejj: "Add" should be capitalized in the commit message.  Otherwise this patch looks good.

specs/zookeeper/zookeeper-runner.spec
```diff
@@ -0,0 +1,18 @@
...
+
+(define nWorker 3)
```
ejj: "nWorker" should be like "count" or something.  In Zookeeper they aren't really called workers.

ejj:
Merged

PR #454, opened 2016-06-15
specs: Implement an etcd specification

ZhongtianWang:
This patch implements a simple etcd specification that boots etcd containers in standalone mode. Right now the image is `zhongtian/etcd`. We should change it to `quilt/etcd` if everything works.


specs/etcd/Dockerfile
```diff
@@ -0,0 +1,11 @@
+From alpine:3.4
```
ejj: Why not base this on the official etcd docker container? https://quay.io/repository/coreos/etcd

specs/etcd/etcd.spec
```diff
@@ -0,0 +1,23 @@
+(import "labels")
+(import "strings")
+(import "log")
+
+(define image "zhongtian/etcd")
+
+(define (create prefix n)
+	(let ((labelNames (strings.Range prefix n))
```
ejj: This should be two spaces over instead of 4.

specs/etcd/etcd.spec
```diff
@@ -0,0 +1,23 @@
...
+        (etcdHosts (strings.Join (map labels.StrToHostname labelNames) ","))
```
ejj: So instead of using this labels.StrToHostname function.  To move the label map earlier (i.e. create ll of the labels) and then use the built in labelHosts function.  The StrToHostname function should basically be considered deprecated.

Btw, if you want an example of the proper way to do it, the spark spec basically does the right thing in this respect.

specs/etcd/etcd.spec
```diff
@@ -0,0 +1,23 @@
...
+        (etcdDockers (makeList n (docker image "run")))
+        (mapEnvs (lambda (c h) (setEnv c "HOST" (labels.StrToHostname h)))))
+    (map mapEnvs etcdDockers labelNames)
+    (setEnv etcdDockers "PEERS" etcdHosts)
+    (map label labelNames etcdDockers)))
+
+(define (link peers)
```
ejj: Ditch this function, just do the connect directly in the New function.

specs/etcd/etcd.spec
```diff
@@ -0,0 +1,23 @@
+(import "labels")
+(import "strings")
+(import "log")
+
+(define image "zhongtian/etcd")
+
+(define (create prefix n)
```
ejj: Come to think of it, I don't think this create function really makes sense.  I would just have a single function, New that has all this code in it.

specs/etcd/etcd.spec
```diff
@@ -0,0 +1,23 @@
...
+    (map mapEnvs etcdDockers labelNames)
+    (setEnv etcdDockers "PEERS" etcdHosts)
+    (map label labelNames etcdDockers)))
+
+(define (link peers)
+  (connect (list 1000 65535) peers peers))
+
+(define (New prefix n)
+	(let ((peers (create prefix n)))
+		(progn
```
ejj: You don't need this progn.  Let as an implicit progn built int.

specs/etcd/etcd.spec
```diff
@@ -0,0 +1,23 @@
...
+			(link peers)
+			(hmap ("peer" peers)))))
```
ejj: There's a weird trailing character here.

specs/etcd/example.spec
```diff
@@ -0,0 +1,19 @@
...
+// Defines the set of addresses that are allowed to access Quilt VMs.
+(define AdminACL (list "local"))
+
+(let ((cfg (list (provider "Amazon")
+		 (region "us-west-1")
+		 (size "m4.large")
+		 (diskSize 32)
+		 (githubKey "<YOUR_GITHUB_USERNAME>"))))
+     (makeList 1 (machine (role "Master") cfg))
+     (makeList nWorker (machine (role "Worker") cfg)))
```
ejj: Actually all of your files have this weird trailing character.

specs/etcd/example.spec
```diff
@@ -0,0 +1,19 @@
...
+// Using unique Namespaces will allow multiple Quilt instances to run on the
+// same cloud provider account without conflict.
+(define Namespace "CHANGE_ME")
+
+// Defines the set of addresses that are allowed to access Quilt VMs.
+(define AdminACL (list "local"))
+
+(let ((cfg (list (provider "Amazon")
+		 (region "us-west-1")
+		 (size "m4.large")
```
ejj: Specify the size in terms of ram and CPU . . . makes it easier to change the provider.

specs/etcd/example.spec
```diff
@@ -0,0 +1,19 @@
...
+
+// Using unique Namespaces will allow multiple Quilt instances to run on the
+// same cloud provider account without conflict.
+(define Namespace "CHANGE_ME")
+
+// Defines the set of addresses that are allowed to access Quilt VMs.
+(define AdminACL (list "local"))
+
+(let ((cfg (list (provider "Amazon")
+		 (region "us-west-1")
```
ejj: I'd ditch specifying the region, just use the Quilt default region, again makes it easier to specify the provider.

specs/etcd/run
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/python
```
ejj: You know, if you had written this in go, you could just build it and copy the binary into the container.  You wouldn't need to install python.  That said, if it's easy to install python, that's fine.

specs/etcd/etcd.spec
ZhongtianWang: Github wants to see an empty newline at the end of every file

ejj: That's a pretty standard requirement . . . we should probably oblige them.

ejj:
Merged

PR #456, opened 2016-06-15
dockerignore: create .dockerignore to ignore .git/

sirspinach:
This ignore shaves 17MB from the build context that is uploaded to the Docker
daemon every `docker build`. The time saved is currently marginal, but this is
good practice. (and .git/ will grow in size over time)


ejj:
Meh, by that logic we should be adding all the code, `Godeps` `vendor` etcd to the .dockerignore.  This leads to a lot of work for not really any benefit, I don't want to go down that path.

PR #459, opened 2016-06-18
Rename 'dsl' to 'stitch'

mmussomele:


ejj:
This patch looks great.  Only comment is that when I ran `make format` it came up with some changes.  Please me sure to run `make check lint` before submitting PRs in future.

PR #463, opened 2016-06-21
quilt: add a target to make lint which checks for lines more than 89 …

akshayknarayan:
…characters in length


ejj:
https://docs.travis-ci.com/user/languages/python#Specifying-Test-Script

I think you can just say something like
```
python:
   - "3.5"
```

akshayknarayan:
Ok, the build is failing, but that's because there are lines in the project over 89 chars long.

Makefile
```diff
@@ -1,6 +1,7 @@
 export GO15VENDOREXPERIMENT=1
 PACKAGES=$(shell GO15VENDOREXPERIMENT=1 go list ./... | grep -v vendor)
 NOVENDOR=$(shell find . -path ./vendor -prune -o -name '*.go' -print)
+SOURCE=$(shell find . -path ./vendor -prune -o -name '*.go' -print | grep -v 'constraintConstants.go' | grep -v 'gceConstants.go' | grep -v 'pb.pb' | grep -v 'cloud_config.go')
```
akshayknarayan: Is there a better way to exclude auto-generated files?

akshayknarayan:
Thanks @ZhongtianWang ! This is ready for review @sirspinach

Makefile
sirspinach: ``grep -Ev '(constraintConstants\.go|gceConstants\.go...)'```

sirspinach: You might want to escape the `.` too.

scripts/format-check.py
```diff
@@ -0,0 +1,32 @@
...
+            in zip(range(len(lines)), lines)
+            if len(l) > 89
+        ]
+        if (len(offendingLines) > 0):
+            print(
+                    "{} contains {} lines over 89 chars long".format(
+                        filepath,
+                        len(offendingLines)
+                    )
+                 )
```
sirspinach: not a suggestion, but I just wanted to say I like the way you format long lines. I'm not sure if this is the most readable way to do it, but it looks cleaner than any multiline print-format statement I've written.

scripts/format-check.py
```diff
@@ -0,0 +1,32 @@
...
+import os
+
+
+def checkFile(filepath):
+    with open(filepath, 'r') as f:
+        lines = f.readlines()
+        offendingLines = [
+            (i, l)
+            for i, l
+            in zip(range(len(lines)), lines)
```
sirspinach: I think you should consider using `enumerate(lines)`, because it does the same thing. 
https://docs.python.org/3.5/library/functions.html#enumerate

sirspinach: or even terser -- remove `lines = f.readlines()` and use `in enumerate(f.readlines())`

sirspinach:
I'm done reviewing.

Makefile
```diff
@@ -1,6 +1,8 @@
 export GO15VENDOREXPERIMENT=1
 PACKAGES=$(shell GO15VENDOREXPERIMENT=1 go list ./... | grep -v vendor)
 NOVENDOR=$(shell find . -path ./vendor -prune -o -name '*.go' -print)
+AUTOGEN=./cluster/provider/constraintConstants.go ./cluster/provider/gceConstants.go ./minion/pb/pb.pb.go ./cluster/provider/cloud_config.go ./minion/network/link_test.go
```
sirspinach: I'm surprised that this variable assignment works out correctly. Makefiles are probably different, but I doubt that this would pass in Bash. 

Can you explain to me what's going on in this line? (what's its function, and how it is evaluated by make)

akshayknarayan: variables in make are space separated, unquoted strings. So in this case, AUTOGEN is a variable that holds the names of some files which we don't want the line length checker to operate on.
on the next line, we take the list of all source files, $(NOVENDOR), and filter out the files we don't want to check - $(filter-out $(AUTOGEN),$(NOVENDOR)).

https://www.gnu.org/software/make/manual/make.html#Variables-Simplify

akshayknarayan:
Incorporated changes from @sirspinach 's review. Ready for more review.

ejj:
Couple of nits.

1) The commit message is too long.

2) The patch that fixes long lines should be before the patch that enforces the check.

scripts/format-check.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/python3
```
ejj: This script looks like it will run in either python2 or python3.  If that's the case could you just say `/usr/bin/python`?  python3 doesn't appear to be installed on mac by default, making this a somewhat annoying dependency

ejj:
Other than those 3 comments, patch looks great.  Let me know when the next version is up, and I'll merge.

akshayknarayan:
done.

ejj:
Merged

PR #464, opened 2016-06-24

ZhongtianWang:


ejj:
Assigned to vivian

Azure.md
```diff
@@ -0,0 +1,43 @@
...
+}
+~~~~
+
+It is worth noting that permissions around AD don't fall in line exactly 
+with permissions around the Azure subscription. You might run into issues 
+such as unable to view or create AD applications. So make sure that you are 
+a global admin of the directory to which you are part of. This happens often 
+if you are working on someone else's subscription and the subscription owner 
+added you as co-administrator in his/her subscription. To check, ask the 
+subscription owner to check with Azure classic portal, under Active Directory
```
secant: Active Directory > Directory > Users (the `-` makes a bullet in markdown)

cluster/provider/azure.go
```diff
@@ -123,30 +171,153 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
 		panic("boot count cannot be negative")
```
secant: Should this be an error instead?

ejj: It's kind of a silly check, `len` of a slice is never less than zero . . .   Not sure this is the right patch to fix it though, perhaps should be a separate patch after or before.

cluster/provider/azure.go
```diff
@@ -156,91 +327,531 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+	// Storage name needs to be globally unique, with a limit of 24 characters
+	storageName := randomString(24)
+	nameAvail := false
+	for !nameAvail {
+		cna, err := clst.storageClient.CheckNameAvailability(
+			storage.AccountCheckNameAvailabilityParameters{
+				Name: to.StringPtr(storageName),
+				Type: to.StringPtr("Microsoft.Storage/storageAccounts")})
+		if err != nil {
+			return storageAccount, err
```
secant: Not sure what our guideline is, but the past functions have been returning `nil` when running into an error

cluster/provider/azure.go
```diff
@@ -156,91 +327,531 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+
+	param := storage.AccountCreateParameters{
+		Location:   to.StringPtr(location),
+		Properties: &properties,
+	}
+
+	cancel := make(chan struct{})
+	_, err := clst.storageClient.Create(resourceGroupName, storageName, param, cancel)
+	if err != nil {
+		return storageAccount, err
```
secant: Same comment as previous comment

cluster/provider/azure.go
```diff
@@ -156,91 +327,531 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+	}
+
+	return clst.storageClient.GetProperties(resourceGroupName, storageName)
+}
+
+func (clst *azureCluster) configureSubnet(vnetName string) (network.Subnet, error) {
+	id := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/"+
+		"Microsoft.Network/virtualNetworks/%s/subnets/%s", clst.subscriptionID,
+		resourceGroupName, vnetName, subnetName)
```
secant: For formatting long lines, I recall [@sirspinach's comment](https://github.com/NetSys/quilt/pull/463#discussion_r67992161).

cluster/provider/azure.go
ZhongtianWang: We can't return nil as a struct. That's why when there is an error, it just returns an empty struct.

cluster/provider/azure.go
ZhongtianWang: I think he was mentioning about formatting lines in python :D. Also, this is not a multiline string.

cluster/provider/azure.go
ZhongtianWang: Oh just realized a problem by looking at the code. I did not re-generate storageName when name is not available... I will change it after the review.

cluster/provider/azure.go
```diff
@@ -1,116 +1,164 @@
...
+	clientID       string
+	clientSecret   string
+	tenantID       string
+	subscriptionID string
+
+	namespace string
 }
 
+// A SecurityRuleSlice allows for slices of Collections to be used in joins
+type SecurityRuleSlice []network.SecurityRule
```
ejj: Does this need to be public?

cluster/provider/azure.go
ejj: @ZhongtianWang are you going to address this?

cluster/provider/azure.go
```diff
@@ -123,30 +171,153 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+	if err != nil {
+		return err
+	}
+
+	virtualNetworks, err := clst.listVirtualNetworks()
+	if err != nil {
+		return err
+	}
+
+	// Map subnets to locations
```
ejj: This is really mapping locations to subnets.  Also comment could use a period at the end.

cluster/provider/azure.go
```diff
@@ -123,30 +171,153 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+	subnets := make(map[string]network.Subnet)
+
+	// A global unique storage account, a subscription unique virtual network and
+	// a subscription unique security group is needed for each location
```
ejj: Comment could use a period at the end.

cluster/provider/azure.go
```diff
@@ -123,30 +171,153 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+	set := make(map[string]struct{})
```
ejj: `set` isn't a great name for this variable.  It's not entirely what it's a set of . . . Storage accounts?  that's what the comment above seems to imply, hard to tell though.  I guess it's a set of regions?  Then why does the comment talk about storage accounts?

cluster/provider/azure.go
```diff
@@ -123,30 +171,153 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
-		name := "quilt-" + uuid.NewV4().String()
-		if err := clst.instanceNew(name, m.Size, m.Region,
-			cloudConfigUbuntu(m.SSHKeys, "wily")); err != nil {
+		if _, ok := set[m.Region]; ok {
+			continue
+		}
+		set[m.Region] = struct{}{}
+
+		if _, ok := storageAccounts[m.Region]; !ok {
+			storageAccount, err := clst.configureStorageAccount(m.Region)
```
ejj: So I think there's a bunch of problems with this.

1) It looks to me like we're creating new storage accounts *every time* Boot() is called.  Is that really the behavior you want?

2) I think we can do this way simpler.  What if we just said "in azure the storage account is always called quilt"  Then, once at boot, we could create the `quilt` account in every region in the Connection function. That seems a bit more robust and likely a whole lot simpler.

ejj: Would that work with azure, or does each VM need it's own storage account?

Btw, we could also incorporate the namesace name in the storage account.   For example `quilt-namespace`.  That way each user could have their own separate storage account.  That said, I don't think that's worth it unless we *absolutely have to*

ejj: Btw, that should also increase the boot time a lot.

Come to think of it, probably the best strategy is to create the storage account the *first time* we boot a VM in a particular region using a standard name, then future boots can take advantage of it.  You can use a global map to check if the account has been created already.

cluster/provider/azure.go
```diff
@@ -123,30 +171,153 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+			if err != nil {
+				return err
+			}
+			storageAccounts[m.Region] = storageAccount
+		}
+
+		vnetName := "quiltvnet-" + m.Region
```
ejj: Kind of a similar question . . . Does it make sense to create this virtual network *every time* we call Boot().  Couldn't we just do it once per region?  That should save us a lot of time and, potentially simplify the code a bit.

Does the vnet Name need to be globally unique btw?  Can we just call it quiltvnet?

cluster/provider/azure.go
```diff
@@ -123,30 +171,153 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+			virtualNetwork, err := clst.configureVirtualNetwork(vnetName, m.Region,
+				subnet)
+			if err != nil {
+				return err
+			}
+			virtualNetworks[m.Region] = virtualNetwork
+		}
+
+		if _, ok := securityGroups[m.Region]; !ok {
+			securityGroupName := "quiltnsg-" + m.Region
```
ejj: What if there are multiple people using azure with different namespaces?   Should they be sharing a security group, or should it be separate?  Same question for the vnet etc.  Not sure of the answer either way, what do you think?

cluster/provider/azure.go
ZhongtianWang: I agree with Vivian. An error makes more sense.

cluster/provider/azure.go
ZhongtianWang: So set is set of regions. Because for each region, we need a separate storage account, security group and virtual network. Using a set to ensure that we are only configuring each region once.

cluster/provider/azure.go
ZhongtianWang: So currently, it does only create new storage accounts the first time we boot a VM in a particular region. That's why each time we call Boot(), it first lists all the storage accounts, security groups and virtual networks, and we don't create anything new if there already exists one at a particular region.

cluster/provider/azure.go
ejj: Can `len` of a slice ever be negative?

cluster/provider/azure.go
ZhongtianWang: The annoying part for storage account is that, its name must be globally unique across entire Azure, and it must have one storage account per region. Plus, it doesn't really matter whether we need to specify `namespace` in storage account. All users could just share one storage account at a region, because the whole purpose of storage account is pretty much only about specifying redundancy and replication.

cluster/provider/azure.go
ZhongtianWang: It is once per region. We can't just call it quiltvnet because the name has to be unique inside a resource group.

cluster/provider/azure.go
ejj: Ah that's right.

More I think about it though, what if we just made the user specify a storage account in the config file.  Really, it could be a map from region to storage account so, if they want multiple regions, that would work it.

It just seems to me like Azure really doesn't want us automating the creation of these things, and what we've ended up with here is quite fragile.  Would make a lot more sense for us to just manually create "quilt-us-west" or whatever it would be called.  What do you think?

cluster/provider/azure.go
ZhongtianWang: Thank you for bringing up that point. We should make everything separate for different namespace, except maybe for storage account. Because Azure doesn't allow concurrent operations on same resources. If there are two people trying to add security rules at the same security group, there will be errors. Plus we don't want two people share ACLs

cluster/provider/azure.go
ZhongtianWang: I think the creation of storage account could be automated the way we are doing it right now. It first check if there already exists a storage account for the region inside the resource group. If not, it randomly generates name until we found an available one. I don't think storage account name is something user should care about. Plus, let's say if user specify they want an account called "quilt-us-west", but it turns out to be already taken by someone else. The user would have to go back to the spec and make up some other names. It doesn't sound very desirable. What do you think?

cluster/provider/azure.go
```diff
@@ -123,30 +171,153 @@ func (clst *azureCluster) Boot(bootSet []Machine) error {
...
+				m.Region)
+			if err != nil {
+				return err
+			}
+			securityGroups[m.Region] = securityGroup
+		}
+	}
+
+	// Boot all VMs concurrently
+	var wg sync.WaitGroup
```
ejj: You have this pattern in three places in the code.  I think I'd make a helper function.  Something like
```
func forEachMachine(func (Machine) err, machines []Machine)
```

You could even think about generalizing it to take a list of interfaces for the security group case.

cluster/provider/azure.go
ejj: So theres two issues with this.

1) When you're listing storage accounts, how do you know you're getting the right one?  What if I'd like a separate storage account for my quilt stuff than my production stuff?  How do I prevent you from using my production storage account?

2) What if I want you to reuse an existing storage account in some cases.  I.E. I'm running Spark in Quilt, and all my data lives in blob storage in a particular account.  How do I tell quilt to use that account?

3) This code just feels fragile . . . I'm not confident in it.

ZhongtianWang: Good point! Then I will just add tag to every resources, so every resources are associated to its namespace only.

ZhongtianWang: In terms of the second point, I'm not quite sure what do you mean. When we run spark on a vm, the data is at its related storage account, in that region with the correct namespace. I think as long as there is only one storage account in the quilt resource group, with the correct namespace and region, there shouldn't be any problems

ZhongtianWang: Ah that brings up another point. If there are multiple users inside one resource group, they could end up trying to get the same name for vnet or sec group, etc... Then I am probably just move everything to tag, and generate name randomly.

ejj: What if I have a separate non-quilt spark instance that I set up on Azure the old fashioned way.  Or I'm generally an azure user and I have some data floating around somewhere in azure that's managed by Quilt.  How do I boot up Azure vms on Quilt that access the data?

ejj: Also, to your previous point.  Perhaps we shouldn't be generating the resource group either.  If, instead, a user passed in the resource group as part of their azure credentials, then we could just use the namespace as the vnet name etc.  Seems much simpler and much less error prone.  Users would just great a storage account and a resource group at the same time they get their initial credentials, and then things work.

ZhongtianWang: Okay. Now I do think in terms of potentially having to reuse an old disk on Azure, whether it's an os disk or it's a data disk, we would need to let user specify:
1. Storage account
2. Name of the disk that we need to reuse
Maybe we can let them do it in the spec? Because we need to let them specify `boot this vm with the osdisk names XXX under the specified storage account`, and it sounds too specific to do it in the credential file...

ZhongtianWang: Then we should also let user specify resource group name and location as well. It would be cool if users could specify all those information in the spec, like creating a boot arguments.

ZhongtianWang: Any thoughts @ejj?

ejj: Sure in the long term all that stuff makes sense.  I guess the immediate question is what we should do with this patch.  I don't think we need to go all the way to making everything configurable immediately.  Did you find out what other people are doing on the storage account stuff?  Are they making users specify it or doing it manually?

ZhongtianWang: Both. I see many people make user specify storage names. Still, there are a small number of people generate names. For example, `uuidrg=$(uuidgen); storageAccountName="sa${uuidrg:0:20}"`. Also, in an Azure template, `"storageAccountName": "[concat(uniquestring(resourceGroup().id),'sa')]"`. There uniquestring is a powershell function that generate random unique string within the resource group.

ZhongtianWang: After all, I think the current approach should be fine for now, since it's easier for people to use. Quilt just generate everything for the user. In later patches we can make it fully configurable.

ZhongtianWang:
Ready for further reviews @ejj.

Here are the updates:
- Rewrote some of the logics so that there is complete separation of different environments. E.g. different storage accounts, security groups, virtual networks for different namespaces.
- Generalized code for concurrent execution on booting, stopping the machines and syncing ACLs.
- Fixed some confusing comments.
- Changed bunch of global variables to constants.

and many small problems in general    : >

ejj:
Could you rebase on master too?

ZhongtianWang:
Updated:
- Re-organized `forEach`, make such function calls look way cleaner by using `forEachMachine` and `forEachSecurityGroup`
- Separated updating inbound and outbound security rule

cluster/provider/azure.go
```diff
@@ -156,91 +294,582 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+	for _, r := range deleteList {
+		rule := r.(network.SecurityRule)
+		if _, err := clst.secRuleClient.Delete(resourceGroupName,
+			securityGroupName, *rule.Name, cancel); err != nil {
+			return err
+		}
+	}
+
+	// Each security rule is required to be assigned one unique priority number
+	// Between 100 and 4096
```
ZhongtianWang: oops, forgot `.`

PR #466, opened 2016-06-25
quilt: Simplify QUILT_PATH

secant:
- Introduces `quilt get`, which allows the user to download specs from other repositories, e.g. `quilt get github.com/secant/specs`
- Changes `quilt spec` to `quilt run spec`
- Makes `QUILT_PATH` only one directory, that lays its own structure out like `GOPATH`
- Modifies all current specs' import statements


util/get.go
```diff
@@ -0,0 +1,45 @@
+package util
+
+import (
+	"errors"
+	"fmt"
+	"golang.org/x/tools/go/vcs"
```
ejj: Nice, is this what go get uses?

ejj:
I skimmed this over.  At a high level, it looks really good.  *much much* cleaner than I expected.

Couple of comments:

In the second patch, their shouldn't be a period in the first line of the commit message.

Also, it looks to me a bit like the second patch implements some functionality, but breaks some things in the code base, that you fix up in the subsequent patches.  Is that the case?  If so, those later patches should be folded into the first one.  No patch in the history should break anything (knowingly), even if later, it's fixed up in the same commit.

Anyways, this wasn't a formal review, just skimmed it.  Someone else (will do it).

ejj:
Also, what's the "in progress" bit?  Just the unit tests?

ejj:
Oh, the documentation too I suppose

specs/spark/sparkPI.spec
```diff
@@ -17,7 +16,7 @@
 
 // Using unique Namespaces will allow multiple Quilt instances to run on the
 // same cloud provider account without conflict.
-(define Namespace "CHANGE_ME")
+(define Namespace "vivian-spark")
```
nlsun: Remember to change this back.

specs/spark/sparkPI.spec
```diff
@@ -26,6 +25,6 @@
 		 (region "us-west-1")
 		 (size "m4.2xlarge")
 		 (diskSize 32)
-		 (githubKey "<YOUR_GITHUB_USERNAME>"))))
+		 (githubKey "secant"))))
```
nlsun: Change this back too. You can also do `git add --patch` to pick out which parts of the file to add.

util/get.go
```diff
@@ -0,0 +1,45 @@
...
+
+	quiltPath := os.Getenv("QUILT_PATH")
+	if quiltPath == "" {
+		return errors.New("your QUILT_PATH environment variable needs to be defined")
+	}
+
+	path := filepath.Join(quiltPath, repo.Root)
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		fmt.Printf("Cloning %s into %s", repo.Root, path)
```
nlsun: Use `log.Info()` instead of `fmt` for printing.

ejj:
Btw, this *doesn't matter at all*, but it's common if you're posting a patch that you don't think should be merged yet, to prepend the title with `[RFC]` as in "Request For Comment".  Commonly used on Linux and a bunch of other projects.

secant:
@ejj Gotcha! I added the WIP because I got way too excited to make a PR, and forgot I had to write tests 😝

util/get.go
secant: Yup!

docs/GettingStarted.md
```diff
@@ -26,6 +26,18 @@ Clone the repository into your Go workspace: `go get github.com/NetSys/quilt`.
...
+## Set up QUILT_PATH
+Your `QUILT_PATH` will be where `quilt` looks for imported specs, and where 
+specs you download with `quilt get` get placed. You can set this to be anywhere;
+we recommend that you set it to be the `quilt/specs` folder. To do this, first
+find the current directory quilt is installed in, by `cd`-ing to `quilt` and 
+executing `pwd`. Copy and paste the output and run the following:
+```
+export QUILT_PATH=<PATH_TO_QUILT>/specs
+```
+It would be a good idea to add this to your `.bashrc` so that it does not have
```
mmussomele: maybe add something along the lines of "(or equivalent shell config file)"

quilt.go
```diff
@@ -31,7 +31,7 @@ func main() {
 	log.SetFormatter(util.Formatter{})
 
 	flag.Usage = func() {
-		fmt.Println("Usage: quilt [stitch | stop [namespace]]" +
+		fmt.Println("Usage: quilt [<stitch> | stop <namespace> | get <import_path>]" +
```
mmussomele: Should be `"Usage: quilt [run <stitch> | stop <namespace> | get <import_path>]"`

quilt.go
```diff
@@ -59,13 +59,23 @@ func main() {
...
 	default:
 		go configLoop(conn, flag.Arg(0))
 	}
 
 	cluster.Run(conn)
 }
 
+func getSpec(importPath string) {
+	err := util.GetSpec(importPath)
+	if err != nil {
```
mmussomele: Since `util.GetSpec` only returns an error, you could do this a little more cleanly with 
```if err := util.GetSpec(importPath); err != nil {
         ...
 }```

Comment (mmussomele) [inline quilt.go]: Although now that I think about it, since `util.GetSpec` does something meaningful (besides a condition check), maybe it deserves its own line. A second opinion would be nice.

File: util/get.go [hunk 8]
```diff
@@ -0,0 +1,41 @@
...
+	if quiltPath == "" {
+		return errors.New("your QUILT_PATH environment variable needs to be defined")
+	}
+
+	path := filepath.Join(quiltPath, repo.Root)
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		log.Info(fmt.Sprintf("Cloning %s into %s", repo.Root, path))
+		err = repo.VCS.Create(path, repo.Repo)
+		if err != nil {
```
mmussomele: See comments on quilt.go line 73 above. Whatever is decided up there should be consistently reflected here.

ejj:
Your coverage could be even better I think:

https://coveralls.io/builds/6791437/source?filename=util%2Fget.go#L29

util/get.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	quiltPath := os.Getenv("QUILT_PATH")
+	if quiltPath == "" {
+		return errors.New("your QUILT_PATH environment variable needs to be defined")
+	}
+
+	path := filepath.Join(quiltPath, repo.Root)
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
mmussomele: Do we need to consider cases where the user doesn't have write permissions? It seems like in the end that would cause `os.IsNotExist(err)` to return false, so you would try to download it, and then that would fail and the error from this line would be returned, which is probably desired, but I'm wondering if that should be more explicit.

quilt.go
```diff
@@ -59,13 +59,23 @@ func main() {
...
+		log.Error(err)
+	}
+	os.Exit(1)
```
mmussomele: Typically a non-ero exit code indicates failure. This should be `os.Exit(0)`

util/util_test.go
```diff
@@ -84,3 +85,32 @@ func ed(a, b []string, exp int) string {
...
+		t.Error(err)
+	}
+
+	importPath = "invalid"
+	expectedErr = "unrecognized import path \"invalid\""
+	err = GetSpec(importPath)
+	compareError(t, expectedErr, err)
+}
+
+func compareError(t *testing.T, expected string, actual error) {
```
mmussomele: You can get the String from an error with `err.Error()`, so this function might not be needed/does unnecessary work.

util/util_test.go
```diff
@@ -84,3 +85,32 @@ func ed(a, b []string, exp int) string {
...
+	if fmt.Sprintf("%s", actual) != expected {
```
mmussomele: You can replace `fmt.Sprintf("%s", actual)` with `actual.Error()`

stitch/imports.go
```diff
@@ -37,36 +39,36 @@ func resolveImportsRec(asts []ast, paths, imported []string) ([]ast, error) {
...
-			f, err := util.Open(modulePath)
-			if err == nil {
-				defer f.Close()
-				sc.Filename = modulePath
-				sc.Init(bufio.NewReader(f))
-				break
-			}
+		sc.Filename = modulePath
+		if _, err := os.Stat(modulePath); os.IsNotExist(err) {
+			err = util.GetSpec(name)
```
mmussomele: What if this fails? Should there be another check?

stitch/imports.go
```diff
@@ -37,36 +39,36 @@ func resolveImportsRec(asts []ast, paths, imported []string) ([]ast, error) {
...
+		sc.Init(bufio.NewReader(f))
 		parsed, err := parse(sc)
 		if err != nil {
 			return nil, err
 		}
 
-		parsed, err = resolveImportsRec(parsed, paths, append(imported, name))
+		// Rename module name to last name in import path
+		folders := strings.Split(name, "/")
+		name = folders[len(folders)-1]
```
mmussomele: These two lines could be done more cleanly (and platform independently) with [filepath.Base](https://golang.org/pkg/path/filepath/#Base).

quilt.go
```diff
@@ -57,12 +57,24 @@ func main() {
...
+		if len(flag.Arg(1)) == 0 {
+			usage()
+		}
+
 		stop(conn, flag.Arg(1))
 	case "get":
+		if len(flag.Arg(1)) == 0 {
+			usage()
+		}
+
```
mmussomele: I think the blank lines on 64, 70, 76 look weird, but maybe I'm just getting picky.

ejj: Picky, I think they look fine.

ejj: Though, you're doing this flag.Arg(1) check a bit redundantly.  You could pull it out of the switch

stitch/imports.go
secant: Ah good catch, I don't have to grab the error at all 😛

util/util_test.go
```diff
@@ -84,3 +85,32 @@ func ed(a, b []string, exp int) string {
...
+	// Update
+	if err := GetSpec(importPath); err != nil {
+		t.Error(err)
+	}
+
+	importPath = "invalid"
+	expectedErr = "unrecognized import path \"invalid\""
+	err = GetSpec(importPath)
+	compareError(t, expectedErr, err)
+}
```
secant: @ejj I'm confused on why the Cloning portion of my test isn't being covered - it should get triggered in line 98 or line 107

ejj: Could be an issue with coveralls.  Does it show up in the html if you do make coverage locally?

If not, perhaps you have a bug?

ejj:
This case is probably worth testing: https://coveralls.io/builds/6823738/source?filename=util%2Fget.go#L29

secant:
That case is covered locally :(
[image](http://i.imgur.com/YMB0jIn.png)

ejj:
Sigh, coveralls is crappier than I'd hoped . . .

secant:
Ready for more review!

docs/GettingStarted.md
```diff
@@ -7,7 +7,7 @@ Quilt supports Go version 1.5 or later.
 
 Find Go using your package manager or on the [Golang website] (https://golang.org/doc/install).
 
-### Setup GOPATH
+### Set up GOPATH
```
ejj: This change (and the next one) don't really have anything to do with `quilt get`.  Should probably be in their own patch later or before.

docs/GettingStarted.md
```diff
@@ -17,15 +17,27 @@ your terminal to set up your GOPATH.
...
 ## Download and Install Quilt
 Clone the repository into your Go workspace: `go get github.com/NetSys/quilt`.
 
 This command also automatically installs Quilt. If the installation was
 successful, then the `quilt` command should execute successfully in your shell.
 
+## Set up QUILT_PATH
+Your `QUILT_PATH` will be where `quilt` looks for imported specs, and where 
+specs you download with `quilt get` get placed. You can set this to be anywhere;
+we recommend that you set it to be the `quilt/specs` folder. To do this, first
```
ejj: I actually don't like this as the initial recommendation.  The issue is, that as they download more specs, they're going to end up in their repository rather than in a reasonable quilt path location.  I think we should change the README to recommend that they set the QUILT_PATH to `~/.quilt` and then that the `quilt get` whatever it is we're using as our example application.

docs/GettingStarted.md
```diff
@@ -85,7 +97,7 @@ change `(define Namespace "CHANGE_ME")` to a different name.
 
 ### Deploying [specs/example.spec](../specs/example.spec)
 While in the `$GOPATH/src/github.com/NetSys/quilt/` directory, execute `quilt
-specs/example.spec`. Quilt will set up several Ubuntu VMs on your cloud
+run specs/example.spec`. Quilt will set up several Ubuntu VMs on your cloud
```
ejj: Different Patch?

docs/Language.md
```diff
@@ -135,34 +135,51 @@ binds and labels that start with a capital letter are exported.
...
-where the module name is the filename (minus the .spec extension), and the module
-body is the contents of the body.
+`import` is a way of importing code in other files. Specs are imported relative 
+to the location of `QUILT_PATH` environment variable. It evaluates to a `module`
+where the module name is the name after the last slash (minus the .spec 
+extension), and the module body is the contents of that spec file. For example,
 
 ```
-// math.spec
+// $QUILT_PATH/math.spec
```
ejj: What if you did `github.com/NetSys/quilt/math`?

docs/Language.md
```diff
@@ -135,34 +135,51 @@ binds and labels that start with a capital letter are exported.
...
 ### QUILT_PATH
 `quilt` looks for imports according to the `QUILT_PATH` environment variable.
-So if you have a spec that imports `stdlib`, and `stdlib.spec` is located at
-`specs/stdlib.spec`, then your `QUILT_PATH` should be `QUILT_PATH="specs"`.
-
-Multiple paths are separated by colons, so you may do `QUILT_PATH="specs:specs/spark"`.
+This variable should only be one directory, much like `GOPATH`. For example, if 
+your `QUILT_PATH="<PATH_TO_QUILT>/specs"`, and you have a spec that imports 
+`stdlib`, and `stdlib.spec` is located at 
+`<PATH_TO_QUILT>/specs/util/stdlib.spec`, then you should
```
ejj: I think we should use something in the github.com/NetSys/quilt in this example too.

docs/Language.md
```diff
@@ -135,34 +135,51 @@ binds and labels that start with a capital letter are exported.
...
-QUILT_PATH="specs" ./quilt config.spec
+QUILT_PATH="<PATH_TO_QUILT>/specs" ./quilt config.spec
 ```
 
 or `export` it into your environment.
 
+### quilt get
+If you would like to download a spec from another repository, execute
+`quilt get <IMPORT_PATH>`, where `<IMPORT_PATH>` is a path to a repository
+containing specs (e.g. `github.com/secant/specs`). `quilt` will clone the
```
ejj: Ditch secan for Netsys

docs/GettingStarted.md
```diff
@@ -7,7 +7,7 @@ Quilt supports Go version 1.5 or later.
```
ejj: This patch adds a bunch of trailing whitespace.

secant: I think the trailing whitespace is gone now.

secant:
Changes have been made!

docs/GettingStarted.md
ejj: Hmm, nope I'm still seeing it in the first commit GettingStarted.md

docs/GettingStarted.md
```diff
@@ -26,6 +26,21 @@ Clone the repository into your Go workspace: `go get github.com/NetSys/quilt`.
 This command also automatically installs Quilt. If the installation was
 successful, then the `quilt` command should execute successfully in your shell.
 
+## Set up QUILT_PATH
+Your `QUILT_PATH` will be where Quilt looks for imported specs, and where 
+specs you download with `quilt get` get placed. You can set this to be anywhere;
+we recommend that you set it to be the `~/.quilt` folder. To do this, first
+execute 
+```bash
+mkdir ~/.quilt
```
ejj: I'd be really nice of  `quilt` actually made the directory if it didn't exist.  That way there's even less setup for new users.

docs/Language.md
```diff
@@ -135,34 +135,51 @@ binds and labels that start with a capital letter are exported.
...
 ```
-// math.spec
+// $QUILT_PATH/NetSys/quilt/math.spec
 (define Square (lambda (x) (* x x)))
 
 
-// main.spec
-(import "math")
+// $QUILT_PATH/main.spec
+(import "NetSys/quilt/math")
```
ejj: Should this be `github.com/Netsys/quilt/math`?  Similar comment for above.

quilt.go
```diff
@@ -54,18 +54,32 @@ func main() {
...
 
 	conn := db.New()
-	switch flag.Arg(0) {
-	case "":
+	if len(flag.Args()) != 2 {
 		usage()
+	}
+
+	switch flag.Arg(0) {
+	case "run":
```
ejj: Shouldn't this change be in the patch that adds the run command?  Maybe it would be easier if you moved that patch first?

specs/spark/README.md
```diff
@@ -40,7 +40,7 @@ If you are sharing a computing cluster with others, it would be a good idea to
 change `(define Namespace "CHANGE_ME")` to a different name.
 
 ### Build `sparkPI.spec`
-Execute `quilt $GOPATH/github.com/NetSys/Quilt/specs/spark/sparkPI.spec` to
+Execute `quilt run $GOPATH/github.com/NetSys/Quilt/specs/spark/sparkPI.spec` to
```
ejj: Same here ... wrong patch right?

util/get.go
```diff
@@ -0,0 +1,40 @@
...
+		log.Info(fmt.Sprintf("Cloning %s into %s", repo.Root, path))
+		if err := repo.VCS.Create(path, repo.Repo); err != nil {
+			return err
+		}
+	} else {
+		log.Info(fmt.Sprintf("Updating %s", repo.Root))
+		repo.VCS.Download(path)
+	}
+
+	log.Info("Done!")
```
ejj: Ditch this log message.

util/util_test.go
```diff
@@ -84,3 +85,32 @@ func ed(a, b []string, exp int) string {
...
+
+func compareError(t *testing.T, expected string, actual error) {
```
ejj: I actually would ditch this function and copy and paste the code.  Two reasons:

1) The function doesn't save all that much code

2) When there's a test failure, it's going to say it was line 114, you'd much rather have it say it was line 109 (or wherever the error was actually triggered).  Makes debugging a bit easier.

util/util_test.go
```diff
@@ -84,3 +85,32 @@ func ed(a, b []string, exp int) string {
...
+	if actual.Error() != expected {
+		t.Errorf("expected %s \n but got %s", expected, actual)
+	}
+}
```
ejj: Done with review of the first patch, basically looks good.

ejj:
Also, note, looks like the build failed.  Line too long.

specs/spark/sparkPI.spec
```diff
@@ -26,6 +25,6 @@
 		 (region "us-west-1")
 		 (size "m4.2xlarge")
 		 (diskSize 32)
-		 (githubKey "<YOUR_GITHUB_USERNAME>"))))
+		 (githubKey "<YOUR_GITHUB_KEY>"))))
```
ejj: Not the patch for this change.

stitch/imports.go
```diff
@@ -37,36 +38,35 @@ func resolveImportsRec(asts []ast, paths, imported []string) ([]ast, error) {
...
-			f, err := util.Open(modulePath)
-			if err == nil {
-				defer f.Close()
-				sc.Filename = modulePath
-				sc.Init(bufio.NewReader(f))
-				break
-			}
+		sc.Filename = modulePath
+		if _, err := os.Stat(modulePath); os.IsNotExist(err) {
+			util.GetSpec(name)
```
ejj: So this I actually don't like.  I don't think we should automatically get the spec by default from the run command.  Instead, I think that if a spec is passed to the get command, at that point, we would get all of the the specs that it imports similar to how the go-get command works.  Probably the best way to implement that is some sort of flag passed to resolve imports.  You may have to execute the entire stitch, not sure.

stitch/imports.go
```diff
@@ -37,36 +38,35 @@ func resolveImportsRec(asts []ast, paths, imported []string) ([]ast, error) {
...
+		name = filepath.Base(name)
+		parsed, err = resolveImportsRec(parsed, path, append(imported, name))
 		if err != nil {
 			return nil, err
 		}
 
 		module := astModule{body: parsed, moduleName: astString(name)}
 		newAsts = append(newAsts, module)
 	}
-
```
ejj: Useless change.

stitch/stitch_test.go
```diff
@@ -1418,12 +1418,12 @@ func runtimeErrImport(t *testing.T, code, expectedErr string, path []string) {
...
+		t.Errorf("%s: got %s but expected %s", code, err, expectedErr)
 		return
 	}
 }
 
-func importErr(t *testing.T, code, expectedErr string, path []string) {
+func importErr(t *testing.T, code, expectedErr string, path string) {
 	var sc scanner.Scanner
 	prog, err := parse(*sc.Init(strings.NewReader(code)))
 	if err != nil {
```
ejj: Overall, the second patch looks good . . . though there's the one major comment.

quilt.go
```diff
@@ -31,8 +31,8 @@ func main() {
 	log.SetFormatter(util.Formatter{})
 
 	flag.Usage = func() {
-		fmt.Println("Usage: quilt [<stitch> | stop <namespace> | get <import_path>]" +
-			" [-log-level=level | -l=level]")
+		fmt.Println("Usage: quilt [run <stitch> | stop <namespace> | get <import_path>]" +
+			" [-log-level=<level> | -l=<level>]")
```
ejj: Yeah, you screwed up the rebasing.  This patch just changes usage, but doesn't actually make the change . . .  Did you do `git show` on each patch before submitting the PR?

secant:
Changes made, should be ready now. @ejj

quilt.go
```diff
@@ -114,8 +114,11 @@ func updateConfig(conn db.Conn, configPath string) error {
 		},
 	}
 	pathStr, _ := os.LookupEnv(quiltPath)
-	pathSlice := strings.Split(pathStr, ":")
-	spec, err := stitch.New(*sc.Init(bufio.NewReader(f)), pathSlice)
+	if pathStr == "" {
+		log.Warn("QUILT_PATH environment variable is not defined")
```
ejj: I think even this is a bit draconian.  Maybe just a Debug level message.  A lot of people will leave it undefined by default.

stitch/get.go
```diff
@@ -0,0 +1,83 @@
...
+// associated with that repoName.
+func GetSpec(repoName string, test ...bool) error {
+	testing := false
+	if len(test) > 0 {
+		testing = test[0]
+	}
+
+	var repo *vcs.RepoRoot
+	var err error
+	if repo, err = vcs.RepoRootForImportPath(repoName, true); err != nil {
```
ejj: This would be cleaner if you just broke the `err != nil` check into it's own if statement.  That way you could use a `:=` and ditch the variable declarations.

stitch/get.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	return path, nil
+}
+
+func resolveSpecImports(folder string) error {
+	return filepath.Walk(folder, checkSpec)
+}
+
+func checkSpec(file string, info os.FileInfo, err error) error {
+	if filepath.Ext(file) == ".spec" {
```
ejj: So in general, in go style, it's prefer to have less code in indent blocks.  I.E. check if its `!=` and return nil.

I also think it's easier to read that way.

stitch/get.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/NetSys/quilt/util"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// DefaultQuiltPath returns the default absolute QUILT_PATH, which is ~/.quilt.
+func DefaultQuiltPath() string {
+	quiltPath := os.Getenv("QUILT_PATH")
+	if quiltPath == "" {
```
ejj:
```
if quiltPath != "" {
    return quiltPath
}
...
```

stitch/stitch.go
```diff
@@ -77,14 +77,15 @@ func (stitchr Range) Accepts(x float64) bool {
 	return stitchr.Min <= x && (stitchr.Max == 0 || x <= stitchr.Max)
 }
 
-// New parses and executes a stitch (in text form), and returns an abstract Stitch handle.
-func New(sc scanner.Scanner, path string) (Stitch, error) {
+// New parses and executes a stitch (in text form), and returns an abstract Stitch
```
ejj: Why the comment change?

stitch/stitch_test.go
```diff
@@ -1540,3 +1543,58 @@ func codeEq(a, b string) bool {
...
+	util.AppFs.Mkdir("nextimport", 777)
+	util.WriteFile("nextimport/nextimport.spec", []byte("(define dummy 1)"), 0644)
+	os.Setenv("QUILT_PATH", ".")
+	err := checkSpec("test/imp.spec", nil, nil)
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestResolveSpecImports(t *testing.T) {
```
ejj: So I think the one issue with these tests, is they don't actually verify the behavior of the code is correct.  They just verify that the code runs without panicking or returning an error.  I think it'd be worth actually verifying the behavior.

stitch/stitch_test.go
```diff
@@ -1540,3 +1543,58 @@ func codeEq(a, b string) bool {
 	b = strings.TrimSpace(codeEqRE.ReplaceAllString(b, " "))
 	return a == b
 }
+
```
ejj: Do the unit tests implemented in this file hit the network btw?  If so, we should probably tweak them.  Let me know and I can make some suggestions.

stitch/stitch.go
secant: `abstract Stitch handle.` exceeds 89 characters (it's 90)

stitch/stitch_test.go
secant: Not sure what you mean by hit the network, but GetSpec is the only test that should download something from the Internet.

stitch/stitch_test.go
secant: Would it be correct to verify behavior by making sure the correct error is thrown? (Not sure how to verify otherwise..)

stitch/stitch_test.go
ejj: Ah so, what I'm saying is that no unit tests should hit the network.  I.e.
no unit tests should make any requests to remote websites.  Using the
network causes all sorts of problems in unit tests, most notably that they
don't work if the remote website goes down, or the computer is not
connected to the internet.  So for those tests in this patch that use the
network, we'll need to mock out the function calls that actually make the
network requests and replace them with dummy functions.

On Fri, Jul 8, 2016 at 11:57 AM, Vivian Fang <notifications@github.com>
wrote:

> In stitch/stitch_test.go
> <https://github.com/NetSys/quilt/pull/466#discussion_r70123868>:
>
> > @@ -1540,3 +1543,58 @@ func codeEq(a, b string) bool {
> >  	b = strings.TrimSpace(codeEqRE.ReplaceAllString(b, " "))
> >  	return a == b
> >  }
> > +
>
> Not sure what you mean by hit the network, but GetSpec is the only test
> that should download something from the Internet.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/NetSys/quilt/pull/466/files/930daf8a5bcfc539bf91e778752b05eea61a6353..28ad0145215d99eb4fa720e0c14a8bf3644b55d6#r70123868>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAl3lmG7BkpC9F2mXV3R0c2jiT96QeH8ks5qTp2CgaJpZM4I-PVX>
> .
>



-- 
Ethan J. Jackson
cs.berkeley.edu/~ejj/

ejj:
Merged

PR #467, opened 2016-06-27
Assertions

akshayknarayan:


inspect/invariant.go
```diff
@@ -57,36 +61,49 @@ func checkInvariants(graph graph, invs []invariant) ([]invariant, *invariant, er
...
+		}, nil
+	case len(sp) >= 2:
+		switch sp[1] {
+		case "true":
+			target = true
+		case "false":
+			target = false
+		default:
+			return invariant{}, fmt.Errorf(
+				"malformed assertion (second argument must be one of \"true\",\"false\"): %s",
```
ejj: This line is too long given 8 character tabs

inspect/invariant.go
```diff
@@ -57,36 +61,49 @@ func checkInvariants(graph graph, invs []invariant) ([]invariant, *invariant, er
...
 
-	// Validate label arguments.
-	var nodes []string
-	for _, n := range sp[2:] {
-		if _, ok := graph.nodes[n]; !ok {
-			return invariant{}, fmt.Errorf("malformed assertion (unknown label): %s", n)
+		// Validate label arguments.
+		for _, n := range sp[2:] {
+			if _, ok := graph.nodes[n]; !ok {
+				return invariant{}, fmt.Errorf("malformed assertion (unknown label): %s", n)
```
ejj: As is this one. (too long)

inspect/machine.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"github.com/NetSys/quilt/cluster/provider"
+	"github.com/NetSys/quilt/db"
+	"github.com/NetSys/quilt/stitch"
+)
+
+var providerToInstances map[db.Provider][]provider.Description
+
+func init() {
+	providerToInstances = map[db.Provider][]provider.Description{
```
ejj: Hmm, I'm not sure you need this...  I think you can just set this when declare initialize providerToInstances.

inspect/machine.go
```diff
@@ -0,0 +1,55 @@
...
+		db.Google: provider.GoogleDescriptions,
+	}
+}
+
+// A computer on which one or more containers may be scheduled.
+type machine struct {
+	zone    string  // cloud provider and region identifier
+	compute int     // number of available cores
+	memory  float64 // amount of available memory
+	//disk    int     // amount of available disk space. Ignore for now TODO
```
ejj: Just delete this, if you'd like to remember it for later you can file at trello card.

inspect/machine.go
```diff
@@ -0,0 +1,55 @@
...
+
+func getMachine(m stitch.Machine) machine {
+	zone := fmt.Sprintf("%s-%s", m.Provider, m.Region)
+	var comp int
+	var mem float64
+
+	if m.Size != "" {
+		descs, ok := providerToInstances[db.Provider(m.Provider)]
+		if !ok {
+			panic(fmt.Errorf("unknown provider: %s", m.Provider))
```
ejj: just fmt.Sprintf

ejj: I also think, this might be mildly cleaner if you did:
```
if m.Size == "" {
   return machine {
        zone: zone,
        comp: int(m.CPU.Min)
        mem: float64(m.RAM.Min)
    }
}

....
```

inspect/main.go
```diff
@@ -21,7 +21,10 @@ func usage() {
 		"Options\n"+
 		" - viz <pdf|ascii>\n"+
 		" - check <path to invariants file>\n"+
-		" - query [must have check] <path to query file>")
+		" - query [must have check] <path to query file>\n"+
```
ejj: This change doesn't really have anything to do with this patch,  Probably should be a separate one before or after.

ejj:
Cool, finished looking over this one.  Just some nits, I think it's ready.

inspect/main.go
akshayknarayan: Added a commit before. This should have been in the patch that added ascii visualization; I must have made a mistake on the rebase.

ejj:
Merged

PR #468, opened 2016-06-27
Makefile coverage stability

mmussomele:


ejj:
Merged with a couple of tweaks.  You introduced some trailing whitespace in the makefiles, make sure to check for it.

Thanks for pounding this out!

PR #469, opened 2016-06-29
stitch: Integrate invariant checking into stitch

akshayknarayan:
Add a stitch builtin, "invariant", which inlines invariant checking. To
make this work, now that labels can be queried independently,
modify graph construction to add nodes for all labels,
rather than inferring the presence of a node from
the list of connections in the stitch.


akshayknarayan:
Note: I didn't migrate the "query" ability since it's unclear how that should be incorporated.

stitch/inspect.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, m := range spec.QueryMachines() {
+		g.machines = append(g.machines, getMachine(m))
+	}
+
+	return g, nil
+}
+
+// InspectMain is the main function for inspect tool. Helps visualize stitches.
+func InspectMain(opts []string) {
+	configPath := opts[1]
```
sirspinach: I think this line isn't failing gracefully:

```
aetherboard:quilt shwangster$ quilt inspect
panic: runtime error: index out of range

goroutine 1 [running]:
panic(0xa13d20, 0xc820010090)
        /usr/local/go/src/runtime/panic.go:481 +0x3e6
github.com/NetSys/quilt/stitch.InspectMain(0xc82000a210, 0x1, 0x1)
        /Users/shwangster/go/src/github.com/NetSys/quilt/stitch/inspect.go:58 +0x91b
main.main()
        /Users/shwangster/go/src/github.com/NetSys/quilt/quilt.go:63 +0x59e
```

sirspinach: This probably should call `usage()` if `len(opts) == 2`.

stitch/inspect.go
```diff
@@ -0,0 +1,99 @@
...
+
+	f, err := os.Open(configPath)
+	if err != nil {
+		panic(err)
```
sirspinach: I'm not sure if a panic message and stacktrace is appropriate for mistyping a file name. Maybe just throw an error instead?

```
aetherboard:quilt shwangster$ quilt inspect a
panic: open a: no such file or directory

goroutine 1 [running]:
panic(0x9c8120, 0xc820210150)
        /usr/local/go/src/runtime/panic.go:481 +0x3e6
github.com/NetSys/quilt/stitch.InspectMain(0xc8200621c0, 0x2, 0x2)
        /Users/shwangster/go/src/github.com/NetSys/quilt/stitch/inspect.go:62 +0xaf
main.main()
        /Users/shwangster/go/src/github.com/NetSys/quilt/quilt.go:63 +0x59e
```

stitch/inspect.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+const quiltPath = "QUILT_PATH"
+
+type argOption struct {
+	result interface{} // Optionally store result.
+}
+
+func usage() {
+	fmt.Fprintln(os.Stderr, "Usage: inspect <path to spec file> [commands]\n"+
```
sirspinach: Perhaps it would be a good idea to explain here what `inspect` does. (I'm still not sure myself haha)

sirspinach:
ok, I'm done reviewing.

stitch/inspect.go
akshayknarayan: good catch, I messed up a rebase.

stitch/inspect.go
akshayknarayan: good idea, will do

stitch/inspect.go
akshayknarayan: good catch, I messed up a rebase.

stitch/inspect.go
sirspinach: By throw an error I meant to say 'print an error message'.

ejj: Btw, I generally agree.  Not a big fan of `panic`ing except when it's used more like an assert fail.  I.e. checking for bugs in the code.

ejj:
Normally, I'm not that strict on code coverage, but in the dsl, I've been aiming for 100%.  It's fairly easy to test and bugs can be quite subtle.  Could you up the coverage for the changes you made in stitch?

akshayknarayan:
incorporated @sirspinach 's comments into the first commit. Thanks!

The commit to remove a machine is because stitch already defines a Machine struct with the same information. It's a separate commit because I want to make it easy to revert later if we decide static analysis needs its own way of tracking machines.

The commit to move the visualization tools into a separate package rectifies the weird situation where visualization had to be in the stitch package to access non-exported fields. The relevant fields were exported and visualization was moved (back to the old "inspect" package).

akshayknarayan:
Changes on master have obsoleted this patch, and rebasing has led to significant changes. Closing this PR, will open a new PR for the rebased branch.

PR #470, opened 2016-06-30
Cluster rewrite

mmussomele:
Also added another unit test to foreman to cover the init function


cluster/foreman_test.go
```diff
@@ -122,6 +122,40 @@ func TestBoot(t *testing.T) {
 	}
 }
```
ejj: This commit should be folded into a different one.  Likely the rewrite patch.

I generally don't like "add test" commits when there's another commit that adds functionality they could be included with.

cluster/cluster.go
```diff
@@ -63,33 +49,51 @@ func newCluster(conn db.Conn, id int, namespace string) cluster {
...
+		inst := provider.New(p)
+		if err := inst.Connect(namespace); err == nil {
+			clst.providers[p] = inst
+		} else {
+			log.Debugf("Failed to connect to provider %s: %s", p, err)
+		}
+	}
+}
+
+func (clst cluster) startForeman() {
```
ejj: I don't see a lot of value in having this as a separate function.

mmussomele: I took it out of this commit (see commit below), but it gets taken out again because I need to use this code in 2 places for recreate providers on namespace change commit.

cluster/foreman_test.go
mmussomele: Done

cluster/foreman.go
```diff
@@ -178,18 +146,15 @@ func (fm *foreman) runOnce() {
...
-
-		err := m.client.setMinion(newConfig)
-		if err != nil {
-			return
-		}
-		err = m.client.bootEtcd(pb.EtcdMembers{IPs: etcdIPs})
-		if err != nil {
-			log.WithError(err).Warn("Failed send etcd members.")
-			return
+		if newConfig != m.config {
```
ejj: The way it was was fine.  It's go style to keep the indentation as small as possible.

cluster/cluster.go
```diff
@@ -25,36 +25,20 @@ type cluster struct {
...
-		for _, row := range dbClusters {
-			clst, ok := clusters[row.ID]
-			if !ok {
-				new := newCluster(conn, row.ID, row.Namespace)
-				clst = &new
-				clusters[row.ID] = clst
-			}
-
-			clst.mark = true
+	<-conn.TriggerTick(60, db.ClusterTable).C
```
ejj: I don't much like how this function is organized. It seems awfully fragile and the way it deals with namespaces changes (in futures patches) doesn't seem optimal.  I think it'd be better to loop over the trigger (like the original code).  On each loop lookup the namespace and re-create the cluster if it changed.

cluster/cluster.go
```diff
@@ -63,33 +47,47 @@ func newCluster(conn db.Conn, id int, namespace string) cluster {
 		id:        id,
```
ejj: Do we still need the id?

cluster/cluster.go
```diff
@@ -63,33 +47,47 @@ func newCluster(conn db.Conn, id int, namespace string) cluster {
...
 		conn:      conn,
```
ejj: Come to think of it, do we still need the cluster struct at all?  Can these be member variables or something instead?

cluster/cluster.go
```diff
@@ -63,33 +47,47 @@ func newCluster(conn db.Conn, id int, namespace string) cluster {
...
 		}
 	}
+
 	go func() {
 		rateLimit := time.NewTicker(5 * time.Second)
 		defer rateLimit.Stop()
+
+		// Wait until the cluster can sync with the db, then sync and initialize
+		// the foreman.
+		<-clst.trigger.C
```
ejj: Do you really need this?  It seems kinda odd...

I think you're waiting for the database to be initialized?  Won't it already be initialized at this point?  I.E. the cluster table is already populated, I'd expect the other tables to be as well right?

cluster/cluster.go
```diff
@@ -63,33 +47,47 @@ func newCluster(conn db.Conn, id int, namespace string) cluster {
...
 		}
 		for _, p := range clst.providers {
 			p.Disconnect()
 		}
 	}()
+
 	return clst
 }
 
+func (clst cluster) startForeman() {
```
ejj: This function seems kind of pointless, I'd ditch it and just move it's code into the above caller.

cluster/cluster.go
```diff
@@ -63,33 +47,47 @@ func newCluster(conn db.Conn, id int, namespace string) cluster {
...
-		err := inst.Connect(namespace)
-		if err == nil {
+		if err := inst.Connect(namespace); err == nil {
 			clst.providers[p] = inst
 		} else {
 			log.Debugf("Failed to connect to provider %s: %s", p, err)
 		}
 	}
+
 	go func() {
```
ejj: Do we really need a go routine still?

cluster/cluster.go
```diff
@@ -63,33 +47,47 @@ func newCluster(conn db.Conn, id int, namespace string) cluster {
...
 			<-rateLimit.C
+			clst.sync()
+			clst.fm.runOnce()
 		}
 		for _, p := range clst.providers {
 			p.Disconnect()
 		}
 	}()
+
 	return clst
```
ejj: No one uses this return result, should we still be returning it?  Basically, at a high level, I think newCluster could be further simplified.

cluster/cluster.go
```diff
@@ -102,6 +100,11 @@ func (clst cluster) get() ([]provider.Machine, error) {
 	return cloudMachines, nil
 }
 
+func (clst cluster) shutdown() {
```
ejj: Nothing seems to call this, why add it?

cluster/cluster.go
```diff
@@ -17,6 +17,7 @@ type cluster struct {
 	trigger db.Trigger
```
ejj: I think I'd combine the cluster namespace patch with the rewrite patch.  There's not a strong reason to keep them separate

cluster/cluster.go
```diff
@@ -149,19 +152,14 @@ func (clst cluster) sync() {
...
 	var machines []db.Machine
 	clst.conn.Transact(func(view db.Database) error {
-		clusters := view.SelectFromCluster(func(c db.Cluster) bool {
-			return c.ID == clst.id
-		})
-
-		machines = view.SelectFromMachine(func(m db.Machine) bool {
-			return m.ClusterID == clst.id
-		})
+		clusters := view.SelectFromCluster(nil)
```
ejj: Do you still need `clusters` at all?  Looks like you're just selecting it in case you need to panic.

cluster/cluster.go
```diff
@@ -190,25 +188,24 @@ func (clst cluster) sync() {
...
+				m := pair.L.(provider.Machine)
+				dbm := pair.R.(db.Machine)
 
 				dbm.CloudID = m.ID
 				dbm.PublicIP = m.PublicIP
 				dbm.PrivateIP = m.PrivateIP
 
-				// If we overwrite the machine's size before the machine has fully
-				// booted, the Stitch will flip it back immediately.
+				// If we overwrite the machine's size before the machine
```
ejj: Why this change?

cluster/cluster.go
```diff
@@ -190,25 +188,24 @@ func (clst cluster) sync() {
...
 			return nil
 		})
 
 		pairs, bootSet, terminateSet := syncDB(cloudMachines, dbMachines)
 
 		clst.conn.Transact(func(view db.Database) error {
 			for _, pair := range pairs {
-				dbm := pair.L.(db.Machine)
-				m := pair.R.(provider.Machine)
+				m := pair.L.(provider.Machine)
```
ejj: Why this change?

cluster/cluster.go
```diff
@@ -252,8 +249,8 @@ func (clst cluster) syncACLs(acls []string, machines []db.Machine) {
 func syncDB(cloudMachines []provider.Machine, dbMachines []db.Machine) (
 	pairs []join.Pair, bootSet []provider.Machine, terminateSet []provider.Machine) {
 	scoreFun := func(left, right interface{}) int {
-		dbm := left.(db.Machine)
-		m := right.(provider.Machine)
+		m := left.(provider.Machine)
```
ejj: Why this change?

cluster/foreman.go
```diff
@@ -50,55 +49,32 @@ type minion struct {
...
 		machines := view.SelectFromMachine(func(m db.Machine) bool {
-			return m.ClusterID == fm.clusterID && m.PublicIP != "" &&
-				m.PrivateIP != "" && m.CloudID != ""
+			return m.PublicIP != "" && m.PrivateIP != "" && m.CloudID != ""
 		})
 
 		fm.updateMinionMap(machines)
 
 		fm.forEachMinion(func(m *minion) {
-			var err error
```
ejj: I think the previous code was easier to read.

cluster/foreman.go
```diff
@@ -116,17 +92,17 @@ func (fm *foreman) runOnce() {
...
-			return c.ID == fm.clusterID
-		})
+		clusters := view.SelectFromCluster(nil)
 
 		fm.spec = ""
-		if len(clusters) == 1 {
+		numClusters := len(clusters)
+		if numClusters == 1 {
 			fm.spec = clusters[0].Spec
+		} else if numClusters > 1 {
```
ejj: I might pull this logic into a helper in db.cluster btw.

ejj: I.E. a view.GetSingleCluster() or something.

ejj:
I think you could get better coverage of provider.go, if you reorganized newCluster a bit too btw.  Would be nice to have some of that logic covered as it's a bit tricky

cluster/cluster.go
mmussomele: I left it in because I thought it might be useful for testing, in order to test what happens when we reboot the cluster. Can definitely be removed though

ejj: Ah, so as a general principle.  I don't like to commit code because it
might be useful later.  If it turns out you need it for tests in a future
commit, then you can add it at that time.

Ethan

On Thu, Jul 7, 2016 at 11:36 PM, mmussomele <notifications@github.com>
wrote:

> In cluster/cluster.go
> <https://github.com/NetSys/quilt/pull/470#discussion_r70030940>:
>
> > @@ -102,6 +100,11 @@ func (clst cluster) get() ([]provider.Machine, error) {
> >  	return cloudMachines, nil
> >  }
> >
> > +func (clst cluster) shutdown() {
>
> I left it in because I thought it might be useful for testing, in order to
> test what happens when we reboot the cluster. Can definitely be removed
> though
>
> —
> You are receiving this because you commented.
> <https://github.com/NetSys/quilt/pull/470/files/a2246cc4e6284b03afcbfc17a621930d3d58ba9c..3acc14ec148ac98184fc2180db5ac3dd6d2dad6d#r70030940>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAl3lrcCQwcFi95NAvou7WJUGRExhDChks5qTe_8gaJpZM4JCsqE>
> .
>



-- 
Ethan J. Jackson
cs.berkeley.edu/~ejj/

cluster/cluster.go
mmussomele: It's still needed to get the ACLs from the single cluster, but adding the db helper simplifies this greatly

cluster/cluster.go
```diff
@@ -11,85 +11,78 @@ import (
...
-	id      int
 	conn    db.Conn
 	trigger db.Trigger
 	fm      foreman
 
+	namespace string
 	providers map[db.Provider]provider.Provider
 
-	mark bool /* For mark and sweep garbage collection. */
+	init bool /* For mark and sweep garbage collection. */
```
ejj: This comment is out of date.

mmussomele:
Note: The proper commit order is
cluster: Rewrite cluster.go and foreman.go
cluster: Remove provider interface's Disconnect() function
provider: Add more thorough tests

cluster/provider/provider_test.go
```diff
@@ -119,3 +119,54 @@ func TestConstraints(t *testing.T) {
 	checkConstraint(testDescriptions, stitch.Range{Min: 3},
```
ejj: Is there a good reason for this to be a separate patch?

cluster/cluster.go
```diff
@@ -11,85 +11,78 @@ import (
...
-	id      int
 	conn    db.Conn
 	trigger db.Trigger
 	fm      foreman
 
+	namespace string
 	providers map[db.Provider]provider.Provider
 
-	mark bool /* For mark and sweep garbage collection. */
+	init bool /* For checking that a cluster has been created before stopping it */
```
ejj: Does this really need to be a member of the struct?  I think it could just be a local variable of Run or, alternatively, you could use a `*cluster` which can be nil on initialization.

ejj:
Series looks good, just the nit about `init`.

Also, either let me know why the test patch should be separate, or fold it in to the rewrite patch.  Ping me when updated and I'll merge.

cluster/cluster.go
mmussomele: Fixed

cluster/cluster.go
```diff
@@ -190,9 +167,7 @@ func (clst cluster) sync() {
 
 		var dbMachines []db.Machine
 		clst.conn.Transact(func(view db.Database) error {
-			dbMachines = view.SelectFromMachine(func(m db.Machine) bool {
-				return m.ClusterID == clst.id
-			})
+			dbMachines = view.SelectFromMachine(nil)
```
secant: This function never errors

cluster/foreman.go
```diff
@@ -145,17 +113,8 @@ func (fm *foreman) runOnce() {
 		}
 
 		m.connected = connected
-	})
```
secant: Remove this empty line

cluster/foreman_test.go
```diff
@@ -122,10 +120,119 @@ func TestBoot(t *testing.T) {
...
+	})
+
+	fm.init()
+	for _, m := range fm.minions {
+		if m.machine.Role != db.Worker {
+			t.Error("Minion machine not set to worker.")
+		}
+	}
+
+	fm = startTestWithRole(pb.MinionConfig_Role(-7))
```
secant: (picky) Could you move `startTestWithRole` to be below this function, instead of a couple functions down?

db/cluster.go
```diff
@@ -19,44 +19,47 @@ type Cluster struct {
...
 	}
 
-	return result
+	numClusters := len(clusters)
+	if numClusters == 1 {
+		return clusters[0], nil
+	} else if numClusters > 1 {
+		log.Panicf("Found %d clusters, there should be 1", numClusters)
+	}
+	return Cluster{}, errors.New("no clusters found")
```
secant: Just wondering, why don't you panic for this? Or why do you panic instead of throwing an error when numClusters > 1?

db/cluster.go
```diff
@@ -19,44 +19,47 @@ type Cluster struct {
...
-	return c.ID == other.ID &&
-		c.Namespace == other.Namespace &&
+	return c.Namespace == other.Namespace &&
 		c.Spec == other.Spec &&
 		util.StrSliceEqual(c.ACLs, other.ACLs)
 }
 
 func (c Cluster) getID() int {
-	return c.ID
+	return 0
```
secant: Does this function need to exist at all?

db/db_test.go
```diff
@@ -76,15 +76,15 @@ func TestMachineSelect(t *testing.T) {
```
secant: Check if `err != nil`?

cluster/foreman_test.go
mmussomele: I suppose I could. I personally thought that it nicely separated the file into "Test functions" and "Util functions", but I'm indifferent to which way it is. Does @ejj have a preference?

cluster/provider/provider_test.go
```diff
@@ -120,3 +120,54 @@ func TestConstraints(t *testing.T) {
...
 		stitch.Range{}, 0, "size4")
 }
+
+func TestNewProviderSuccess(t *testing.T) {
+	defer func() {
+		if r := recover(); r != nil {
+			t.Error("provider.New paniced on valid provider")
```
secant: panicked

db/cluster.go
mmussomele: It doesn't panic when there are 0 clusters because the db may simply have not updated or synced up yet. But if there's more than 1 cluster, something is very wrong since there should only ever be 1.

db/cluster.go
mmussomele: Cluster needs to implement the `row` interface, which requires a getID() method. Since there will only be one cluster, the implementation is just trivial.

db/db_test.go
mmussomele: Actually there's no reason to have err declared here, since the Transact always returns nil. I just get rid of it.

mmussomele:
Made all the fixes and resolved the conflicts.

ejj:
Merged

PR #473, opened 2016-07-03
Azure support

ZhongtianWang:


ZhongtianWang:
I reopened a new PR on this. It's cleaner to see :D

Azure.md
```diff
@@ -0,0 +1,45 @@
...
+Directory application that can access resources](https://azure.microsoft.com/en-us/documentation/articles/resource-group-create-service-principal-portal/). 
+Make sure you set the scope of the role to your subscription.
+
+Following through the tutorial, you have now created an AD application as 
+well as a service principal for that application. You have assigned the 
+service principal to a role. Now, you need to give Quilt access to perform 
+operations. To do that, create a file `credentials.json` in directory 
+`~/.azure` with the following format:
+
+~~~~
```
secant: Just for consistency across the docs, use
~~~
    ```json
    code here
    ```
~~~

cluster/provider/azure.go
```diff
@@ -122,34 +226,126 @@ func (clst *azureCluster) List() ([]Machine, error) {
...
+
+	return forEachMachine(bootFunc, bootSet)
 }
 
-// Delete Azure instances (blocking by calling instanceDel).
 func (clst *azureCluster) Stop(machines []Machine) error {
-	for _, m := range machines {
-		if err := clst.instanceDel(m.ID); err != nil {
+	stopFunc := func(m Machine) error {
+		cancel := make(chan struct{})
```
secant: Are you using the channel you made in this function?

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		return errors.New(dir + " contains invalid clientSecret")
+	}
+
+	if clst.tenantID, ok = cred["tenantID"]; !ok {
+		return errors.New(dir + " contains invalid tenantID")
+	}
+
+	if clst.subscriptionID, ok = cred["subscriptionID"]; !ok {
+		return errors.New(dir + " contains invalid subscriptionID")
+	}
```
secant: Not sure if this is enough redundancy to warrant a helper function?

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		Properties: &properties,
+		Tags:       &map[string]*string{nsTag: &clst.namespace},
+	}
+
+	cancel := make(chan struct{})
+	_, err = clst.azureClient.storageCreate(resourceGroupName, storageName, param,
+		cancel)
+	if err != nil {
+		return storageAccount, err
+	}
```
secant: You can make lines 465-467 a one liner if-statement

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+	var storageAccount storage.Account
+
+	// Storage name needs to be globally unique, with a limit of 24 characters.
+	storageName := randomString(24)
+	cna, err := clst.azureClient.storageCheckName(
+		storage.AccountCheckNameAvailabilityParameters{
+			Name: &storageName,
+			Type: stringPtr("Microsoft.Storage/storageAccounts")})
+	if err != nil {
+		return storageAccount, err
```
secant: Also in all of these `err != nil` bodies, do you want to return `storageAccount` or `nil`?

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		AddressPrefix: stringPtr(vnetSubnetAddressPrefix),
+	}
+
+	subnet := network.Subnet{
+		ID:         &id,
+		Name:       stringPtr(subnetName),
+		Properties: &properties,
+	}
+
+	return subnet, nil
```
secant: Does this function ever error?

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		Name:       &vnetName,
+		Location:   &location,
+		Properties: &properties,
+		Tags:       &map[string]*string{nsTag: &clst.namespace},
+	}
+
+	cancel := make(chan struct{})
+	_, err := clst.azureClient.vnetCreate(resourceGroupName, vnetName,
+		virtualNetwork, cancel)
+	return virtualNetwork, err
```
secant: Same thing, are you supposed to return nil or virtualNetwork when it errors?

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+func (clst *azureCluster) listSecurityGroups() (map[string]network.SecurityGroup,
+	error) {
+	result, err := clst.azureClient.securityGroupList(resourceGroupName)
+	if err != nil {
+		return nil, err
+	}
+
+	secGroups := map[string]network.SecurityGroup{}
+	for _, secGroup := range *result.Value {
+		if tag := (*secGroup.Tags)[nsTag]; tag == nil || *tag != clst.namespace {
```
secant: This line also gets repeated a lot. Maybe you could make a function that returns whether a tag is valid or not. Thoughts?

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		Properties: &properties,
+		ID:         &id,
+		Tags: &map[string]*string{
+			nsTag: &clst.namespace,
+			vmTag: &vmName,
+		},
+	}
+
+	cancel := make(chan struct{})
+	_, err := clst.azureClient.publicIPCreate(resourceGroupName, nicName, publicIP,
```
secant: `publicIP` already contains `nicName`

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+	virtualNetwork := network.VirtualNetwork{
+		Name:       &vnetName,
+		Location:   &location,
+		Properties: &properties,
+		Tags:       &map[string]*string{nsTag: &clst.namespace},
+	}
+
+	cancel := make(chan struct{})
+	_, err := clst.azureClient.vnetCreate(resourceGroupName, vnetName,
+		virtualNetwork, cancel)
```
secant: `virtualNetwork` contains `vnetName`

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		Location:   &location,
+		Properties: &properties,
+		Tags: &map[string]*string{
+			nsTag: &clst.namespace,
+			vmTag: &vmName,
+		},
+	}
+
+	cancel := make(chan struct{})
+	_, err := clst.azureClient.ifaceCreate(resourceGroupName, nicName, iface, cancel)
```
secant: `iface` has `nicName`

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+	virtualMachine := compute.VirtualMachine{
+		Name:       &vmName,
+		Location:   &location,
+		Properties: &properties,
+		Tags:       &map[string]*string{nsTag: &clst.namespace},
+	}
+
+	cancel := make(chan struct{})
+	_, err = clst.azureClient.vmCreate(resourceGroupName, vmName, virtualMachine,
+		cancel)
```
secant: `vmName` is in `virtualMachine`

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		outboundName := strings.Replace(address, "/", "-", -1) + "-out"
+		outboundProperties := network.SecurityRulePropertiesFormat{
+			Protocol:                 network.Asterisk,
+			SourcePortRange:          &asterisk,
+			SourceAddressPrefix:      &asterisk,
+			DestinationPortRange:     &asterisk,
+			DestinationAddressPrefix: &address,
+			Access:    network.Allow,
+			Direction: network.Outbound,
+		}
```
secant: Again, not sure how picky I'm supposed to be on repetition, but maybe something can be done about repeating the first five fields? Maybe that would be uglier... 😅

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		cloudInRules); err != nil {
+		return err
+	}
+
+	if err := clst.syncSecurityRules(securityGroupName, localOutRules,
+		cloudOutRules); err != nil {
 		return err
 	}
 
 	return nil
```
secant: I think you can just return `clst.syncSecurityRules(...)`

cluster/provider/azure.go
```diff
@@ -159,92 +355,706 @@ func (clst *azureCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		rule := r.(network.SecurityRule)
+		delete(currPriorities, *rule.Properties.Priority)
+		if _, err := clst.azureClient.securityRuleDelete(resourceGroupName,
+			securityGroupName, *rule.Name, cancel); err != nil {
+			return err
+		}
+	}
+
+	priority := int32(100)
+	for i := 0; i < len(addList); i++ {
```
secant: Is it better to do `range addList`?

cluster/provider/azure_test.go
```diff
@@ -0,0 +1,533 @@
+package provider
```
secant: Should there be more coverage? [coveralls](https://coveralls.io/builds/6849323/source?filename=cluster%2Fprovider%2Fazure.go)

cluster/provider/azure_test.go
```diff
@@ -0,0 +1,533 @@
...
+				DestinationAddressPrefix: &address,
+				Access:    network.Allow,
+				Direction: network.Outbound,
+			}
+			outboundRule := network.SecurityRule{
+				Name:       &outboundName,
+				Properties: &outboundProperties,
+			}
+			expectedRules = append(expectedRules, outboundRule)
+		}
```
secant: Is it considered testing if the expected output is generated by doing exactly what the function you're testing does?

secant:
I'm done ^^

cluster/provider/azure.go
ZhongtianWang: Yes, for `vmDelete`, `ifaceDelete` and `publicIPDelete`, all of them need a `cancel` channel.

cluster/provider/azure.go
ZhongtianWang: We can't return nil as struct, that's why I'm just using an empty `storageAccount ` struct

cluster/provider/azure.go
ZhongtianWang: Good catch, this one doesn't really return any errors 😸

cluster/provider/azure.go
ZhongtianWang: I think it's a pretty good idea :D. thanks

cluster/provider/azure.go
ZhongtianWang: This is just an attempt to stay in sync with Azure's api. I does seem weird, but I think it should be fine for now

cluster/provider/azure.go
ZhongtianWang: I think it's fine for now. We can leave it like this so in the future it would be easier if we ever need to add more complex security rules.

cluster/provider/azure.go
ZhongtianWang: yeaaaaaa 💃

ejj:
btw vivian, I tweaked spec_test, you should rebase and update.

cluster/provider/azure_test.go
ZhongtianWang: In SyncACL, it actually calls `securityRuleList` to get all the cloud rules and check whether each sec group contains all the rules

cluster/provider/azure.go
ZhongtianWang: Not sure, but I think it should be fine :P

cluster/provider/azure_test.go
ejj: What if `securityRuleList` has a bug, will you catch it?

ZhongtianWang: `securityRuleList` is just calling `securityRuleClient.List`. And in mocked version, it just lists all the rules in the cloud. If `securityRuleList` has a bug, then the test would fail. So this test calls the `SetACL` function, and all rules are synced to the mock cluster, then we list all the rules from the mock cluster to see if they all match localACLs. Hmmm I think I'm not sure what do you mean by correct output and consistent output.

ejj: It looks like, to generated `expectedRules` you just copy and pasted the
code that you're testing.  So, in effect, you're verifying that the
behavior of the code you're testing is the same as the behavior of the code
you've copied which (of course) it is.  The question is, how do you know
the code you copied doesn't have a bug?  You don't really . . .

On Tue, Jul 5, 2016 at 5:49 PM, Zhongtian Wang <notifications@github.com>
wrote:

> In cluster/provider/azure_test.go
> <https://github.com/NetSys/quilt/pull/473#discussion_r69644489>:
>
> > +			outboundName := strings.Replace(address, "/", "-", -1) + "-out"
> > +			outboundProperties := network.SecurityRulePropertiesFormat{
> > +				Protocol:                 network.Asterisk,
> > +				SourcePortRange:          &asterisk,
> > +				SourceAddressPrefix:      &asterisk,
> > +				DestinationPortRange:     &asterisk,
> > +				DestinationAddressPrefix: &address,
> > +				Access:    network.Allow,
> > +				Direction: network.Outbound,
> > +			}
> > +			outboundRule := network.SecurityRule{
> > +				Name:       &outboundName,
> > +				Properties: &outboundProperties,
> > +			}
> > +			expectedRules = append(expectedRules, outboundRule)
> > +		}
>
> securityRuleList is just calling securityRuleClient.List. And in mocked
> version, it just lists all the rules in the cloud. If securityRuleList
> has a bug, then the test would fail. So this test calls the SetACL
> function, and all rules are synced to the mock cluster, then we list all
> the rules from the mock cluster to see if they all match localACLs. Hmmm I
> think I'm not sure what do you mean by correct output and consistent output.
>
> —
> You are receiving this because you commented.
> <https://github.com/NetSys/quilt/pull/473/files/d799fdce781821bb1cc9ff983c43ebdea84511d7..c1418cdc5ef747434ed4069d56a2d4b18f789614#r69644489>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAl3lobh-pZhNmwdrCId73gde-MinWKpks5qStF7gaJpZM4JDzqk>
> .
>



-- 
Ethan J. Jackson
cs.berkeley.edu/~ejj/

ZhongtianWang: In my original thought, since the code I copied in is exactly what we need to generate the correct rules, and the test just check the correctness of behaviors of `SetACLs`, `syncSecurityGroup` and `syncSecurityRules`. I guess that is what consistency means :P. So should I instead derive the acls from the list result of all the security rules and check whether they are the same to what was passed in?

ejj: Yeah I would just manually write out what the right result would be, perhaps using a helper to express them.  At least that way we can have some confidence that the code is correct.

cluster/provider/azure.go
```diff
@@ -122,34 +226,122 @@ func (clst *azureCluster) List() ([]Machine, error) {
...
-	return nil
+
+	return forEachMachine(stopFunc, machines)
 }
 
-// Disconnect.
 func (clst *azureCluster) Disconnect() {
-	// nothing
+	cancel := make(chan struct{})
+	clst.azureClient.rgDelete(resourceGroupName, cancel)
```
secant: per Matt's [PR](https://github.com/NetSys/quilt/pull/470/commits/a9c7455b47bb9545695e0af12c8020d5876fb0ee), this  should be removed

ZhongtianWang: Ah didn't see that patch. Alright!

ZhongtianWang:
Removed logic for `Disconnect`. :D

ejj:
Merged.  Still need to implement ChoozeSize, and get better test coverage . . . but that can happen in later patches.

PR #475, opened 2016-07-06
Integrate invariants into stitch

akshayknarayan:
Github is again showing the commits out of order.
Correct commit order: "Integrate inv..." -> "Remove redundant..." -> "Move vis..."


Makefile
```diff
@@ -1,18 +1,19 @@
 export GO15VENDOREXPERIMENT=1
 PACKAGES=$(shell GO15VENDOREXPERIMENT=1 go list ./... | grep -v vendor)
 NOVENDOR=$(shell find . -path ./vendor -prune -o -name '*.go' -print)
-LINE_LENGTH_EXCLUDE=./cluster/provider/constraintConstants.go \
-		    ./cluster/provider/gceConstants.go \
-		    ./minion/pb/pb.pb.go \
-		    ./cluster/provider/cloud_config.go \
-		    ./minion/network/link_test.go
+LINE_LENGTH_EXCLUDE=./constants/awsConstants.go \
```
ejj: Not sure this is the right patch for this change, or if it's even necessary.

Makefile
```diff
@@ -1,18 +1,19 @@
...
+					./minion/network/link_test.go \
+					./minion/pb/pb.pb.go \
 
 REPO = quilt
 DOCKER = docker
 SHELL := /bin/bash
 
 .PHONY: all
-all: quilt minion inspect
+all: quilt minion
```
ejj: Why?

cluster/provider/awsSpot.go
```diff
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/NetSys/quilt/constants"
```
ejj: I think this constants change probably belongs in it's own patch.

constants/azureConstants.go
```diff
@@ -0,0 +1,4 @@
+package constants
+
+// AzureDescriptions enumerates Microsoft Azure instance offerings TODO
```
ejj: Do `XXX:  AzureDescriptions ...` and ditch the "TODO"

Makefile
ejj: Oh, you're removing inspect entirely?

Makefile
```diff
@@ -1,18 +1,19 @@
 export GO15VENDOREXPERIMENT=1
```
ejj: So, one thing that would make this patch much easier to review, is if you broke it into basically three patches.  1st one would Move the inspect code into the stitch directory, basically unchanged, and make inspect a sub-command of the main quilt command.  Second patch would do the constant stuff.  Third patch would integrate the invariant checking into stitch.  That way, it's much easier to tell what substantively changed, versus just moved.

stitch/invariant_test.go
```diff
@@ -0,0 +1,137 @@
+package stitch
+
+import (
+	"strings"
+	"testing"
+	"text/scanner"
+)
+
+func initSpec(src string) (Stitch, error) {
```
ejj: One problem with these tests, is they don't catch the case where an invariant fails.  As a result, all of these tests would pass if the invariant code did absolutely nothing at all.  I think with only a bit more code they could catch the negative case as well.

constants/azureConstants.go
akshayknarayan: this fails the linter: `comment on exported var AzureDescriptions should be of the form "AzureDescriptions ..."`

ejj: Ah, darn.  Well replace the TODO with XXX then

ejj:
Thanks for reorganizing the commit history, much easier to review now.

Makefile
ejj: The removal of the inspect target belongs in the next patch I think.

ejj:
First three patches look good, one minor rebasing error. Ill look at the last three today.

constants/azureConstants.go
```diff
@@ -1,4 +1,4 @@
 package constants
 
-// XXX: AzureDescriptions enumerates Microsoft Azure instance offerings
```
ejj: Oh, I think this change should be in a previous patch too, you probably break make lint earlier and then fix it now.

akshayknarayan:
Correct commit order:
3793949 (HEAD -> stitch-inv, origin/stitch-inv) inspect: Move visualization tool to separate package
e34a076 stitch: Integrate invariant checking into stitch
a534f34 inspect: Move inspect to within stitch package
a693365 cluster: Move instance descriptions to separate package
f4c0364 stitch: Remove redundant struct defining a machine
e343290 (upstream/master, origin/master, master) cluster: Word wrap to 89 characters

ejj:
Merged the first two patches (you may want to rebase).  Third patch fails unit tests (likely rebasing error).

stitch/eval.go
```diff
@@ -11,6 +11,7 @@ type evalCtx struct {
 	connections map[Connection]struct{}
 	placements  map[Placement]struct{}
 	machines    *[]*astMachine
+	invariants  *[]invariant
```
ejj: Did you consider using a `map[invariant]struct{}`?  Not sure which is better in this case.  Are duplicate invariants meaningful?  Does order matter?  If the answers are no and no, a set might be better.

stitch/func.go
```diff
@@ -744,6 +745,98 @@ func hmapValuesImpl(ctx *evalCtx, args []ast) (ast, error) {
...
+func invariantImpl(ctx *evalCtx, args []ast) (ast, error) {
+	var astArgs []string
+	for _, arg := range args {
+		astArgs = append(astArgs, arg.String())
+	}
+	invStr := strings.Join(astArgs, " ")
+
+	if len(args) == 0 {
+		return invariant{}, fmt.Errorf(
+			"malformed invariant"+
```
ejj: Fits without the "+" and the break of the format string (same for quite a few)

stitch/inspect.go
```diff
@@ -0,0 +1,97 @@
...
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+	"text/scanner"
+)
+
+const quiltPath = "QUILT_PATH"
+
+func usage() {
```
ejj: Should this be in a previous patch?

ejj: Actually, come to think of it, this whole file seems kind of out of place in this patch.  Should it be in it's own, or one of the earlier/later patches?

ejj: Hmm, actually I'm not sure either way, I guess if the invariants moved into stitch, the inspect implementation needs to as well ... what do you think?

stitch/inspect.go
```diff
@@ -0,0 +1,97 @@
...
+	fmt.Fprintln(
+		os.Stderr,
+		`inspect is a tool that helps visualize Stitch specifications.
```
ejj: Should this be `quilt inspect is a tool...`?

stitch/inspect.go
```diff
@@ -0,0 +1,97 @@
...
+Usage: inspect <path to spec file> <pdf|ascii>
```
ejj: Similar quesiton.

stitch/inspect.go
```diff
@@ -0,0 +1,97 @@
...
+	sc := scanner.Scanner{
+		Position: scanner.Position{
+			Filename: configPath,
+		},
+	}
+	pathStr, _ := os.LookupEnv(quiltPath)
+	pathSlice := strings.Split(pathStr, ":")
+	spec, err := New(*sc.Init(bufio.NewReader(f)), pathSlice)
+	if err != nil {
+		panic(err)
```
ejj: I'm not sure a panic is appropriate.  Better to print an error and exit in this case I think.

stitch/inspect.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	graph, err := initializeGraph(spec)
+	if err != nil {
+		panic(err)
```
ejj: Same here I think.

stitch/invariant.go
```diff
@@ -59,121 +57,86 @@ func checkInvariants(graph graph, invs []invariant) ([]invariant, *invariant, er
...
+		case inv.nodes[2]:
+			betweenNodes = append(betweenNodes, node)
+		}
 	}
 
-	betweenNode := inv.nodes[2]
+	allPassed := true
+	for _, from := range fromNodes {
+		for _, to := range toNodes {
+			paths, ok := paths(from, to)
```
ejj: I might pull this loop body into a function if you think it makes sense.  The indentation level is getting pretty deep.

stitch/placement.go
```diff
@@ -131,6 +138,16 @@ func (g graph) findAvailabilitySet(label string) availabilitySet {
 //   - Create a new set for nodes that could not be moved to an existing set.
 func (g *graph) placeNodes() {
 	for node, wantExclusives := range g.placement {
+		if _, ok := g.nodes[node]; !ok {
+			panic(
```
ejj: I don't dislike this, but it's pretty inconsistent with the rest  of the code . . .

Also is a panic appropriate?

ejj:
Finished reviewing the main invariant patch.  Basically looks good.

stitch/inspect.go
akshayknarayan: It was easier to move it into stitch (access to unexported fields) in this commit, stripping out the invariant stuff. In a later commit I exported the relevant vars and re-created the inspect/ package.

stitch/eval.go
akshayknarayan: I looked into this, but it seems `invariant` can't be used as a map key since it contains a slice.

stitch/placement.go
akshayknarayan: panic I think is appropriate, since a node inside a Placement struct should always be a valid node.

Are you saying you don't like the indentation style?

akshayknarayan:
Alright, made changes to address comments.

akshayknarayan:
@ejj rebased, ready for review

ejj:
Merged

PR #476, opened 2016-07-07
network: Fix duplicate ACL rules

ZhongtianWang:
Before this patch, we always try to add duplicate icmp rules, and
that spawns endless scary log messages as it retries infinitely to
add rules even though they are already existed in br-int tables.

Also, updates in ContainerTable and LabelTable creates two seperate
triggers. But we need both entities before we install ACLs. The
current fix to this is to only update ACLs when we have same number
of containers and labels.


ZhongtianWang:
Currently I'm checking `len(containers) == len(labels)` since `ContainerTable` and `LabelTable` create separate triggers. This will be changed when I push the label patch.

mmussomele:
Seems pretty good to me 👍

minion/network/network.go
```diff
@@ -135,7 +135,9 @@ func runMaster(conn db.Conn) {
 		}
 	}
 
-	updateACLs(connections, labels, containers)
+	if len(containers) == len(labels) {
```
ejj: Why this change?

minion/network/network.go
```diff
@@ -176,16 +178,22 @@ func updateACLs(connections []db.Connection, labels []db.Label,
 			max := conn.MaxPort
```
ejj: Weren't you going to skip if `fromIP` or `toIP` were `""`?

minion/network/network.go
ZhongtianWang: ahhh I made such a silly mistake. It's better to just check `fromIP` and `toIP` inside `CreateACL`.

minion/network/network.go
```diff
@@ -172,20 +172,30 @@ func updateACLs(connections []db.Connection, labels []db.Label,
 		for _, fromDbc := range labelDbcMap[conn.From] {
 			fromIP := fromDbc.IP
 			toIP := labelIPMap[conn.To]
+			if len(fromIP) == 0 || len(toIP) == 0 {
```
ejj: would be more typical go style to do `fromIP == ""`

ejj:
Merged

PR #477, opened 2016-07-09
azure: Add mutex to azure mock

ZhongtianWang:
Add mutex to avoid race condition in concurrent map writes.

### DIFF
--- a/cluster/provider/azure_test.go
+++ b/cluster/provider/azure_test.go
@@ -1,6 +1,7 @@
 package provider
 
 import (
+	"sync"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


ejj:
You're missing a bunch of functions, notably security groups list.

I would just add the following to the start of every function in fakeClient
```
mutex.Lock()
dever mutex.Unlock()
```

Right now you're trying to do fine-grained locking which is going to lead to mistakes, just shove it around everything.

ZhongtianWang:
That was very silly mistake I made. Thanks for the tip. That's a much better idea..

PR #480, opened 2016-07-11
db: Simplify db table alert system

mmussomele:
Instead of listening through a channel in a goroutine, Transact now
uses a mutex lock to ensure safe concurrent access. Also, instead of
tracking the table seq, we simply flag tables as needing to be
alerted or not, which allows us to ditch row.equal and just use
reflect.DeepEqual for better correctness with minimal performance
hits.


db/db.go
```diff
@@ -23,18 +24,15 @@ type Trigger struct {
...
-type transaction struct {
-	do   func(db Database) error
-	done chan error
-}
-
 // A Conn is a database handle on which transactions may be executed.
-type Conn chan transaction
+type Conn struct {
+	db   Database
+	lock *sync.Mutex
```
ejj: I think if you ditch the `lock` and just embed this, you can do `conn.Lock()` and `conn.Unlock()`  Somewhat of a go idiom

db/db.go
```diff
@@ -43,28 +41,32 @@ func New() Conn {
...
+	err := do(cn.db)
+
+	var alertTables []*table
+	for _, table := range cn.db.tables {
+		if table.shouldAlert {
+			alertTables = append(alertTables, table)
+			table.shouldAlert = false
+		}
+	}
+
```
ejj: I'd ditch this newline and add a newline after the Unlock.  Visually it's appealing to make the lock/unlock pair feel like a block.

ejj:
Two minor nits, looks good otherwise.

ejj:
Merged

PR #481, opened 2016-07-11
network: Wait for ovsdb-server on boot

ZhongtianWang:
Before this patch, network tries to boot before ovsdb-server boots,
and such behavior spawns some unnessesary warning logs. This patch
fixes this behavior by letting network try to connect to ovsdb
before it boots.

### DIFF
--- a/minion/network/network.go
+++ b/minion/network/network.go
@@ -7,6 +7,7 @@ package network
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/NetSys/quilt/db"
 	"github.com/NetSys/quilt/minion/docker"
@@ -25,6 +26,15 @@ const gatewayMAC = "02:00:0a:00:00:01"
 
 // Run blocks implementing the network services.
 func Run(conn db.Conn, dk docker.Client) {
+	for {
+		if odb, err := ovsdb.Open(); err == nil {
+			odb.Close()
+			break
+		}
+		log.Debug("Could not connect to ovsdb-server. Retry in 5 seconds.")
+		time.Sleep(5 * time.Second)
+	}
+
 	for range conn.TriggerTick(30, db.MinionTable, db.ContainerTable,
 		db.ConnectionTable, db.LabelTable, db.EtcdTable).C {
 		runWorker(conn, dk)


sirspinach:
looks good. I think this is ready for Ethan review.

sirspinach:
Your commit msg misspells "unnessesary" and is missing a period at the end though.

sirspinach:
The first line of the commit message is too long. Also, on the first boot is the worker "reconnecting" to ovsdb or connecting to ovsdb?

screen shot 2016-07-11 at 3 58 16 pm

minion/network/network.go
```diff
@@ -25,9 +25,10 @@ const gatewayMAC = "02:00:0a:00:00:01"
 
 // Run blocks implementing the network services.
 func Run(conn db.Conn, dk docker.Client) {
+	firstRun := true
```
ejj: I don't like this approach much, it's a bit ugly, plus it requires touching a lot of the code.

Why not just put an infinite loop here, before the TriggerTick loop, in which you check if you can connect ovsdb?  No point in running the master either until that works.

ZhongtianWang:
Ahh haha. I was having trouble shortening the commit message, since I couldn't break it into two lines either. Now that I had food things are getting much clearer... And infinite loop works

minion/network/network.go
```diff
@@ -25,6 +25,13 @@ const gatewayMAC = "02:00:0a:00:00:01"
 
 // Run blocks implementing the network services.
 func Run(conn db.Conn, dk docker.Client) {
+	for {
+		if odb, err := ovsdb.Open(); err == nil {
+			odb.Close()
+			break
+		}
```
ejj: Probably should sleep to avoid spinning.

ejj: Oh, also probably worth a debug level log message. Printing the error.

ZhongtianWang: whoa. Never used debug level log before. That would be useful!

ZhongtianWang:
Updated with debug level log

ejj:
Merged with tweaks.

PR #483, opened 2016-07-18
minion/etcd: Rewrite network.go be simpler

mmussomele:
network.go used to be really messy and slow, since etcd can
only write ~60 times per second. The number of writes has been
greatly reduced, to at most 3 per trigger.


mmussomele:
The patch is now ready for review. It passes all unit tests and the SparkPI example successfully runs with the new network code.

mmussomele:
The patch has now been tested on a live run of quilt, using the sparkPI spec. After fixing a bug where writing to etcd was always skipped, everything works as expected.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
 
-// A directory containers the first and seceond level of a Tree requested from the
-// etcd store.
-type directory map[string]map[string]string
+// Keeping all the store data types in a struct makes it much less verbose to pass them
+// around while operating on them
+type storeData struct {
+	containers storeContainerSlice
+	dockerIPs  map[string]string
+	labels     map[string]*storeLabel
```
ejj: Does this really need to be a `*storeLabel`?  Would a `storeLabel` do instead?

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
+}
+
+type storeContainer struct {
+	DockerID string
+	Command  []string
+	Labels   []string
+	Env      map[string]string
+}
+
+type storeContainerSlice []*storeContainer
```
ejj: Same, here, why `*storeContainer`?

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
 // Nondeterminism is hard to test.
 var rand32 = rand.Uint32
 
-// A directory containers the first and seceond level of a Tree requested from the
-// etcd store.
-type directory map[string]map[string]string
+// Keeping all the store data types in a struct makes it much less verbose to pass them
+// around while operating on them
+type storeData struct {
+	containers storeContainerSlice
```
ejj: I'd like it slightly more if we just wrote out the type, and did the cast to `storeContainerSlice` at the point that we needed it.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
 
-func runNetworkWorker(conn db.Conn, store Store) {
-	// If the directories don't exist, create them so we may watch them.  If they
-	// exist already these will return an error that we won't log, but that's ok
-	// cause the loop will error too.
-	store.Mkdir(labelDir)
-	store.Mkdir(containerDir)
-
+func runNetwork(conn db.Conn, store Store) {
+	store.Mkdir(minionDir)
```
ejj: What if this fails?

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
 		if err != nil {
-			log.WithError(err).Warn("Failed to read from cluster store.")
+			log.WithError(err).Warn()
 			continue
 		}
 
 		conn.Transact(func(view db.Database) error {
-			readContainerTransact(view, containerDir)
-			readLabelTransact(view, labelDir)
+			updateDBContainers(view, etcdData)
```
ejj: Theirs a fairly subtle concurrency bug here ...

You're reading a bunch of containers from the DB in the above transaction, and then in this transaction you go through and delete/modify those same containers.  This will cause problems if some other user of the database  deleted containers as well.  Ideally we'd reorganize this function so there's a single transaction where we do the actual etcd read and write before and after that transaction

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
+			updateDBLabels(view, etcdData)
 			return nil
 		})
 	}
 }
 
-func readContainerTransact(view db.Database, dir directory) {
-	minion, err := view.MinionSelf()
-	worker := err == nil && minion.Role == db.Worker
+func readEtcd(store Store) (*storeData, error) {
```
ejj: Why not just return `storeData`?

ejj: Actually, I'm going to stop making this comment.  In a bunch of places you use pointers to structs where just passing around the struct by value would work.  Consider it commented on at every instance =)

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
+	dockerIPs, err2 := store.Get(dockerToIPStore)
+	labels, err3 := store.Get(labelToIPStore)
 
-	for _, container := range view.SelectFromContainer(nil) {
-		container.IP = ""
-		var labels []string
-		if children, ok := dir[container.DockerID]; ok {
-			json.Unmarshal([]byte(children["Labels"]), &labels)
+	if err != nil || err2 != nil || err3 != nil {
+		err = errors.New("failed to read from etcd store")
```
ejj: I think this is a tad too clever.  Problem with it is that we're actually losing the text of the error message that was returned by `store.Get()` which would turn out to be quite useful when debugging.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
+	}
 
-			container.IP = children["IP"]
-			ip := net.ParseIP(container.IP).To4()
-			if ip != nil {
-				container.Mac = fmt.Sprintf("02:00:%02x:%02x:%02x:%02x",
-					ip[0], ip[1], ip[2], ip[3])
-			}
-		}
+	etcdContainerSlice := storeContainerSlice{}
```
ejj: Same here, just write out the type, this makes it harder to read.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
+	if err != nil {
+		log.WithError(err).Warning("Failed to update " +
+			"containers in Etcd")
 	}
 
-	for _, key := range dirKeys {
-		pairs = append(pairs, join.Pair{L: view.InsertLabel(), R: key})
+	err = updateEtcdDocker(store, etcdData, containers)
+	if err != nil {
+		log.WithError(err).Warning("Failed to update " +
```
ejj: Why these line breaks?  They seem to fit in one line.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
-	lKey := func(val interface{}) interface{} {
-		return val.(db.Label).Label
-	}
-	pairs, dbls, dirKeys := join.HashJoin(db.LabelSlice(view.SelectFromLabel(nil)),
-		join.StringSlice(dir.keys()), lKey, nil)
-
-	for _, dbl := range dbls {
-		view.Remove(dbl.(db.Label))
+func updateEtcd(store Store, etcdData *storeData, containers []db.Container) {
+	err := updateEtcdContainer(store, etcdData, containers)
```
ejj: Also, not a huge deal, but in this case you might consider this syntax:
`        if err := updateEtcdContainer(store, etcdData, containers); err != nil {`

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
-func runNetworkMaster(conn db.Conn, store Store) {
-	for range wakeChan(conn, store) {
-		leader := false
-		var containers []db.Container
-		conn.Transact(func(view db.Database) error {
-			leader = view.EtcdLeader()
-			containers = view.SelectFromContainer(nil)
-			return nil
-		})
+func updateEtcdContainer(s Store, etcdData *storeData, containers []db.Container) error {
```
ejj: This function never returns an error.  I think either, it should not log (let the upper layers), or log and return nothing.  Not sure which is better, though I think all of the update functions in this series should be consistent on this point.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
+	etcdContainers := etcdData.containers
 
-		if !leader {
-			continue
+	// Extract the relevant information from the database containers
```
ejj: I think the three comments in this function are obvious from the code and therefore unnecessary.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
-	for _, container := range containers {
-		if container.DockerID != "" {
-			ids = append(ids, container.DockerID)
+func updateEtcdDocker(s Store, etcdData *storeData, containers []db.Container) error {
+	etcdDockerMap := etcdData.dockerIPs
+
+	// Extract the relevant information from the database containers
+	dbDockerMap := map[string]string{}
+	for _, c := range containers {
+		dbDockerMap[c.DockerID] = c.IP
```
ejj: Should we skip of c.IP is ""?  Can it be ""?

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
 
-func writeStoreContainers(store Store, containers []db.Container) error {
-	var ids []string
-	for _, container := range containers {
-		if container.DockerID != "" {
-			ids = append(ids, container.DockerID)
+func updateEtcdDocker(s Store, etcdData *storeData, containers []db.Container) error {
+	etcdDockerMap := etcdData.dockerIPs
+
+	// Extract the relevant information from the database containers
```
ejj: Comment is obvious.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
+	return nil
 }
 
-func writeStoreContainers(store Store, containers []db.Container) error {
-	var ids []string
-	for _, container := range containers {
-		if container.DockerID != "" {
-			ids = append(ids, container.DockerID)
+func updateEtcdDocker(s Store, etcdData *storeData, containers []db.Container) error {
+	etcdDockerMap := etcdData.dockerIPs
```
ejj: Meh, not sure this adds any value.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
+
+	return nil
 }
 
-func writeStoreContainers(store Store, containers []db.Container) error {
-	var ids []string
-	for _, container := range containers {
-		if container.DockerID != "" {
-			ids = append(ids, container.DockerID)
+func updateEtcdDocker(s Store, etcdData *storeData, containers []db.Container) error {
```
ejj: Something about how this function seems wrong to me, but I can't quite put my finger on it . . .

At any rate, the use of dbDockerMap is confusing, if the source of truth is what's in Etcd, why are we comparing it to what's in the database?  I thought you had update the database at this point, but no, it's still the stale information.  I think where this will bite us, is in the case where Etcd and the Database disagree on what a particular IP should be.  In that case, Etcd should win, but I think in your code the Database will win.

minion/etcd/network.go
ejj: Btw, On this point, since none of these functions actually return errors, what you could do is have them take a `storeData` as an argument (instead of a `*storeData`), and then return a new `storeData`.  That way, we're getting away from mutable data.

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
+	// at the same time. The IPs are retrieved from the map of container IPs that
+	// updateEtcdDocker created.
+	containerIPs := etcdData.dockerIPs
+	dbLabelMap := map[string]*storeLabel{}
+	multiHostLabelIP := map[string]string{}
 	for _, c := range containers {
 		for _, l := range c.Labels {
-			ids = append(ids, l)
+			_, ok := dbLabelMap[l]
+			dbLabelMap[l] = &storeLabel{IP: containerIPs[c.DockerID], MH: ok}
```
ejj: This seems wrong.  If it's MultiHost, then why are we setting it to the containerIP?

ejj: Oh, next line fixes it.  That's weird . . .

minion/etcd/network.go
```diff
@@ -27,290 +32,354 @@ const gatewayIP = "10.0.0.1"
...
-		if len(dbcs) != 1 {
-			continue
-		}
+	pairs, dbcs, _ := join.HashJoin(dbContainers, etcdContainerSlice, dbKey, etcdKey)
 
-		dbc := dbcs[0]
-		if dbc.IP == "" || dir[label]["IP"] == dbc.IP {
-			continue
-		}
+	// If etcd hasn't heard of the container, clear it's data
```
ejj: Is this the right thing to d on the masters?

minion/etcd/network.go
mmussomele: Now it loops, continually trying to create the directory every five seconds until it succeeds.

minion/etcd/network.go
mmussomele: I don't think it can be "" normally, since it would have been fixed by syncIPs before and updateDBContainers would have updated the DB. The only way it could be "" would be if it was a new container that etcd doesn't know about yet, in which case it needs an IP anyway, which syncIPs will take care of.

minion/etcd/network.go
mmussomele: Etcd is the source of truth for IP addresses, but the DB knows which containers exist. So we only use the IP from etcd if the container exists in the DB. But I see what you mean by the variable name. It's been changed to be less confusing

minion/etcd/network.go
mmussomele: The reason was that it got overwritten later if it was marked as multihost. But this code has changed so this doesn't really matter anymore.

minion/etcd/network.go
mmussomele: good point, not sure why I did that actually

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
+// Keeping all the store data types in a struct makes it much less verbose to pass them
+// around while operating on them
+type storeData struct {
+	containers []storeContainer
+	dockerIPs  map[string]string
+	labels     storeLabels
+}
+
+type storeLabels struct {
+	MultiHost  map[string]string
```
ejj: All these could be lower case.

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
+	SingleHost map[string]string
+}
+
+type storeContainer struct {
+	DockerID string
```
ejj: Same for these.

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
+	Command  []string
+	Labels   []string
+	Env      map[string]string
+}
+
+type storeContainerSlice []storeContainer
```
ejj: In some ways, I think these boiler plate functions would be best left at the end of the file.  Not sure though.

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
 }
 
-func runNetworkWorker(conn db.Conn, store Store) {
-	// If the directories don't exist, create them so we may watch them.  If they
-	// exist already these will return an error that we won't log, but that's ok
-	// cause the loop will error too.
-	store.Mkdir(labelDir)
-	store.Mkdir(containerDir)
+func runNetwork(conn db.Conn, store Store) {
+	if err := store.Mkdir(minionDir); err != nil {
```
ejj: I'm not sure I agree with you that this is sufficient.  Why do you think it's going to try again later?  Shouldn't you have to write the loop to make it try again later?  Otherwise, each iteration of the below wakeChan loop is going to fail more-or-less indefinitely.

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
-		if err2 != nil {
-			err = err2
-		}
-
-		if err != nil {
-			log.WithError(err).Warn("Failed to read from cluster store.")
-			continue
-		}
+		leader := false
+		var containers []db.Container
```
ejj: I'd move this line and the line above to immediately above the conn.Transact, so it's clear they're used there.  so you'd have the readEtcd call, then a newline, then these two lines.

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
 		conn.Transact(func(view db.Database) error {
-			readContainerTransact(view, containerDir)
-			readLabelTransact(view, labelDir)
+			leader = view.EtcdLeader()
+			containers = view.SelectFromContainer(func(c db.Container) bool {
+				return c.DockerID != ""
+			})
+
+			// The leader updates both the store and the etcdData even
+			// if the read fails
```
ejj: Why?  Perhaps if the read fails, we should just give up and try again later?  Not clear to me that updating Etcd is going to be the best outcome, it may make things worse than doing nothing.

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
+			if leader {
+				etcdData = updateEtcd(store, etcdData, containers)
```
ejj: Here's an idea, think about it and let me know if it's too difficult, in which case we can do it in a separate patch.

updateEtcd essentially has two goals.  (1) update `etcdData` with new IP addresses, containers, labels and what not.  and (2) write the new `etcdData` to etcd.  What if we split those two tasks up?  (1) could happen inside the transaction, and (2) could happen immediately after. Might even make the code a bit cleaner.  What do you think?

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
+			}
+
+			if err != nil {
```
ejj: Yeah, as noted with the above point.  Unless theirs a really good reason, we can just return an error when the read fails, and move this error check up out of the transaction.

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
 	}
 }
 
-func readContainerTransact(view db.Database, dir directory) {
-	minion, err := view.MinionSelf()
-	worker := err == nil && minion.Role == db.Worker
+func readEtcd(store Store) (storeData, error) {
+	containers, err := store.Get(containerStore)
+	dockerIPs, err2 := store.Get(dockerToIPStore)
+	if err2 != nil {
```
ejj: Why so clever with the errors?  Just do the dumb thing -- check and return after each call?

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
-	var ids []string
-	for _, container := range containers {
-		if container.DockerID != "" {
-			ids = append(ids, container.DockerID)
+func updateEtcdDocker(s Store, etcdData storeData, containers []db.Container) (storeData,
+	error) {
+
+	newDockerMap := map[string]string{}
+	for _, c := range containers {
+		newDockerMap[c.DockerID] = c.IP
```
ejj: I don't think you quite got my point on the last round, the issue isn't just the name of the function.  The issue is that you're pulling the IP address for the `db.Container` as the starting point.  In most cases this will work fine, but in the case where `db.Container` is wrong about the IP address . . . i.e. `db.Container` and the `storeContainer` from etcd differ, you're going to end up choosing the IP from `db.Container` instead of the IP from `storeContainer`.

ejj: I think the right algorithm, btw, is somethin like this:

1) build a `map[string]string` out of 
2) loop through the db.Containers and set the ip to "" in the map from (1) iff there's not already an entry
3) syncIPs

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
-	for label, dbcs := range labelMap {
-		if len(dbcs) != 1 {
-			continue
+	// Etcd is the source of truth for IPs. If the label exists in both etcd and the
+	// db and it is a multihost label, then assign it the IP that etcd has.
+	// Otherwise, it stays unassigned and syncIPs will take care of it.
+	for id := range newLabels.MultiHost {
+		if ip, ok := etcdData.labels.MultiHost[id]; ok {
+			newLabels.MultiHost[id] = ip
+		} else if ip, ok = etcdData.labels.SingleHost[id]; ok {
```
ejj: huh?  Why do we assign it the SingleHost IP if it's not in the etcd MultiHost list?  This is wrong.

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
+			newLabels.MultiHost[id] = ip
 		}
+	}
 
-		dbc := dbcs[0]
-		if dbc.IP == "" || dir[label]["IP"] == dbc.IP {
-			continue
-		}
+	syncIPs(newLabels.MultiHost, net.IPv4(10, 1, 0, 0))
```
ejj: Don't we need to assign Ips for SingleHost as well?

I think this function is pretty broken, need to look through it again.

minion/etcd/network.go
```diff
@@ -27,290 +31,354 @@ const gatewayIP = "10.0.0.1"
...
+			dbc.IP = storeIP
+			ip := net.ParseIP(storeIP).To4()
+			if ip != nil {
+				dbc.Mac = fmt.Sprintf("02:00:%02x:%02x:%02x:%02x",
+					ip[0], ip[1], ip[2], ip[3])
+			}
 		}
-		dir[id] = map[string]string{}
+
+		view.Commit(dbc)
```
ejj: Each Commit is a `reflect.DeepEqual` which is rather expensive.  I might rework this loop so that you only Commit if the data actually changed.  Pretty sure that showed up when we were doing the scale testing.

minion/etcd/network.go
mmussomele: Resolved.

minion/etcd/network.go
mmussomele: Resolved.

minion/etcd/network.go
mmussomele: Resolved.

minion/etcd/network.go
mmussomele: Ah ok I understand now. Will fix.

minion/etcd/network.go
mmussomele: Agreed, needs some work

minion/etcd/network.go
mmussomele: Resolved.

minion/etcd/network.go
mmussomele: Resolved. We don't want to update the DB if the Etcd writes fail, so we need the writes in the transact.

minion/etcd/network.go
mmussomele: Resolved.

minion/etcd/network.go
```diff
@@ -27,290 +32,392 @@ const gatewayIP = "10.0.0.1"
...
+			// since we'll likely update it with bad information otherwise.
+			if wErr != nil {
+				log.WithError(wErr).Error("Failed etcd write.")
+				return nil
+			}
+
+			// If the etcd read failed, we only want to update the db if it
+			// failed because a key was missing (has not been created yet).
+			// In all other cases, we exit early.
+			if err != nil {
```
ejj: Why not just check the read error immediately after the call to the call to `readEtcd` and if it's not an `ErrorCodeKeyNotFound` then just `continue`?  In that cause, I think it's fine to leave the database un-updated, and it would make this transaction simpler.

I've noticed in this patch, you'll have a tendency to deal with an error quite a few lines of code later than it's returned.  In general, unless there's a really strong reason to, I'd like to avoid that as it makes the logic hard to follow.

minion/etcd/network.go
```diff
@@ -27,290 +32,392 @@ const gatewayIP = "10.0.0.1"
...
+			// etcd writes they perform).
+			var wErr error
+			if leader {
+				etcdData, wErr = updateEtcd(store, etcdData, containers)
+			}
+
+			// If any of the writes failed, we exit without updating the db,
+			// since we'll likely update it with bad information otherwise.
+			if wErr != nil {
+				log.WithError(wErr).Error("Failed etcd write.")
```
ejj: So one odd thing about this, is all the errors also say essentially "failed etcd write", so when you print it out here, you're going to do so twice.  Couple ways you could solve it ...

perhaps the cleanest is when updateEtcd has an error, just return it, and then outside of the transaction catch the error and log with a message like "Network Etcd Transaction Failed" or something...

minion/etcd/network.go
```diff
@@ -27,290 +32,392 @@ const gatewayIP = "10.0.0.1"
...
+			ip := net.ParseIP(storeIP).To4()
+			if ip != nil {
+				dbc.Mac = fmt.Sprintf("02:00:%02x:%02x:%02x:%02x",
+					ip[0], ip[1], ip[2], ip[3])
+			}
 		}
 
-		dbc := dbcs[0]
-		if dbc.IP == "" || dir[label]["IP"] == dbc.IP {
+		if containerEqual(dbc, oldDbc) {
```
ejj: Meh, sorry.  I told you to do this but now looking at it I think I was wrong.  Sorry!  Let's just do the commit and ditch this if check and the containerEqual function.

mmussomele:
Merged.

PR #484, opened 2016-07-19
cluster.Run should take care of nil clst the first time it runs.

ZhongtianWang:
Also, the default region for Azure should be using their backend formatted region name.


ejj:
Merged

PR #487, opened 2016-07-22
ovsdb: Redesign ovsdb with simplier interface and improved performance

ZhongtianWang:
Before this patch, there were many redundant ovsdb functions that
causes large amounts of unnecessary ovsdb transactions. That
becomes the performance bottleneck for quilt when booting a large
cluster. This patch simplifies the ovsdb interface, and reduces
O(N) number of ovsdb transactions in terms of number of VMs to O(1).

### DIFF
--- a/minion/network/network.go
+++ b/minion/network/network.go
@@ -70,26 +70,27 @@ func runMaster(conn db.Conn) {
 		return
 	}
 
-	ovsdb, err := ovsdb.Open()
+	ovsdbClient, err := ovsdb.Open()
 	if err != nil {
 		log.WithError(err).Error("Failed to connect to OVSDB.")
 		return
 	}
-	defer ovsdb.Close()
+	defer ovsdbClient.Close()
 
-	ovsdb.CreateSwitch(lSwitch)
-	lportSlice, err := ovsdb.ListPorts(lSwitch)
+	ovsdbClient.CreateLogicalSwitch(lSwitch)
+	lports, err := ovsdbClient.ListLogicalPorts(lSwitch)
 	if err != nil {
 		log.WithError(err).Error("Failed to list OVN ports.")
 		return
 	}
 
-	// The garbageMap starts of containing every logical port in OVN.  As we find
+	// The garbageMap starts off containing every logical port in OVN.  As we find
 	// that these ports are still useful, they're deleted from garbageMap until only
-	// leftover garbage ports are remaining.  These are then deleted.
-	garbageMap := make(map[string]struct{})
-	for _, lport := range lportSlice {
-		garbageMap[lport.Name] = struct{}{}
+	// leftover garbage ports are remaining. These leftover logical ports are then
+	// deleted from ovsdb.
+	garbageMap := make(map[string]ovsdb.LPort)
+	for _, lport := range lports {
+		garbageMap[lport.Name] = lport
 	}
 
 	for _, dbl := range labels {
@@ -105,9 +106,10 @@ func runMaster(conn db.Conn) {
 		log.WithFields(log.Fields{
 			"IP": dbl.IP,
 		}).Info("New logical port.")
-		err := ovsdb.CreatePort(lSwitch, dbl.IP, labelMac, dbl.IP)
-		if err != nil {
-			log.WithError(err).Warnf("Failed to create port %s.", dbl.IP)
+
+		if err := ovsdbClient.CreateLogicalPort(lSwitch, dbl.Label, labelMac,
+			dbl.IP); err != nil {
+			log.WithError(err).Warnf("Failed to create port %s.", dbl.Label)
 		}
 	}
 
@@ -120,8 +122,9 @@ func runMaster(conn db.Conn) {
 		log.WithFields(log.Fields{
 			"IP": dbc.IP,
 		}).Info("New logical port.")
-		err := ovsdb.CreatePort("quilt", dbc.IP, dbc.Mac, dbc.IP)
-		if err != nil {
+
+		if err := ovsdbClient.CreateLogicalPort(lSwitch, dbc.IP, dbc.Mac,
+			dbc.IP); err != nil {
 			log.WithFields(log.Fields{
 				"error": err,
 				"name":  dbc.IP,
@@ -131,9 +134,9 @@ func runMaster(conn db.Conn) {
 
 	// Ports still in the map don't have a corresponding label otherwise they would
 	// have been deleted in the preceding loop.
-	for lport := range garbageMap {
-		log.Infof("Delete logical port %s.", lport)
-		if err := ovsdb.DeletePort(lSwitch, lport); err != nil {
+	for _, lport := range garbageMap {
+		log.Infof("Delete logical port %s.", lport.Name)
+		if err := ovsdbClient.DeleteLogicalPort(lSwitch, lport); err != nil {
 			log.WithError(err).Warn("Failed to delete logical port.")
 		}
 	}
@@ -202,7 +205,7 @@ func updateACLs(connections []db.Connection, labels []db.Label,
 		}
 	}
 
-	acls := make(map[ovsdb.AclCore]struct{})
+	coreACLs := make(map[ovsdb.AclCore]struct{})
 
 	// Drop all ip traffic by default.
 	new := ovsdb.AclCore{
@@ -210,14 +213,14 @@ func updateACLs(connections []db.Connection, labels []db.Label,
 		Match:     "ip",
 		Action:    "drop",
 		Direction: "to-lport"}
-	acls[new] = struct{}{}
+	coreACLs[new] = struct{}{}
 
 	new = ovsdb.AclCore{
 		Priority:  0,
 		Match:     "ip",
 		Action:    "drop",
 		Direction: "from-lport"}
-	acls[new] = struct{}{}
+	coreACLs[new] = struct{}{}
 
 	for match := range matchSet {
 		new = ovsdb.AclCore{
@@ -225,34 +228,31 @@ func updateACLs(connections []db.Connection, labels []db.Label,
 			Direction: "to-lport",
 			Action:    "allow",
 			Match:     match}
-		acls[new] = struct{}{}
+		coreACLs[new] = struct{}{}
 
 		new = ovsdb.AclCore{
 			Priority:  1,
 			Direction: "from-lport",
 			Action:    "allow",
 			Match:     match}
-		acls[new] = struct{}{}
+		coreACLs[new] = struct{}{}
 	}
 
 	for _, acl := range ovsdbACLs {
 		core := acl.Core
-		if _, ok := acls[core]; ok {
-			delete(acls, core)
+		if _, ok := coreACLs[core]; ok {
+			delete(coreACLs, core)
 			continue
 		}
 
-		err := ovsdbClient.DeleteACL(lSwitch, core.Direction, core.Priority,
-			core.Match)
-		if err != nil {
+		if err := ovsdbClient.DeleteACL(lSwitch, acl); err != nil {
 			log.WithError(err).Warn("Error deleting ACL")
 		}
 	}
 
-	for acl := range acls {
-		err := ovsdbClient.CreateACL(lSwitch, acl.Direction, acl.Priority,
-			acl.Match, acl.Action, false)
-		if err != nil {
+	for aclCore := range coreACLs {
+		if err := ovsdbClient.CreateACL(lSwitch, aclCore.Direction,
+			aclCore.Priority, aclCore.Match, aclCore.Action); err != nil {
 			log.WithError(err).Warn("Error adding ACL")
 		}
 	}
--- a/minion/network/network_test.go
+++ b/minion/network/network_test.go
@@ -14,10 +14,10 @@ import (
 var testSwitch = "testSwitch"
 
 func TestACL(t *testing.T) {
-	// Inject our fake client
-	client := fakeOvsdb{}
-	ovsdb.Open = func() (ovsdb.Ovsdb, error) {
-		return &client, nil
+	client := ovsdb.NewFakeOvsdbClient()
+	client.CreateLogicalSwitch(lSwitch)
+	ovsdb.Open = func() (ovsdb.Client, error) {
+		return client, nil
 	}
 
 	defaultRules := []ovsdb.AclCore{
@@ -39,7 +39,14 @@ func TestACL(t *testing.T) {
 		containers []db.Container, coreExpACLs []ovsdb.AclCore,
 		resetClient bool) {
 		if resetClient {
-			client = fakeOvsdb{}
+			client = ovsdb.NewFakeOvsdbClient()
+			client.CreateLogicalSwitch(lSwitch)
+		}
+		updateACLs(connections, labels, containers)
+
+		res, err := client.ListACLs(lSwitch)
+		if err != nil {
+			t.Error(err)
 		}
 		coreExpACLs = append(defaultRules, coreExpACLs...)
 
@@ -51,7 +58,10 @@ func TestACL(t *testing.T) {
 		}
 
 		updateACLs(connections, labels, containers)
-		res, _ := client.ListACLs(testSwitch)
+		res, err = client.ListACLs(testSwitch)
+		if err != nil {
+			t.Error(err)
+		}
 		sort.Sort(ACLList(expACLs))
 		sort.Sort(ACLList(res))
 		if !reflect.DeepEqual(expACLs, res) {
@@ -538,39 +548,22 @@ type fakeOvsdb struct {
 	acls []ovsdb.Acl
 }
 
-func (odb *fakeOvsdb) CreateACL(lswitch string, dir string, priority int,
-	match string, action string, doLog bool) error {
-	odb.acls = append(odb.acls,
-		ovsdb.Acl{
-			Core: ovsdb.AclCore{
-				Direction: dir,
-				Priority:  priority,
-				Match:     match,
-				Action:    action,
-			},
-			Log: doLog,
-		})
+// Unused in the tests.
+func (odb *fakeOvsdb) Close() {}
+
+func (odb *fakeOvsdb) CreateSwitch(lswitch string) error {
 	return nil
 }
 
-func (odb *fakeOvsdb) DeleteACL(lswitch string, dir string, priority int,
-	match string) error {
-	for i := 0; i < len(odb.acls); i++ {
-		acl := odb.acls[i]
-		if dir != "*" && acl.Core.Direction != dir {
-			continue
-		}
-		if match != "*" && acl.Core.Match != match {
-			continue
-		}
-		if priority >= 0 && acl.Core.Priority != priority {
-			continue
-		}
-		odb.acls = append(odb.acls[:i], odb.acls[i+1:]...)
-		// Because deleting an element shifts the i+1th element into the ith
-		// spot, we need to stay at the same index to not skip the next element.
-		i--
-	}
+func (odb *fakeOvsdb) ListLogicalPorts(lswitch string) ([]ovsdb.LPort, error) {
+	return nil, nil
+}
+
+func (odb *fakeOvsdb) CreateLogicalPort(lswitch string, lport ovsdb.LPort) error {
+	return nil
+}
+
+func (odb *fakeOvsdb) DeleteLogicalPort(lswitch string, lport ovsdb.LPort) error {
 	return nil
 }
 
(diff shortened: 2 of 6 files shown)


minion/ovsdb/ovsdb_test.go
```diff
@@ -1,10 +1,1006 @@
...
-		t.Error("Condition should have length 3.")
+	"github.com/NetSys/quilt/join"
+	uuid "github.com/satori/go.uuid"
+	ovs "github.com/socketplane/libovsdb"
+)
+
+// Default OFPort used during creation of interface.
+const defaultOFPort = 123
+
+type fakeOvsdbClient struct {
```
ejj: You have this fakeOVSDB stuff in two places, once here, and once in the network test.  What if you did something similar to what the etcd code does, and make a `MockOVSDB` that's exported from the OVSDB package?  Then you can write the mock code once, and both packages can take advantage of it.

minion/network/network.go
```diff
@@ -88,9 +88,9 @@ func runMaster(conn db.Conn) {
 	// The garbageMap starts of containing every logical port in OVN.  As we find
```
mmussomele: *starts off

minion/network/network.go
```diff
@@ -88,9 +88,9 @@ func runMaster(conn db.Conn) {
...
 	// that these ports are still useful, they're deleted from garbageMap until only
 	// leftover garbage ports are remaining.  These are then deleted.
```
mmussomele: "These are then deleted"
From what?

minion/network/worker.go
```diff
@@ -515,183 +497,91 @@ func updatePorts(odb ovsdb.Ovsdb, containers []db.Container) {
...
 	for _, l := range lefts {
-		if l.(ovsPort).name == l.(ovsPort).bridge {
-			// The "bridge" port for the bridge should never be deleted
+		if l.(ovsdb.Interface).Type == ovsdb.InterfaceTypeGeneve ||
+			l.(ovsdb.Interface).Type == ovsdb.InterfaceTypeInternal {
+			// The "bridge" port and the geneve port should never be deleted
 			continue
 		}
-		if err := delPort(odb, l.(ovsPort)); err != nil {
+		if err := odb.DeleteInterface(l.(ovsdb.Interface)); err != nil {
```
mmussomele: It seems inconsistent that this refers to things as "Interface" but then the message says "port". Same for all the things below

minion/network/worker.go
```diff
@@ -968,13 +858,18 @@ func generateTargetOpenFlow(dk docker.Client, odb ovsdb.Ovsdb,
 		_, peerQuilt := patchPorts(dbc.DockerID)
 		dbcMac := dbc.Mac
 
-		ofQuilt, err := odb.GetOFPortNo(peerQuilt)
+		ofPorts, err := odb.ListOFPorts()
```
mmussomele: This call "ListOFPorts" but it returns a map, seems weird

minion/ovsdb/ovsdb.go
```diff
@@ -17,40 +15,75 @@ import (
 type Ovsdb interface {
```
mmussomele: The List* functions here should have consistent return types. Some return maps, others slices. I would either rename them for clarity or make them consistent.

ejj:
Btw, the commit message should say "Co-Authored-By: Ethan J. Jackson <ejj@eecs.berkeley.edu>" since I wrote part of the patch.

secant:
First patch should be ovsdb: Redesign ovsdb with **simpler** interface and improved performance

minion/network/network_test.go
```diff
@@ -538,121 +538,81 @@ type fakeOvsdb struct {
...
 }
 
-func (odb *fakeOvsdb) DeleteOFPort(bridge, name string) error {
-	return nil
+func (odb *fakeOvsdb) ListACLs(lswitch string) ([]ovsdb.Acl, error) {
+	var result []ovsdb.Acl
+	for _, acl := range odb.acls {
+		result = append(result, acl)
+	}
+	return result, nil
```
secant: Why not just return `odb.acls`?

minion/network/network_test.go
```diff
@@ -538,121 +538,81 @@ type fakeOvsdb struct {
...
+			continue
+		}
+		if acl.Core.Match != "*" && odbACL.Core.Match != acl.Core.Match {
+			continue
+		}
+		if acl.Core.Priority >= 0 &&
+			odbACL.Core.Priority != acl.Core.Priority {
+			continue
+		}
+		odb.acls = append(odb.acls[:i], odb.acls[i+1:]...)
```
secant: Does this mean that there can be duplicate ACLs?

minion/network/network_test.go
secant: (Commenting here because top comment got hidden)

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,724 @@ type AclCore struct {
...
 	}
 
-	bridge := make(map[string]interface{})
-	bridge["name"] = lswitch
-
 	insertOp := ovs.Operation{
 		Op:    "insert",
 		Table: "Logical_Switch",
-		Row:   bridge,
+		Row:   map[string]interface{}{"name": lswitch},
```
secant: Can `Row` just be `bridge` here?

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,724 @@ type AclCore struct {
...
+		Where: []interface{}{ovs.NewCondition("name", "==", lswitch)},
+	})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("transaction error: retrieving lswitch %s: %s",
+			lswitch, err)
 	}
-	if len(results) <= 0 {
-		return acls, nil
+	if len(switchReply[0].Rows) <= 0 {
```
secant: Will this ever be <0? (same questions applies to other instances of this)

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,724 @@ type AclCore struct {
...
-	return val.(string), nil
-}
 
-// GetOFInterfaceAttachedMAC gets the attached-mac of the interface given a Row
-func (ovsdb Client) GetOFInterfaceAttachedMAC(iface Row) (string, error) {
-	val, err := ovsdb.getFromMapInRow(iface, "external_ids", "attached-mac")
-	if err != nil {
-		return "", err
+	if len(ops) == 0 {
+		return nil
```
secant: You could check if `len(muts) == 0` at the top of the for loop

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,724 @@ type AclCore struct {
...
+
+	ifaceType, ok := row["type"].(string)
+	if !ok {
+		return iface, errors.New("interface missing type")
+	}
+	iface.Type = ifaceType
+
+	optRow, ok := row["options"]
+	if !ok {
+		return iface, errors.New("missing Interface key: options")
```
secant: Should make the past three error messages consistent

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,724 @@ type AclCore struct {
...
+		log.Debug("interface missing attached-mac.")
+	}
+
+	if id, ok := externalIDs["iface-id"]; ok {
+		iface.IfaceID = id
+	} else {
+		log.Debug("interface missing iface-id.")
+	}
+
+	return iface, nil
```
secant: This block is pretty redundant, is it worth it to simplify (like pass in a list of keys to check or something)?

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,724 @@ type AclCore struct {
...
-			name, peer, err)
+// This does not cover all cases, they should just be added as needed
+func newMutation(column, mutator string, value interface{}) mutation {
+	switch typedValue := value.(type) {
+	case ovs.UUID:
+		uuidSlice := []ovs.UUID{typedValue}
+		mutateValue, err := ovs.NewOvsSet(uuidSlice)
+		if err != nil {
+			// An error can only occur if the input is not a slice
+			panic("newMutation(): an impossible error occurred")
```
secant: This sounds weird :P

minion/ovsdb/ovsdb_mock.go
```diff
@@ -0,0 +1,447 @@
...
+	}
+
+	conditions := parseOperationConditions(op)
+
+	result := ovs.OperationResult{}
+	for _, row := range table.rows {
+		for _, cond := range conditions {
+			if cond.matches(row) {
+				result.Rows = append(result.Rows, row)
+				continue
```
secant: I don't think you need this `continue`

minion/ovsdb/ovsdb_mock.go
```diff
@@ -0,0 +1,447 @@
...
+	}
+
+	table, ok := db.tables[op.Table]
+	if !ok {
+		return ovs.OperationResult{}, nil
+	}
+
+	conditions := parseOperationConditions(op)
+	mutations := parseOperationMutations(op)
+	mutationCount := 0
```
secant: You don't update this anywhere

minion/ovsdb/ovsdb_test.go
```diff
@@ -1,10 +1,570 @@
...
+	if err := ovsdbClient.CreateSwitch(lswitch); err == nil {
+		t.Error("create duplicate switch did not yield an error")
+	}
+}
+
+func TestLogicalPorts(t *testing.T) {
+	ovsdbClient := NewFakeOvsdbClient()
+
+	checkCorrectness := func(ovsdbLPorts []LPort, localLPorts ...LPort) {
+		scoreFun := func(left, right interface{}) int {
```
secant: I think this function can be defined outside of `checkCorrectness`

minion/ovsdb/ovsdb_test.go
```diff
@@ -1,10 +1,570 @@
...
+	if len(ovsdbLPorts) != 0 {
+		t.Errorf("expected 0 logical port. Got %d instead.", len(ovsdbLPorts))
+	}
+}
+
+func TestACLs(t *testing.T) {
+	ovsdbClient := NewFakeOvsdbClient()
+
+	checkCorrectness := func(ovsdbACLs []Acl, localACLs ...Acl) {
+		key := func(val interface{}) interface{} {
```
secant: `key` can be defined outside this function

minion/ovsdb/ovsdb_test.go
```diff
@@ -1,10 +1,570 @@
...
+	checkCorrectness := func(ovsdbIfaces []Interface, localIfaces ...Interface) {
+		key := func(val interface{}) interface{} {
+			iface := val.(Interface)
+			return struct {
+				name, bridge string
+			}{
+				name:   iface.Name,
+				bridge: iface.Bridge,
+			}
+		}
```
secant: See previous comment

minion/ovsdb/ovsdb_test.go
```diff
@@ -1,10 +1,570 @@
...
+
+	// Create a new Bridge.
+	_, err = ovsdbClient.transact("Open_vSwitch", ovs.Operation{
+		Op:    "insert",
+		Table: "Bridge",
+		Row:   map[string]interface{}{"name": lswitch2},
+	})
+	if err != nil {
+		t.Error(err)
+	}
```
secant: Is there any way to check a bridge was actually created?

secant: Oh, bridge = iface?

minion/ovsdb/ovsdb_test.go
```diff
@@ -1,10 +1,570 @@
...
+
+	portMap, err := ovsdbClient.GetOFPortsMap()
+	if err != nil {
+		t.Error(err)
+	}
+
+	if port, ok := portMap[iface1.Name]; !ok || port != defaultOFPort {
+		t.Error("GetOFPortsMap failed listing all ports.")
+	}
```
secant: Is it worth it to check the contents of `portMap`?

minion/ovsdb/ovsdb.go
ZhongtianWang: Only Row is part of the ovs Operation struct

minion/ovsdb/ovsdb.go
ZhongtianWang: It should be changed to `==` :D

minion/ovsdb/ovsdb.go
ZhongtianWang: Since it's a `for` loop, if `muts` is zero length, it will be ignored

minion/ovsdb/ovsdb.go
ZhongtianWang: We can use JSON unmarshalling. But then it would be hard to figure out what exactly is missing. I think it should be fine like this. What do you think?

minion/ovsdb/ovsdb.go
ZhongtianWang: This was actually old code, just changed place. So yeah, there is no way it would have error, and other people's code just ignore the error as well, so I guess we could just ignore it too.

minion/ovsdb/ovsdb_mock.go
ZhongtianWang: Oops, nice catch!

minion/ovsdb/ovsdb_test.go
ZhongtianWang: Good point. Changed

minion/ovsdb/ovsdb_test.go
ZhongtianWang: I see why you think that way. So `GetInterfaceMap` returns a map of interfaces. The key of the map is the bridge name, so `if len(ifaceMap) != 2` is actually checking whether it successfully created 2 bridges.

minion/ovsdb/ovsdb_test.go
ZhongtianWang: Yeah, it checked by checking whether port from portMap equals defaultOFPort

minion/ovsdb/ovsdb.go
ejj: Do you need this interface at all still?  There's only one implementation ...

minion/ovsdb/ovsdb.go
```diff
@@ -17,40 +15,75 @@ import (
...
+	DeleteInterface(iface Interface) error
+	ModifyInterface(iface Interface) error
+	SetBridgeMac(bridge, mac string) error
+	GetOFPortsMap() (map[string]int, error)
+}
+
+// This is simply a wrapper around all the libovsdb calls we are currently using. This
+// makes it very easy to mock for testing.
+type ovsdbAPI interface {
+	disconnect()
```
ejj: If you capitalize these, so it's `Disconnect` and `Transact` . . . do you need `ovsdbClient`?  *ovs.OvsdbClient already implements them right?

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,720 @@ type AclCore struct {
...
-	}
-	return nil
+	ovsdb.disconnect()
 }
 
 // CreateSwitch creates a new logical switch in OVN.
 func (ovsdb Client) CreateSwitch(lswitch string) error {
-	check, err := ovsdb.selectRows("OVN_Northbound", "Logical_Switch",
-		newCondition("name", "==", lswitch))
+	check, err := ovsdb.transact("OVN_Northbound", ovs.Operation{
```
ejj: Is this check necessary?  If we attempt to create the switch, and it already exists, won't the transaction below just fail?  That should save us a round trip to the database which would be a big win.

You do this in several places btw, most importantly in CreateLogicalPort, I'd clean all of them up.

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,720 @@ type AclCore struct {
...
-	}
-	if totalCount != expectedCount {
-		return errors.New("unexpected number of rows altered")
-	}
-	return nil
+	ovsdb.disconnect()
 }
 
 // CreateSwitch creates a new logical switch in OVN.
 func (ovsdb Client) CreateSwitch(lswitch string) error {
```
ejj: Btw, I'd rename this `CreateLogicalSwitch` so it's consistent with `CreateLogicalPort`

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,720 @@ type AclCore struct {
...
 	}
-	return lports, nil
+	return result, nil
 }
 
-// CreatePort creates a new logical port in OVN.
-func (ovsdb Client) CreatePort(lswitch, name, mac, ip string) error {
+// CreateLogicalPort creates a new logical port in OVN.
+func (ovsdb Client) CreateLogicalPort(lswitch string, lport LPort) error {
 	// OVN Uses name an index into the Logical_Port table so, we need to check
```
ejj: Ah, yeah I don't think this comment is entirely correct.  I think that the transaction will fail if a port with the same name already exists.  Could you verify?

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,720 @@ type AclCore struct {
...
-				Log: result["log"].(bool),
-			}
-			acls = append(acls, acl)
+	logicalSwitch := switchReply[0].Rows[0]
+
+	aclReply, err := ovsdb.transact("OVN_Northbound", ovs.Operation{
+		Op:    "select",
+		Table: "ACL",
+		Where: []interface{}{
+			ovs.NewCondition("_uuid", "!=", ovs.UUID{GoUUID: "_"}),
```
ejj: This is weird.  Did you ever figure out why we need it?  Minimally, it deserves a comment . . .

ejj: Also, it might be worth making a constant called `NoCondition` or something.  You could use it all over to save a bunch of useless typing.

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,720 @@ type AclCore struct {
...
+	return errorCheck(results, 2)
 }
 
-// DeleteOFPort deletes an openflow port with the corresponding bridge and
-// port names.
-func (ovsdb Client) DeleteOFPort(bridge, name string) error {
-	rows, err := ovsdb.selectRows("Open_vSwitch", "Port",
-		newCondition("name", "==", name))
+// GetInterfacesMap gets all openflow interfaces mapped by bridge.
+func (ovsdb Client) GetInterfacesMap() (map[string][]Interface, error) {
```
ejj: I think this function would be simpler if you just returned a `[]Interface`, if the caller wants to convert it to a map, they're welcome to, but it's kind of weird to add that logic to the ovsdb client.

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,720 @@ type AclCore struct {
...
+	return result, nil
 }
 
-// CreateOFPort creates an openflow port on specified bridge
+// CreateInterface creates an openflow port on specified bridge.
 //
 // A port cannot be created without an interface, that is why the "default"
 // interface (one with the same name as the port) is created along with it.
-func (ovsdb Client) CreateOFPort(bridge, name string) error {
+func (ovsdb Client) CreateInterface(iface Interface) error {
```
ejj: Why not just take a bridge and a name?  Seems cleaner to me...  Why would the caller have a handle on an iface?

ejj: Note, I know `minion/network/worker.go` does create a handle to the interface, but that code is weird, it shouldn't do that . . .

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +100,720 @@ type AclCore struct {
...
-	results, err := ovsdb.selectRows("Open_vSwitch", "Bridge",
-		newCondition("name", "==", bridge))
-	if err != nil {
-		return nil, err
+// DeleteInterface deletes an openflow interface.
+func (ovsdb Client) DeleteInterface(iface Interface) error {
+	deleteOp := ovs.Operation{
+		Op:    "delete",
+		Table: "Port",
+		Where: []interface{}{ovs.NewCondition("name", "==", iface.Name)},
```
ejj: Btw, even for this condition, i.e. a "name == something" would be nice to have a helper.

ejj:
Oh, also I think the first patch should say it's co-authored by me, since I wrote the initial starting point before handing it off.

ejj:
Merged

ejj:
Oops, actually merged the other one.

minion/ovsdb/ovsdb.go
ZhongtianWang: Since in `worker.go` we are already joining target ports and current ports, which are all interface slices, it makes sense to just pass `interface` to `CreateInterface` since we already have it. We could take the `bridge, name` separately instead, but it seems even messier by passing `iface.Bridge` and `iface.Name` instead.

minion/ovsdb/ovsdb.go
ZhongtianWang: Agreed

minion/ovsdb/ovsdb.go
ZhongtianWang: Even it's not required by ovsdb spec, `libovsdb` package does not seem to support it. I checked other people's code, and they all did it this way to make it work :/

ZhongtianWang:
hmmm? I've already added co-author in the commit message. I added it as a separate line in commit message. Or should I do it in some other format?

minion/ovsdb/ovsdb.go
ZhongtianWang: Yeah. It uses name as index. And it would fail if port with the same name exists.

ejj:
Ah my bad, the Co-authored by looks good.

minion/ovsdb/ovsdb.go
ejj: Sounds good, just add a comment then explaining?

minion/ovsdb/ovsdb.go
ejj: Meh, I think worker.go is weird, no reason for that weirdness to bleed into the library.  Think of it this way, if you didn't know that worker.go behaved in that odd way, would you write this function with this interface?  I dunno, I like the way it was better . . .

minion/ovsdb/ovsdb.go
ZhongtianWang: I thought it was necessary since `CreateSwitch` is called periodically. I do agree with you on other functions like `CreateLogicalPort` that we do not need to check. But for `CreateSwitch`, I think it should be fine to leave as it is. It only gives one extra RTT the first time Quilt runs. After that it just check whether we already created one already. What do you think?

minion/ovsdb/ovsdb.go
ZhongtianWang: I just realized that it doesn't need to be `[]interface{}{ovs.NewCondition("name", "==", iface.Name)}`. Instead, just `ovs.NewCondition("name", "==", iface.Name)` is enough. I don't really think a helper just for this is that necessary.

minion/ovsdb/ovsdb.go
ZhongtianWang: Yeah, we need it for `worker`, isn't it? Otherwise, how can worker call our ovsdb functions?

ZhongtianWang: Ohhhh you are right. LOL we don't need this.

minion/ovsdb/ovsdb.go
ZhongtianWang: We need to mock Transact for unit test. And `ovsdbClient` is the actual client to libovsdb.

minion/ovsdb/ovsdb.go
ejj: I really do think you need to change this . . .  It's particularly odd that CreateLogicalPort doesn't take an Lport but this takes an Interface, the library is consistent at the moment . . .

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +81,694 @@ type AclCore struct {
...
-// be wildcarded by passing a value less than 0
-func (ovsdb Client) CreateACL(lswitch string, dir string, priority int, match string,
-	action string, doLog bool) error {
-	acl := make(map[string]interface{})
-	if dir != "*" {
-		acl["direction"] = dir
+// direction and match may be wildcarded by passing the value "*". priority may also
+// be wildcarded by passing a value less than 0.
+func (ovsdb Client) CreateACL(lswitch string, direction string, priority int,
+	match string, action string, doLog bool) error {
```
ejj: I actually wouldn't take `doLog` as an argument, and just always pass in `false` below.  The logging feature is stupid and we'll never use it, I don't want to expose it outside of this library.

minion/ovsdb/ovsdb.go
```diff
@@ -67,779 +81,694 @@ type AclCore struct {
...
 	}
-	return errorCheck(results, 1, 1)
+	return errorCheck(results, 1)
 }
 
-func (ovsdb Client) updateOFInterface(name string, update Row) error {
-	updateOp := ovs.Operation{
-		Op:    "update",
+// GetOFPortsMap get all the openflow ports mapped by bridge.
+func (ovsdb Client) GetOFPortsMap() (map[string]int, error) {
```
ejj: So here's an idea, what if you add an `ofport` field to the `Interface` which you could populate as part of the `ListInterfaces()` function.  If you do that, you can get rid of this function all together, instead having the caller simply invoke ListInterfaces() and build the map as it desires.

minion/ovsdb/ovsdb.go
ZhongtianWang: whoa! How did I just missed that...

minion/ovsdb/ovsdb.go
ZhongtianWang: Gotcha!

ejj:
Merged

PR #490, opened 2016-07-25
Rewrite tester

secant:


quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,615 @@
...
+
+	p := payload{
+		Channel:   os.Getenv("SLACK_CHANNEL"),
+		Color:     color, // Green
+		Pretext:   pretext,
+		Username:  "quilt-bot",
+		Iconemoji: iconemoji,
+		Fields:    []field{f}}
+
+	hookurl := "https://hooks.slack.com/services/T04Q3TL41/B0M25TWP5/soKJeP5HbWcjk" +
```
kklin: If someone has this URL, then they can post whatever they want to our slack right? Should probably be configured through an environment variable eventually..

quilt-tester/Dockerfile
```diff
@@ -1,12 +1,11 @@
...
-pip install awscli
+RUN apk add --no-cache bash git python curl openssh-client
 Copy bin/* /bin/
-Copy config/config.spec /config/
 Copy tests /tests
 Copy config/id_rsa /root/.ssh/id_rsa
 RUN chmod 0600 /root/.ssh/id_rsa
-
+RUN mkdir /.quilt && chmod -R 777 /.quilt && mkdir /.quilt/config
+Copy config/infrastructure.spec /.quilt/config/infrastructure.spec
```
ZhongtianWang: I don't think `config/infrastructure.spec` actually exists at this point. Are you referring to `config.spec`?

ZhongtianWang:
In `Makefile`'s `docker-build-tester` target, would be useful if we could include `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/quilt-tester quilt-tester.go`

quilt-tester/Dockerfile
secant: Whoops I forgot to add it into my commit. Sorry about that!

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,615 @@
...
+		t, _ := ioutil.ReadAll(resp.Body)
+		l.logInfo("Error posting to Slack.", infoMsg)
+		l.logInfo(string(t), errMsg)
+	}
+}
+
+func downloadSpecs(importPath string) {
+	quiltPath := os.Getenv("QUILT_PATH")
+	l.logInfo(fmt.Sprintf("Downloading %s into %s", importPath, quiltPath), infoMsg)
+	cmd := exec.Command("/quilt", "get", importPath)
```
ZhongtianWang: For some reason, when I'm trying to run it, the tester halts here ^^^. Do you know what could be happening?

secant: Probably something along permission denied (because you're writing to `/`). You should change the QUILT_PATH to be somewhere else if you're trying to run locally ^^

ZhongtianWang: hmmm why does it makes a difference? If there is a permission problem, quilt should have quitted already and this command should finish with error. What do you have as QUILT_PATH when you test locally?

secant: Ah sorry! You need to change the command to be `exec.Command("quilt"...` instead of `exec.Command("/quilt")` The Dockerfile copies over the quilt binary to `/`.

secant: Ah sorry! You need to change the command to be `exec.Command("quilt"...` instead of `exec.Command("/quilt")` The Dockerfile copies over the quilt binary to `/`.

ZhongtianWang: I don't think `/quilt` is the problem, since `/` it's not in PATH, it fine to run it using `/quilt`.

ZhongtianWang: Wait I don't think this is the problem. I tried using the tester container's bash to run quilt, but it doesn't even support `quilt get`.

secant: Oh, you're doing this in a container? I'm not sure why it's erroring then... `/quilt get` should work, because `quilt` is not in PATH, but in `/`.

secant: What's the error message you're getting?

ZhongtianWang: Agh my bad. I have an outdated quilt docker image on my machine.

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,615 @@
...
+			panic(err)
+		}
+		return output, fmt.Errorf("Quilt timed out while waiting for "+
+			"%s. Output can be found in container.log.", trigger)
+
+	case err := <-done:
+		if err != nil {
+			return output, err
+		}
+		return output, nil
```
ZhongtianWang: Can just be `return output, err`

ZhongtianWang:
I think we should keep the binary files out, and they can just be generated in tester. It would be easier to add more tests in the future.

quilt-tester/config/infrastructure.spec
```diff
@@ -0,0 +1,12 @@
+// We will have three worker machines.
+(define nWorker 3)
+
+(define Namespace "tester-52-53-247-146")
```
ZhongtianWang: Is there any specific reason why the namespace is `tester-52-53-247-146`?

ejj: Yeah, we need to fix this before merging.

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+var configSpec, configSpecName, namespace, testRoot, specDir, url string
```
mmussomele: I would use something like
`var (
         ....
)`

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+}
+
+func verbMsg(msg string) string {
+	return "\n" + msg + "\n"
+}
+
+func writeTo(file string, message string) error {
+	f, err := os.OpenFile(file, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
+	if err != nil {
+		fmt.Println("Could not open", file)
```
mmussomele: Should this be a log call?

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+	failedDir   string
+	logDir      string
+	currTest    string
+	container   string
+	quiltTester string
+}
+
+type format func(string) string
+
+func (l *log) logInfo(msg string, fn format) {
```
mmussomele: I think it would be simpler if these three log function just took in the message and left the caller responsible for the formatting call

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+	}
+}
+
+func (l *log) logContainer(msg string, fn format) {
+	if err := writeTo(l.container, fn(msg)); err != nil {
+		panic(err)
+	}
+}
+
+func infoMsg(msg string) string {
```
mmussomele: I would make this `infoMessage`. Same for the ones below.

mmussomele: Actually, what would be even better is a name that indicats that this function reformats the string.

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+		fmt.Println("Could not open", file)
+		return err
+	}
+
+	defer f.Close()
+	_, err = f.WriteString(message)
+	return err
+}
+
+func fileContents(file string) (string, error) {
```
mmussomele: maybe `readFileContents`?

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+
+	toDelete := &ec2.TerminateInstancesInput{
+		InstanceIds: ids,
+	}
+
+	req, _ := svc.TerminateInstancesRequest(toDelete)
+	return req.Send()
+}
+
+// End Cleanup Functions
```
mmussomele: These comments are unneeded. Instead, I would put individual comments on each function if you need to describe their purpose.

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+		return err
+	}
+
+	l.logInfo(contents, verbMsg)
+	l.logInfo("End "+ts.name+".spec", infoMsg)
+	_, cmd, _ := runSpec(ts.spec)
+
+	// Wait for the containers to start
+	l.logInfo("Waiting 5 minutes for containers to start up", infoMsg)
+	time.Sleep(300 * time.Second)
```
mmussomele: `time.Sleep(5 * time.Minute)`?

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+var configSpec, configSpecName, namespace, testRoot, specDir, url string
+var l log
+var suitesPassed, suitesFailed []string
+var machines []map[string]string
```
mmussomele: I don't really like that these are all just floating global variables. Seems pretty bug-prone. Is there a reason they can't be passed around as function arguments?

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+func runTest(test string, m map[string]string) error {
+	dir, file := filepath.Split(test)
+	contents, err := fileContents(filepath.Join(dir, "src", file, file) + ".go")
+	if err != nil {
+		l.logInfo(fmt.Sprintf("Could not read test %s", test), infoMsg)
+		l.logInfo(err.Error(), errMsg)
+		return err
+	}
+
+	testLog := file + "-" + m["publicIP"] + ".txt"
```
mmussomele: I personally prefer `fmt.Sprintf` to string concatentation, but maybe that's just me.

quilt-tester/quilt-tester.go
mmussomele: Also is there a reason machines need to be a map? Why can't we just reuse `db.Machine` or create a new struct type?

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+	case err := <-done:
+		return output, err
+	}
+}
+
+func generateTestSuites(testDir string) []testSuite {
+	var suites []testSuite
+	// First, we need to ls the testDir, and find all of the folders. Then we can
+	// generate a testSuite for each folder.
+	testFiles, err := filepath.Glob(filepath.Join(testDir, "*"))
```
mmussomele: could you use https://golang.org/pkg/io/ioutil/#ReadDir?

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+			err = errors.New("failed test")
+		}
+	}
+
+	l.logTest("Begin test source", infoMsg)
+	l.logTest(contents, verbMsg)
+	l.logTest("End test source", infoMsg)
+	l.logTest("Begin test output", infoMsg)
+	l.logTest(string(output), verbMsg)
+	fmt.Println(string(output))
```
mmussomele: why log and then print?

mmussomele: also should these message print as the events they report actually happen?

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+
+func runSpec(spec string) (string, *exec.Cmd, error) {
+	cmd := exec.Command("/quilt", "run", spec)
+	if spec == configSpecName {
+		output, err := runAndWaitFor(cmd, "New connection.", 500)
+		return output, cmd, err
+	}
+	l.logContainer(fmt.Sprintf("Running %v for 60 seconds", cmd.Args), infoMsg)
+	cmd.Start()
+	time.Sleep(60 * time.Second)
```
mmussomele: `time.Minute`?

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+	l.logTest(contents, verbMsg)
+	l.logTest("End test source", infoMsg)
+	l.logTest("Begin test output", infoMsg)
+	l.logTest(string(output), verbMsg)
+	fmt.Println(string(output))
+	l.logTest("End test output", infoMsg)
+	return err
+}
+
+func runSpec(spec string) (string, *exec.Cmd, error) {
```
mmussomele: I think this should just be 2 functions, since whenever you call it, you know whether the spec is the config spec already and there really isn't any shared code

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+	pipe, err := cmd.StderrPipe()
+	if err != nil {
+		panic(err)
+	}
+
+	if err := cmd.Start(); err != nil {
+		panic(err)
+	}
+
+	done := make(chan error, 1)
```
mmussomele: What if the command finishes and send to `done`, then the goroutine eats the `done` signal? The select at the bottom will just be left waiting around for the timer to go off. Or worse, the select at the bottom eats the error from `cmd.Wait()` and the goroutine is left running potentially forever.

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+				}
+			}
+		}
+	}()
+
+	select {
+	case <-timer:
+		done <- nil
+		if err := cmd.Process.Kill(); err != nil {
+			panic(err)
```
mmussomele: Why do you panic everywhere? If you want to exit it may be cleaner to use https://golang.org/pkg/os/#Exit

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+	}
+	l.logContainer(fmt.Sprintf("Running %v for 60 seconds", cmd.Args), infoMsg)
+	cmd.Start()
+	time.Sleep(60 * time.Second)
+	return "", cmd, nil
+}
+
+func runAndWaitFor(cmd *exec.Cmd, trigger string, timeout int) (string, error) {
+	l.logContainer(fmt.Sprintf("Running %v until %s", cmd.Args, trigger), infoMsg)
+	output := ""
```
mmussomele: this isn't used for a while, maybe it should be declared right before the goroutine like infoStr

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+			case <-done:
+				return
+			default:
+				pipe.Read(info)
+				// Remove trailing null chracters
+				infoStr = string(bytes.Trim(info, "\x00"))
+				if len(infoStr) == 0 {
+					continue
+				}
+				fmt.Println(infoStr)
```
mmussomele: Should this be a log?

quilt-tester/quilt-tester.go
mmussomele: Actually, even better, you could create new loggers and set their output to be the respective files.

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+		if info.IsDir() {
+			specs, err := filepath.Glob(filepath.Join(testPath, "*.spec"))
+			if err != nil {
+				l.logInfo(fmt.Sprintf("Could not get spec file for %s",
+					testPath), infoMsg)
+				l.logInfo(err.Error(), errMsg)
+				continue
+			}
+			spec := specs[0]
+			allFiles, err := filepath.Glob(filepath.Join(testPath, "*"))
```
mmussomele: same here with ioutil.ReadDir

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+		cleanup()
+		return err
+	}
+
+	populateMachines(output)
+	l.logInfo("Booted Quilt", infoMsg)
+	l.logInfo("Machines", infoMsg)
+	l.logInfo(fmt.Sprintf("%v", machines), verbMsg)
+	l.logInfo("Wait 5 minutes for containers to start up", infoMsg)
+	time.Sleep(300 * time.Second)
```
mmussomele: `time.Sleep(5*time.Minute)`?

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+func populateMachines(output string) {
+	grabbed := false
+	numM := 0
+	numRe := regexp.MustCompile(`count=(\d)`)
+	numM, _ = strconv.Atoi(numRe.FindStringSubmatch(output)[1])
+	machineRe := regexp.MustCompile(`Machine-(\d+){Role=(.*), ` +
+		"Provider=(.*), Region=(.*), Size=(.*), DiskSize=(.*), CloudID=(.*), " +
+		"PublicIP=(.*), PrivateIP=(.*)}")
+	bootsRe := regexp.MustCompile("Successfully booted machines.")
+	boots := bootsRe.FindAllStringIndex(output, -1)
```
mmussomele: I think this section might be more readable if you first compile all the regex matches, then extract the information out of the output with them. Or group these according to the information you're trying to extract with blank lines.

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+	output = output[boots[len(boots)-1][0]:] // Trim to the last successful boot
+	lines := strings.Split(output, "\n")
+	for _, line := range lines {
+		if strings.Contains(line, "db.Machine") {
+			grabbed = true
+		}
+
+		if grabbed && len(machines) >= numM {
```
mmussomele:
```
if !grabbed {
        continue
}
```

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+		"PublicIP=(.*), PrivateIP=(.*)}")
+	bootsRe := regexp.MustCompile("Successfully booted machines.")
+	boots := bootsRe.FindAllStringIndex(output, -1)
+	output = output[boots[len(boots)-1][0]:] // Trim to the last successful boot
+	lines := strings.Split(output, "\n")
+	for _, line := range lines {
+		if strings.Contains(line, "db.Machine") {
+			grabbed = true
+		}
+
```
mmussomele:
```
if !grabbed {
        continue
}
```

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+			m := make(map[string]string)
+			m["name"] = matches[1]
+			m["role"] = matches[2]
+			m["provider"] = matches[3]
+			m["region"] = matches[4]
+			m["cloudID"] = matches[5]
+			m["size"] = matches[6]
+			m["disksize"] = matches[7]
+			m["publicIP"] = matches[8]
+			m["privateIP"] = matches[9]
```
mmussomele: These slice accesses don't match up with the subexpressions from the regex. i.e. cloudID contains the size of the machine in the end.

quilt-tester/quilt-tester.go
```diff
@@ -0,0 +1,613 @@
...
+		panic(err)
+	}
+
+	// Get our specs
+	os.Setenv("QUILT_PATH", "/.quilt")
+	downloadSpecs("github.com/NetSys/quilt")
+	generateNamespace()
+	suites := generateTestSuites(testDir)
+	runTestSuites(suites)
+}
```
mmussomele: Overall, the biggest feeling I get from the tester is that it still feels like a script. Many of the ways that things are done are reminiscent of a python or bash script, probably because the tester was translated instead of rewritten to fit Go style. It also feels like you reinvent some tools we already have, like the logger.
In addition, I think the readability of many sections of the code could benefit from some blank lines between distinct sections of functions.

ejj: This is a reasonable complaint, though on some level I want to move on from this patch. If th code is significantly better than what we had with the bash scripts, then let's just get it as nice as we can with a reasonably low amount of effort, and revise later.  Since it's not production code, I'm a bit less worried about it being perfect quality.

quilt-tester/tests/basic/src/docker_test/docker_test.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+func main() {
+	output, err := exec.Command("docker", "ps", "-a").Output()
+	if err != nil {
+		fmt.Println(output)
```
mmussomele: Log and return/os.exit?

quilt-tester/tests/basic/src/log_test/log_test.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+func main() {
+	output, err := exec.Command("docker", "logs", "minion").CombinedOutput()
+	if err != nil {
+		fmt.Println(output)
```
mmussomele: Same here

quilt-tester/tests/spark/src/spark_test_monly/spark_test_monly.go
```diff
@@ -0,0 +1,50 @@
...
+	"os"
+	"os/exec"
+	"strings"
+)
+
+func main() {
+	// Grab the container with master
+	output, err := exec.Command("swarm", "ps", "-a").Output()
+	if err != nil {
+		panic(err)
```
mmussomele: Same here with the log/return. Is there a reason this one doesn't print the output?

quilt-tester/quilt-tester.go
mmussomele: Agreed, this should be fixed.

ejj:
This PR was folded into #498

PR #491, opened 2016-07-27
Makefile: Add clean-coverage target

mmussomele:
Whenever we run coverage, the main directory of quilt gets filled
with all kinds of coverage metadata that interferes with
autocomplete. This patch adds a simple target to the Makefile that
deletes said coverage data.


Makefile
```diff
@@ -35,6 +35,10 @@ COV_PKG = $(subst github.com/NetSys/quilt,,$(PACKAGES))
 coverage: $(addsuffix .cov, $(filter-out $(COV_SKIP), $(COV_PKG)))
 	gover
 
+clean-coverage:
```
ejj: I'd just make it a `clean` target that deletes all of the build artifacts, coverage stuff included

mmussomele: Good idea, done!

ejj:
Merged

PR #493, opened 2016-07-27
Invariants

akshayknarayan:
This PR is two small features.
1. Don't allow DFS to work through the "public" label (bug fix)
2. Add an invariant type that checks that two labels are directly connected.

For the second feature, it's possible that in the future this invariant and the main reach invariant could be combined in an invariant that takes a depth argument, but (imho) this isn't worth doing right now.

### DIFF
--- a/stitch/eval.go
+++ b/stitch/eval.go
@@ -12,6 +12,7 @@ type evalCtx struct {
 	placements  map[Placement]struct{}
 	machines    *[]*astMachine
 	invariants  *[]invariant
+	annotations map[string]string
 
 	containers  *[]astContainer
 	containerID *int
@@ -44,6 +45,7 @@ func (ctx *evalCtx) deepCopy() *evalCtx {
 		containers:  ctx.containers,
 		placements:  ctx.placements,
 		invariants:  ctx.invariants,
+		annotations: ctx.annotations,
 		containerID: ctx.containerID,
 		parent:      parentCopy,
 	}
@@ -299,6 +301,7 @@ func newEvalCtx(parent *evalCtx) evalCtx {
 		make(map[Placement]struct{}),
 		&[]*astMachine{},
 		&[]invariant{},
+		make(map[string]string),
 		&[]astContainer{},
 		&id}
 }
--- a/stitch/func.go
+++ b/stitch/func.go
@@ -17,6 +17,8 @@ type funcImpl struct {
 	lazy    bool // True if arguments should not be evaluated automatically.
 }
 
+var annotations map[string]struct{}
+
 var funcImplMap map[astIdent]funcImpl
 
 // We have to initialize `funcImplMap` in an init function or else the compiler
@@ -31,6 +33,10 @@ func init() {
 	less := compareFun(func(a, b int) bool { return a < b })
 	more := compareFun(func(a, b int) bool { return a > b })
 
+	annotations = map[string]struct{}{
+		"ACL": {}, // The container can enforce ACLs.
+	}
+
 	funcImplMap = map[astIdent]funcImpl{
 		"!":                {notImpl, 1, false},
 		"%":                {mod, 2, false},
@@ -42,6 +48,7 @@ func init() {
 		"=":                {eqImpl, 2, false},
 		">":                {more, 2, false},
 		"and":              {andImpl, 1, true},
+		"annotate":         {annotateImpl, 2, true},
 		"append":           {appendImpl, 2, false},
 		"apply":            {applyImpl, 2, false},
 		"bool":             {boolImpl, 1, false},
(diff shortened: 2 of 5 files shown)


stitch/invariant.go
```diff
@@ -80,6 +84,30 @@ func reachImpl(graph Graph, inv invariant) bool {
...
+			toNodes = append(toNodes, node)
+		}
+	}
+
+	allPassed := true
+	for _, from := range fromNodes {
+		for _, to := range toNodes {
+			_, ok := from.Connections[to.Name]
+			pass := ok == inv.target
+			allPassed = allPassed && pass
```
secant: Not sure how many nodes are being processed (and if runtime is even a concern), but you could `return False` if `pass` is `False` and `return True` outside of the for loop.

stitch/invariant_test.go
```diff
@@ -30,6 +30,22 @@ func TestReach(t *testing.T) {
...
+(label "c" (docker "ubuntu"))
+
+(connect 22 "a" "b")
+(connect 22 "b" "c")
+
+(invariant reachDirect false "a" "c")
+(invariant reachDirect true "b" "c")`
+	_, err := initSpec(stc)
+	if err != nil {
+		t.Error(err)
```
secant: Is there any way to verify that `neighborImpl` is returning True/False?

Also what does Impl stand for? 😅

akshayknarayan: This test is analogous to calling `neighborImpl` right? `reachDirect` is the stitch name for `neighborImpl`.

Impl -> "Implementation" :P I borrowed the naming scheme from stitch/func.go

akshayknarayan:
Pushed two new commits which add the ability to annotate nodes in the stitch with static analysis-relevant characteristics, and a new invariant type which does annotation-aware reachability.

The implementation of `reachACL` (`stitch/invariant.go`) is a bit messy, but I think it's not at the threshold of writing some overarching reach() function which reduces code re-occurrence.

stitch/eval.go
```diff
@@ -44,6 +45,7 @@ func (ctx *evalCtx) deepCopy() *evalCtx {
 		containers:  ctx.containers,
 		placements:  ctx.placements,
 		invariants:  ctx.invariants,
+		annotations: ctx.annotations,
```
kklin: Could you make annotations an attribute of label? Cleaner imo.

stitch/func.go
```diff
@@ -745,6 +752,47 @@ func hmapValuesImpl(ctx *evalCtx, args []ast) (ast, error) {
 	return astList(ret), nil
 }
 
+func annotateImpl(ctx *evalCtx, args []ast) (ast, error) {
+	var astArgs []string
```
kklin: So your way of parsing the args is kind of circumventing the eval and type system. e.g., the following wouldn't work:
`(define myAnnotation ACL)
(annotate myAnnotation "webTier")`.

One option is to make the "ACL" a string instead of an ident.

stitch/func.go
```diff
@@ -745,6 +752,47 @@ func hmapValuesImpl(ctx *evalCtx, args []ast) (ast, error) {
...
+	// Annotation format: (annotate ACL <node labels...>)
+	if len(args) < 2 {
+		return nil, fmt.Errorf(
+			"malformed annotation (must specify annotation): %s",
+			argStr,
+		)
+	}
+
+	_, ok := annotations[astArgs[0]]
+	if !ok {
```
secant: Can be rewritten as
```
if _, ok := annotations[astArgs[0]]; !ok {
    ...
}
```

stitch/graph.go
```diff
@@ -173,6 +173,36 @@ func (n Node) dfs() []string {
...
 }
 
+// Find all nodes reachable from the given node.
+// Do not pass through "ACL"-annotated nodes.
+func (n Node) dfsWithACL() []string {
+	reached := map[string]struct{}{}
+	reached[PublicInternetLabel] = struct{}{}
+
+	var explore func(t Node)
+	explore = func(t Node) {
```
secant: Is this better as opposed to `explore := func(t Node)`? If so, is this because of the recursive call to `explore`?

stitch/graph.go
```diff
@@ -173,6 +173,36 @@ func (n Node) dfs() []string {
...
+		for label, node := range t.Connections {
+			if node.Annotation == "ACL" {
+				reached[label] = struct{}{}
+			}
+
+			_, explored := reached[label]
+			if !explored {
```
secant: `if _, explored := reached[label]; !explored {`

stitch/invariant.go
```diff
@@ -69,15 +78,64 @@ func reachImpl(graph Graph, inv invariant) bool {
...
+	var fromNodes []Node
+	var toNodes []Node
+	for _, node := range graph.Nodes {
+		if node.Label == inv.nodes[0] {
+			fromNodes = append(fromNodes, node)
+		}
+		if node.Label == inv.nodes[1] {
+			toNodes = append(toNodes, node)
+		}
+	}
```
secant: Since this block is repeated, could you break this out into a function?

stitch/graph.go
akshayknarayan: the `var explore` part is needed because it's recursive.

stitch/invariant.go
```diff
@@ -109,6 +114,30 @@ func neighborImpl(graph Graph, inv invariant) bool {
...
+
+	for _, from := range fromNodes {
+		for _, to := range toNodes {
+			if pass := contains(from.dfsWithACL(),
```
kklin: More readable to do `if reachable := contains(from.dfsWithACL(), to.Name); reachable != inv.target`
I'll change it.

PR #494, opened 2016-07-28
db/container: Add StitchID field to db.Container

akshayknarayan:
Allow cross-referencing in the minion between containers defined
in the stitch and containers running in quilt


akshayknarayan:
Addressed by https://github.com/NetSys/quilt/commit/106273b4f196b588d48c2e48e6f302801ee08136

PR #495, opened 2016-07-29
Makefile: Fix clean target to delete sub-package coverage data

mmussomele:


Makefile
```diff
@@ -31,8 +31,8 @@ check: format-check
 
 clean:
 	go clean -x $(PACKAGES)
-	rm -f *.cov.coverprofile
-	rm -f *.cov.html
+	rm -f *.cov.coverprofile cluster/*.cov.coverprofile minion/*.cov.coverprofile specs/*.cov.coverprofile
```
kklin: Maybe just using a `find` command would be better? Otherwise these directories will need to be kept up to date.

mmussomele: I would, but `find . -name "*.cov.coverprofile" -delete` ins't supported on all platforms

mmussomele: although I suppose `find . -name "*.cov.coverprofile" -exec rm -f {} \;` should work too

mmussomele:
merged

PR #498, opened 2016-07-30
Add Quiltctl

kklin:
This series divides Quilt into a daemon and a client. The daemon is responsible for enforcing Stitches. The client sends Stitches to the daemon to be enforced. It is also capable of querying the state of the daemon, getting imports, etc.


Makefile
```diff
@@ -61,7 +66,7 @@ format-check:
 lint: format
 	cd -P . && go vet $(PACKAGES)
 	for package in $(PACKAGES) ; do \
-		if [[ $$package != *minion/pb* ]] ; then \
+		if [[ $$package != *minion/pb* && $$package != *api/pb* ]] ; then \
```
mmussomele: Not an issue right now, but as we get more packages that we exclude from golint, we might consider doing something like what we do with `LINE_LENGTH_EXCLUDE` and `COV_SKIP`

api/server/server.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Run accepts incoming `quiltctl` connections and responds to them.
+func Run(conn db.Conn, listenAddr string) error {
+	proto, addr, err := api.ParseListenAddress(listenAddr)
+	if err != nil {
+		return err
+	}
+
+	var sock net.Listener
+	server := server{conn}
```
mmussomele: I don't this should have the same name as the struct type

api/server/server.go
```diff
@@ -0,0 +1,105 @@
...
+	for {
+		sock, err = net.Listen(proto, addr)
+
+		if err != nil {
```
mmussomele:
```
if err == nil {
        break
}

log.WithError(err).Error("Failed to open socket.")
```
?

api/server/server.go
```diff
@@ -0,0 +1,105 @@
...
+
+	s := grpc.NewServer()
+	pb.RegisterAPIServer(s, server)
+	s.Serve(sock)
+
+	return nil
+}
+
+func (s server) QueryMachines(
+	cts context.Context, query *pb.DBQuery) (*pb.MachineReply, error) {
```
mmussomele: I would put a newline after this to visually separate the method header from the body.

api/server/server.go
```diff
@@ -0,0 +1,105 @@
...
+	var machines []db.Machine
+	s.dbConn.Transact(func(view db.Database) error {
+		machines = view.SelectFromMachine(nil)
+		return nil
+	})
+
+	var resp pb.MachineReply
+	for _, m := range machines {
+		resp.Machines = append(resp.Machines, &pb.Machine{
```
mmussomele: Would it be worth having an `api/util` package for converting between pb structs and db structs?

api/server/server.go
```diff
@@ -0,0 +1,105 @@
...
+			PublicIP:  m.PublicIP,
+			PrivateIP: m.PrivateIP,
+		})
+	}
+
+	return &resp, nil
+}
+
+func (s server) QueryContainers(
+	cts context.Context, query *pb.DBQuery) (*pb.ContainerReply, error) {
```
mmussomele: Same here with the newline

minion/main.go
```diff
@@ -2,8 +2,11 @@
 package main
 
 import (
+	"fmt"
 	"time"
 
+	"github.com/NetSys/quilt/api"
+	apiServer "github.com/NetSys/quilt/api/server"
```
mmussomele: Why rename this?

quiltctl/containerCommand.go
```diff
@@ -0,0 +1,37 @@
+package main
```
mmussomele: Is there a reason not to put containerCommand and machineCommand in one file named command?

quiltctl/containerCommand.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/NetSys/quilt/api/client"
+)
+
+type containerCommand struct {
+	host string
+}
+
+func (cCmd *containerCommand) getFlagSet() *flag.FlagSet {
+	flags := flag.NewFlagSet("containers", flag.ExitOnError)
+
```
mmussomele: Why so many blank lines here?

quiltctl/containerCommand.go
```diff
@@ -0,0 +1,37 @@
...
+		daemonConnectError(cCmd.host, err)
+	}
+
+	allContainers, err := c.QueryContainers()
+	if err != nil {
+		daemonResponseError(err)
+	}
+
+	for _, c := range allContainers {
+		fmt.Printf("%v\n", c)
```
mmussomele: Why a print instead of a log?

quiltctl/machineCommand.go
```diff
@@ -0,0 +1,37 @@
+package main
```
mmussomele: Same comments as containerCommand.go

api/server/server.go
kklin: Fixed.

api/server/server.go
kklin: Fixed.

api/server/server.go
kklin: Fixed.

quiltctl/quiltctl.go
```diff
@@ -0,0 +1,71 @@
...
+
+	cmd, err := parseSubcommand(os.Args[1], os.Args[2:])
+	if err != nil {
+		log.WithError(err).Error("Unable to parse subcommand.")
+		usage()
+	}
+
+	cmd.run()
+
+	os.Exit(0)
```
mmussomele: Why not just let main exit?

api/server/server.go
kklin: I don't think it'd be that useful, the conversions only happen once.

api/server/server.go
kklin: Fixed.

minion/main.go
kklin: The minion package contains a type named `server`

quiltctl/containerCommand.go
kklin: Many more commands get added in the next couple patches (`get`, `run`, `stop`, `inspect`, `ssh`)

quiltctl/containerCommand.go
kklin: Personal preference, I guess. First part creates the flagSet, middle sets the flag mappings, final part returns.

quiltctl/quiltctl_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	flags.StringVar(&tCmd.arg, "arg", "", "the test arg")
+
+	return flags
+}
+
+func (tCmd *testCommand) run() {
+	if tCmd.arg != "val" {
+		tCmd.t.Errorf("Bad argument value for testCommand: expected %s, got %s",
+			"val", tCmd.arg)
```
mmussomele: Why format the "val" into the error?

quiltctl/containerCommand.go
kklin: This is the output of the command, while the logs were errors. Plus, this will go to stdout.

quiltctl/error.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"os"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// Utility functions for displaying common errors.
+
+func daemonConnectError(host string, err error) {
```
kklin: Fixed the method name, not sure if the commands should all go in the same file.

cluster/foreman.go
```diff
@@ -109,6 +110,11 @@ func (fm *foreman) runOnce() {
 
 		connected := err == nil
 		if connected && !m.connected {
+			fm.conn.Transact(func(view db.Database) error {
```
mmussomele: What if the machine becomes disconnected later? Should this transact be below, with `m.connected = connected`?

quiltctl/quiltctl_test.go
kklin: Copy and paste, and fast editing :P. Fixed.

quiltctl/quiltctl.go
kklin: Done.

quiltctl/getCommand.go
```diff
@@ -0,0 +1,56 @@
+package main
```
mmussomele: If you put the commands in the same file, bring this one along.

quiltctl/getCommand.go
```diff
@@ -0,0 +1,56 @@
...
+			gCmd.usage()
+		}
+	}
+
+	if err := stitch.GetSpec(spec); err != nil {
+		log.WithError(err).Errorf("Error getting import `%s`.\n", spec)
+		os.Exit(1)
+	}
+
+	fmt.Println("Successfully installed import!")
```
mmussomele: Log?

quiltctl/getCommand.go
```diff
@@ -0,0 +1,56 @@
...
+		if len(nonFlagArgs) > 0 {
+			spec = nonFlagArgs[0]
+		} else {
+			log.Error("No import specified!")
+			gCmd.usage()
+		}
+	}
+
+	if err := stitch.GetSpec(spec); err != nil {
+		log.WithError(err).Errorf("Error getting import `%s`.\n", spec)
```
mmussomele: Why the `\n`?

mmussomele: Why the `\n`?

quiltctl/inspectCommand.go
```diff
@@ -0,0 +1,21 @@
+package main
```
mmussomele: Fold into the single command file.

quiltctl/getCommand.go
kklin: Ah, I was used to fmt.Printf requiring a newline. I do this in a lot of places, will fix.

quiltctl/runCommand.go
```diff
@@ -0,0 +1,98 @@
+package main
```
mmussomele: Fold into command file

quiltctl/runCommand.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/NetSys/quilt/api/client"
+	"github.com/NetSys/quilt/stitch"
+	"github.com/NetSys/quilt/util"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+type runCommand struct {
+	stitch string
+
```
mmussomele: Why this blank line?

quiltctl/runCommand.go
```diff
@@ -0,0 +1,98 @@
...
+		log.WithError(err).Errorf("%s failed to compile.\n", spec)
+		os.Exit(1)
+	}
+
+	err = c.RunStitch(compiled)
+	if err != nil {
+		daemonResponseError(err)
+	}
+
+	fmt.Println("Successfully started run!")
```
mmussomele: Log?

quiltctl/stopCommand.go
```diff
@@ -0,0 +1,62 @@
...
+	c, err := client.New(sCmd.host)
+	if err != nil {
+		daemonConnectError(sCmd.host, err)
+	}
+
+	if err = c.RunStitch(specStr); err != nil {
+		daemonResponseError(err)
+	}
+
+	fmt.Printf("Successfully began stopping namespace `%s`!\n", namespace)
```
mmussomele: same here

quiltctl/getCommand.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (gCmd *getCommand) run() {
+	spec := gCmd.importPath
+	if spec == "" {
+		nonFlagArgs := gCmd.flags.Args()
+		if len(nonFlagArgs) > 0 {
+			spec = nonFlagArgs[0]
+		} else {
+			log.Error("No import specified!")
```
mmussomele: Why so excited? XD

quiltctl/sshCommand.go
```diff
@@ -0,0 +1,95 @@
+package main
```
mmussomele: fold into command file

quiltctl/sshCommand.go
```diff
@@ -0,0 +1,95 @@
...
+			"private key: quiltctl ssh 5 -i ~/.ssh/quilt")
+		sCmd.flags.PrintDefaults()
+	}
+
+	return sCmd.flags
+}
+
+func (sCmd *sshCommand) run() {
+	args := sCmd.flags.Args()
+	tgtMachine, err := strconv.Atoi(args[0])
```
mmussomele: Why not just name it `targetMachine`? Not clear what `tgt` means

quiltctl/sshCommand.go
```diff
@@ -0,0 +1,95 @@
...
+			host = m.PublicIP
+			break
+		}
+	}
+
+	if host == "" {
+		log.Errorf("Unable to find machine `%d`.\n", tgtMachine)
+		log.Error("Available machines:")
+		for _, m := range machines {
+			log.Errorf("%v\n", m)
```
mmussomele: It might be cleaner to join these into a single string and log once.

mmussomele:
Should some of these commits should be folded together? It seems like migrating all the quilt commands into quiltctl should be one patch, adding new commands should be another, etc

kklin:
I'm not sure how I'd break up the commits otherwise. I wanted the stitch patch to be separate because it's half moving to quiltctl, and half restructuring the way we compile and then evaluate. Also, I like having separate commits for new features, makes it easier to revert if need be.

mmussomele:
Also, since quilt is now a daemon, should it's log message be redirected to a file instead of stdout? Like in quilt.go, should we have `log.SetOutput(some file writer)`?

quiltctl/getCommand.go
kklin: Heh, I use exclamation marks in a couple place.. Wonder if I should just change all of them to periods. I think I'd get annoyed after awhile.

quiltctl/getCommand.go
kklin: Some feedback for users is nice

quiltctl/getCommand.go
kklin: Yeah, I ditched them.

kklin:
I think that's a good idea, I'll make a trello card for it. Doesn't have to happen in this series though.

kklin:
Pushed changes, ready for another round.

kklin:
Moved quiltctl commands to their own package.
I made an empty test file for the `command` package so that the build would pass.. I didn't see anything really worth testing, although as the commands increase in complexity I'd expect them to have some tests..

quiltctl/command/container.go
```diff
@@ -0,0 +1,40 @@
...
+	"flag"
+	"fmt"
+
+	"github.com/NetSys/quilt/api"
+	"github.com/NetSys/quilt/api/client"
+)
+
+// Container contains the options for querying containers.
+type Container struct {
+	host string
```
mmussomele: Since this is the only field that Container has, why not simply have `type Container string`?

mmussomele: Or if you want the meaning of the field to remain, `type ContainerHost string`?

quiltctl/command/machine.go
```diff
@@ -0,0 +1,40 @@
...
+	"flag"
+	"fmt"
+
+	"github.com/NetSys/quilt/api"
+	"github.com/NetSys/quilt/api/client"
+)
+
+// Machine contains the options for querying machines.
+type Machine struct {
+	host string
```
mmussomele: Same here. `type MachineHost string`?

quiltctl/quiltctl.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/NetSys/quilt/quiltctl/command"
+
+	log "github.com/Sirupsen/logrus"
+	"google.golang.org/grpc/grpclog"
+)
+
+var commands = map[string]command.SubCommand{
+	"machines":   &command.Machine{},
+	"containers": &command.Container{},
```
mmussomele: If the structs from before become strings, these could be replaced with `new(command.MachineHost)` and `new(command.ContainerHost)`

quiltctl/command/container.go
kklin: The container and machine subcommands will eventually take more flags, e.g.
```
quiltctl machines --query "key=Role,value=Master" --columns "ID,PublicIP"
```
Besides, I don't think we lose much by leaving it as a struct.

quiltctl/quiltctl.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func main() {
+	/* XXX: GRPC spews a lot of uselss log message so we tell to eat its logs.
+	 * Once we have more sophistcated logging support, we should enable the log
+	 * messages when in debug mode. */
+	grpclog.SetLogger(l_mod.New(ioutil.Discard, "", 0))
```
mmussomele: `quilt.go` has support for setting the log level from the command line. Perhaps the level parsing code could be moved to `util/log.go` and both quilt and quiltctl could use it.

quiltctl/command/get.go
```diff
@@ -0,0 +1,59 @@
...
+func (gCmd *Get) Run() {
+	spec := gCmd.importPath
+	if spec == "" {
+		nonFlagArgs := gCmd.flags.Args()
+		if len(nonFlagArgs) > 0 {
+			spec = nonFlagArgs[0]
+		} else {
+			log.Error("No import specified.")
+			gCmd.usage()
+		}
```
mmussomele:
```
if len(nonFlagArgs) == 0 {
        log.Error("No import specified.")
        gCmd.usage()
}
spec = nonFlagArgs[0]
```

mmussomele: `gCmd.usage` exits, so you can reduce indentation here

quiltctl/quiltctl.go
kklin: There's actually something kind of related to this that I think needs to be figured out, curious if you have any ideas.
The way the quiltctl code is setup right now is that each subcommand has its own flagSet; however, there's some flags that are common to multiple commands (e.g. `host`, or `logLevel`)
I think the command format should eventually be like:
`quiltctl -H tcp://8.8.8.8 machines --queryString "Role=Master"`

quiltctl/command/get.go
kklin: Thanks, I'll make the same change for `run` and `stop`.

quiltctl/command/run.go
```diff
@@ -0,0 +1,100 @@
...
+func (rCmd *Run) Run() {
+	spec := rCmd.stitch
+	if spec == "" {
+		nonFlagArgs := rCmd.flags.Args()
+		if len(nonFlagArgs) > 0 {
+			spec = nonFlagArgs[0]
+		} else {
+			log.Error("No spec specified.")
+			rCmd.usage()
+		}
```
mmussomele: Same here with reworking the if-else into just an if

quiltctl/quiltctl.go
mmussomele: logLevel should definitely be a flag of quiltctl, but I'm not sure about H. It doesn't make much sense to say something like:
`quiltctl -H tcp://8.8.8.8 stop test-namespace` but that would be valid if H was a quiltctl flag right?

mmussomele: As we add more commands that require the host flag, we could wrap them further in a QueryCommand interface or something if things were getting repetitive.

ejj: -H is what docker uses fwiw

Ethan (Phone)

kklin:
Pushed changes to reduce if-indentation.

quiltctl/quiltctl.go
```diff
@@ -0,0 +1,63 @@
...
+		usage()
+	}
+
+	cmd.Run()
+}
+
+func parseSubcommand(name string, args []string) (command.SubCommand, error) {
+	cmd, ok := commands[name]
+	if !ok {
+		return nil, fmt.Errorf("unrecongized subcommand: %s", name)
```
sirspinach: unrecognized is misspelled. (haha this is my contribution?)

stitch/stitch.go
```diff
@@ -108,6 +97,38 @@ func New(sc scanner.Scanner, path string, download bool) (Stitch, error) {
 	return spec, nil
 }
 
+// Compile takes a Stitch (possible with imports), and transforms it into a
```
sirspinach: possibly with?

kklin:
Fixed typos pointed out by shwang.

api/client/client.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return Client{}, err
+	}
+
+	pbClient := pb.NewAPIClient(cc)
+	return Client{pbClient: pbClient}, nil
+}
+
+// QueryMachines retrieves the machines tracked by the Quilt daemon.
+func (c Client) QueryMachines() ([]db.Machine, error) {
```
ejj: I think you could unit test these pretty darn easily if you mock out pbClient . . .  Not hugely valuable, but I am trying to get the code coverage up.

api/client/client.go
```diff
@@ -0,0 +1,94 @@
...
+		PrivateIP: pbMachine.PrivateIP,
+	}
+}
+
+// QueryContainers retrieves the containers tracked by the Quilt daemon.
+func (c Client) QueryContainers() ([]db.Container, error) {
+	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
+	reply, err := c.pbClient.QueryContainers(ctx, &pb.DBQuery{})
+	if err != nil {
+		return []db.Container{}, err
```
ejj: I think you can just `return nil, err`

api/common_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"errors"
+	"testing"
+)
+
+func TestAddrParsing(t *testing.T) {
+	t.Parallel()
+
+	checkAddrResult :=
```
ejj: This is kinda ugly, just make it a regular function in the top level?  I think you'd just need to add a `t *testing.T` argument.

kklin:
@mmussomele: Updated the first commit to generalize the DB querying and return responses as JSON, could you review it when you have time?

mmussomele:
Sure, I'll probably get around to it tonight

api/client/client.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/NetSys/quilt/api"
+	"github.com/NetSys/quilt/api/pb"
+	"github.com/NetSys/quilt/db"
+
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+)
+
+// Client provides methods to interact with the Quilt daemon.
+type Client struct {
```
ejj: Can you do `type Client pb.APIClient`?

api/client/client.go
```diff
@@ -0,0 +1,85 @@
...
+		return machines, nil
+	case db.ContainerTable:
+		var containers []db.Container
+		if err := json.Unmarshal(replyBytes, &containers); err != nil {
+			return nil, err
+		}
+		return containers, nil
+	}
+
+	panic(fmt.Sprintf("unsupported table type: %s", table))
```
ejj: I'd slightly prefer this be the default case of the switch statement.  Tad more clear IMO.

api/common.go
```diff
@@ -0,0 +1,24 @@
...
+
+package api
+
+import (
+	"fmt"
+	"strings"
+)
+
+// DefaultSocket is the socket the Quilt daemon listens on by default.
+const DefaultSocket = "unix:///tmp/quilt.sock"
```
ejj: /tmp is a weird place for this, where does docker put it's socket?  OVS puts it in the ovs equivalent of `/var/run/quilt/quilt.sock`

api/server/server.go
```diff
@@ -0,0 +1,83 @@
...
+
+		if err == nil {
+			break
+		}
+		log.WithError(err).Error("Failed to open socket.")
+
+		time.Sleep(30 * time.Second)
+	}
+
+	// Cleanup the socket if we're interrupted.
```
ejj: Huh . . . what happens if we don't do this?  I'm surprised we have to do something so wonky

api/server/server.go
```diff
@@ -0,0 +1,83 @@
...
+	s.Serve(sock)
+
+	return nil
+}
+
+func (s server) Query(cts context.Context, query *pb.DBQuery) (*pb.QueryReply, error) {
+	var rows interface{}
+	switch db.TableType(query.Table) {
+	case db.MachineTable:
+		s.dbConn.Transact(func(view db.Database) error {
```
ejj: Not sure it matters all that much, but it might be *slightly* cleaner if you put the switch inside the transaction, instead of building a transaction within each switch.  The other nice thing is that as we add more tables it will be simpler for those that don't have a SelectFrom implemented on their conn.

ejj:
First and Second patches looks good.

ejj:
Third patch looks good.

ejj:
Fourth Patch looks good.

api/common.go
kklin: You're right, but /var/run is owned by root it looks like.. We could run the daemon as root, but then the client will also have to be run by root to write to the socket. We could circumvent this with some permissions twiddling, but I'm not sure it's worth putting in the effort before we think through what's most usable.

api/server/server.go
kklin: I had it there because the server would fail to create the socket the next time it started up. Doing it this way does give us locking (no more than one instance of quilt/machine) for free, though. I don't feel strongly either way.

api/client/client.go
kklin: No, `pb.APIClient` is an interface.

Makefile
```diff
@@ -20,14 +20,19 @@ install:
 	cd -P . && go install .
 
 check: format-check
-	go test $(PACKAGES)
+	# XXX: Hack to make `make check` skip the quilt-tester test scripts.
```
kklin: @ejj: wasn't sure how to deal with this, so I just threw this here to pass the build.

kklin:
Updated the PR with:
- More coverage in `quiltctl/command`. The `Command` interface changed a bit to facilitate this. Also cut down on the number of random `os.Exit`s.
- Updates to quilt-tester
    - Switched over to using quiltctl
    - Changed the code structure a bit, moving a couple things into structs and improving the error handling.
    - Changed the test suite structure to get rid of the `src` subdirectories
    - Made tests run in a namespace based on the machine's IP.

api/server/server.go
```diff
@@ -0,0 +1,105 @@
...
+	foundTable := true
+	s.dbConn.Transact(func(view db.Database) error {
+		switch db.TableType(query.Table) {
+		case db.MachineTable:
+			rows = view.SelectFromMachine(nil)
+			return nil
+		case db.ContainerTable:
+			rows = view.SelectFromContainer(nil)
+		default:
+			foundTable = false
```
ejj: Why not just return `fmt.Errorf(.....)`?  Then you can ditch this foundTable variable and check check the error that Transact returns.

quilt-tester/config/infrastructure.spec
```diff
@@ -0,0 +1,12 @@
+// We will have three worker machines.
+(define nWorker 3)
+
+(define Namespace "tester-52-53-247-146")
```
ejj: Do you think it's hard to fix this?  I'd really like to before we merge, it's just going to cause problems.

kklin: It's fixed, I can change the spec so that it's clear.

quilt-tester/exec.go
```diff
@@ -0,0 +1,245 @@
...
+// runQuiltDaemon starts the daemon.
+func runQuiltDaemon() {
+	cmd := exec.Command("/quilt", "daemon")
+	execCmd(cmd, "QUILT")
+}
+
+// waitFor waits until `pred` is satisfied, or `timeout` seconds have passed.
+func waitFor(pred func() bool, timeout int) error {
+	timeoutChan := time.After(time.Duration(timeout) * time.Second)
+	for range time.Tick(1 * time.Second) {
```
ejj: Meh, just `for {` and `time.Sleep`  Simpler, doesn't leak the ticker.

kklin: How does that work with the timeout?

quilt-tester/exec.go
```diff
@@ -0,0 +1,245 @@
...
+	go logAndUpdate(*bufio.NewScanner(stdoutPipe), &stdout,
+		l, fmt.Sprintf(logFormatter, "stdout"))
+	go logAndUpdate(*bufio.NewScanner(stderrPipe), &stderr,
+		l, fmt.Sprintf(logFormatter, "stderr"))
+
+	if err := cmd.Start(); err != nil {
+		return "", "", err
+	}
+
+	cmdFinished := make(chan error, 1)
```
ejj: Huh.  I don't quite get the point of this `cmdFinished` thing.  Why can't you just call `cmd.Wait()`?

quilt-tester/exec.go
```diff
@@ -0,0 +1,245 @@
...
+
+	svc := ec2.New(session.New(), &aws.Config{Region: aws.String("us-west-1")})
+	req, _ := svc.TerminateInstancesRequest(toDelete)
+	return req.Send()
+}
+
+func cleanupMachines(namespace string) {
+	l := log.testerLogger
+
+	l.infoln("Cleaning up first with `quilt stop`.")
```
ejj: Same question, why the oddness with the logging.

quilt-tester/exec.go
```diff
@@ -0,0 +1,245 @@
...
+		l.println(fmt.Sprintf(logFmt, logStr))
+	}
+}
+
+// execCmd executes the given command, and returns the stdout and stderr output.
+// `logLineTitle` is the prefix for logging to the container log.
+func execCmd(cmd *exec.Cmd, logLineTitle string) (string, string, error) {
+	l := log.cmdLogger
+
+	l.infoln(fmt.Sprintf("%s: Starting command: %v", logLineTitle, cmd.Args))
```
kklin: We have the `fileLogger`s to save logs to the files on the website. Are you saying we should use logrus loggers instead of rolling our own?

quilt-tester/exec.go
kklin: Totally can...

quilt-tester/exec.go
ejj: You could still use the time.After chan and check on each loop.  I'm just saying the time.Tick isn't really providing any value.

quilt-tester/exec.go
ejj: Yeah, if it's not too hard.  If it's a big project, then don't bother.

kklin: It'd take more effort than I'd like to dedicate to the tester at this point.

kklin:
- Addressed previous round comments
- Added a commit to close the grpc connection (sigh)
- Fixed a bug where the quilt-tester pass/fail counts didn't update
- Added some unit tests for quilt-tester

PR #499, opened 2016-08-02
minion/etcd: Lower log level when failing to create the minion dir

mmussomele:
The tester was failing because we were logging a warning when etcd
failed to create the minion dir a few times at boot. This is
expected, so it should only log an info level message.


mmussomele:
merged

PR #501, opened 2016-08-05
Monitoring

akshayknarayan:
Basic monitoring functionality.

### DIFF
--- a/minion/docker/docker.go
+++ b/minion/docker/docker.go
@@ -61,6 +61,7 @@ type client interface {
 	StartContainer(id string, hostConfig *dkc.HostConfig) error
 	CreateExec(opts dkc.CreateExecOptions) (*dkc.Exec, error)
 	StartExec(id string, opts dkc.StartExecOptions) error
+	InspectExec(id string) (*dkc.ExecInspect, error)
 	UploadToContainer(id string, opts dkc.UploadToContainerOptions) error
 	DownloadFromContainer(id string, opts dkc.DownloadFromContainerOptions) error
 	RemoveContainer(opts dkc.RemoveContainerOptions) error
@@ -115,16 +116,16 @@ func (dk Client) Run(opts RunOptions) (string, error) {
 
 // Exec executes a command within the container with the supplied name.
 func (dk Client) Exec(name string, cmd ...string) error {
-	_, err := dk.ExecVerbose(name, cmd...)
+	_, _, err := dk.ExecVerbose(name, cmd...)
 	return err
 }
 
 // ExecVerbose executes a command within the container with the supplied name.  It also
 // returns the output of that command.
-func (dk Client) ExecVerbose(name string, cmd ...string) ([]byte, error) {
+func (dk Client) ExecVerbose(name string, cmd ...string) ([]byte, int, error) {
 	id, err := dk.getID(name)
 	if err != nil {
-		return nil, err
+		return nil, -1, err
 	}
 
 	var inBuff, outBuff bytes.Buffer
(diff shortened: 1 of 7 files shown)


minion/monitor/monitor.go
```diff
@@ -0,0 +1,179 @@
...
+		log.WithError(err).Warn("container not found")
+		return false
+	}
+	out, err := dk.ExecVerbose(dCont.Name[1:], "ping", "-q", "-c3", ip)
+	if err != nil {
+		log.WithError(err).Warn(string(out))
+		return false
+	}
+
+	lines := strings.Split(string(out), "\n")
```
kklin: Could we just use the exit code of ping?

minion/monitor/monitor.go
```diff
@@ -0,0 +1,180 @@
...
+
+type connection struct {
+	from db.Container
+	to   db.Container
+}
+
+// Monitor checks via ping that each connection outlined in the stitch is live.
+func Monitor(conn db.Conn, dckr docker.Client) {
+	tick := time.Tick(15 * time.Second)
+	for {
```
kklin: Would be cleaner if you `range`d over the tick

minion/docker/docker.go
```diff
@@ -153,26 +154,32 @@ func (dk Client) ExecVerbose(name string, cmd ...string) ([]byte, error) {
...
 	for scanner.Scan() {
 		outBuff.WriteString(scanner.Text() + "\n")
 	}
 	if err := scanner.Err(); err != nil {
-		return nil, err
+		return nil, -1, err
 	}
 
-	return outBuff.Bytes(), nil
+	return outBuff.Bytes(), exitCode, nil
```
secant: (Nitpicky) You could return `execInspect.ExitCode` and not declare `exitCode`

minion/monitor/monitor.go
```diff
@@ -0,0 +1,170 @@
...
+)
+
+type connection struct {
+	from db.Container
+	to   db.Container
+}
+
+// Monitor checks via ping that each connection outlined in the stitch is live.
+func Monitor(conn db.Conn, dckr docker.Client) {
+	for tick := time.Tick(15 * time.Second); true; <-tick {
```
secant: Can be written as `for tick := range time.Tick(15 * time.Second)`

minion/monitor/monitor.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		log.WithError(err).Warn("[monitor] Invalid spec.")
+		return
+	}
+
+	var conns []connection
+	for _, e := range edges {
+		fr, err := nodeToContainer(e.From, containers)
+		if err != nil {
+			log.WithError(err).Warn("[monitor] Could not find container.")
```
secant: Maybe specify which container could not be found?

minion/monitor/monitor.go
```diff
@@ -0,0 +1,170 @@
...
+		log.Warn(fmt.Sprintf("[monitor] %s -> %s unreachable!",
+			fail.from.IP, fail.to.IP))
+	}
+}
+
+func getEdges(conn db.Conn) ([]stitch.Edge, error) {
+	var minion db.Minion
+	tick := time.Tick(30 * time.Second)
+	for {
+		<-tick
```
secant: `time.Sleep(30 * time.Second)` instead?

ejj:
Let's shelve this for now.  We didn't use it in the paper and I don't see an immediate application for our early target users.  Perhaps we should put it on a branch?  Not sure.

PR #502, opened 2016-08-05
Minion network

mmussomele:
Change ports to have the same name as their IP and rework network.go:runMaster to use a hashjoin

### DIFF
--- a/minion/network/network.go
+++ b/minion/network/network.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/NetSys/quilt/db"
+	"github.com/NetSys/quilt/join"
 	"github.com/NetSys/quilt/minion/docker"
 	"github.com/NetSys/quilt/minion/ovsdb"
 
@@ -25,6 +26,15 @@ const ovnBridge = "br-int"
 const GatewayIP = "10.0.0.1"
 
 const gatewayMAC = "02:00:0a:00:00:01"
+
+type dbport struct {
+	bridge string
+	ip     string
+	mac    string
+}
+
+// dbslice is a wrapper around []dbport to allow us to perform a join
+type dbslice []dbport
 
 // Run blocks implementing the network services.
 func Run(conn db.Conn, dk docker.Client) {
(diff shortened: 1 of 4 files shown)


minion/network/network.go
```diff
@@ -63,77 +79,78 @@ func runMaster(conn db.Conn) {
 		})
 
 		connections = view.SelectFromConnection(nil)
+
+		for _, l := range labels {
```
ejj: Why do this within the transaction?

minion/network/network.go
```diff
@@ -63,77 +79,78 @@ func runMaster(conn db.Conn) {
...
+
 		return nil
 	})
 
 	if !leader {
 		return
 	}
 
-	ovsdb, err := ovsdb.Open()
+	openvsdb, err := ovsdb.Open()
```
ejj: This change isn't appropriate for this commit.  Could do it in a later one if you'd like, though I think it was fine as it was.

minion/network/network.go
```diff
@@ -63,77 +79,78 @@ func runMaster(conn db.Conn) {
...
-		garbageMap[lport.Name] = struct{}{}
+	portKey := func(val interface{}) interface{} {
+		return val.(ovsdb.LPort).Name
 	}
 
-	for _, dbl := range labels {
-		if !dbl.MultiHost {
-			continue
+	dbKey := func(val interface{}) interface{} {
+		switch val := val.(type) {
```
ejj: The join should also include the bridge for future proofing.

ejj:
Merged the first patch, some high level comments on the second patch (also some in slack).  Someone else should review as well.

minion/network/network.go
```diff
@@ -64,79 +79,78 @@ func runMaster(conn db.Conn) {
...
-			}).Warn("Failed to create port.")
+	for _, dbp := range dbps {
+		switch dbp := dbp.(type) {
+		case db.Label:
+			if !dbp.MultiHost {
+				continue
+			}
+			makePort(lSwitch, dbp.IP, labelMac, dbp.IP)
+		case db.Container:
+			makePort("quilt", dbp.IP, dbp.Mac, dbp.IP)
```
sirspinach: I think this should be `lSwitch` again, rather than `"quilt"`.

minion/network/network.go
```diff
@@ -64,79 +79,78 @@ func runMaster(conn db.Conn) {
...
+		default:
+			panic("Not Reached")
 		}
+	}
 
-		if _, ok := garbageMap[dbl.Label]; ok {
-			delete(garbageMap, dbl.Label)
-			continue
-		}
+	_, ovsps, dbps := join.HashJoin(ovsdb.LPortSlice(lports), dbData, portKey, dbKey)
```
sirspinach: I think ovsPorts is easier to read than ovsps. Your call here.

PR #504, opened 2016-08-05
RFC: New Scheduler

ejj:
New scheduler is here!  Still some cleanup to do, haven't read the diff yet, commit messages need to be cleaned up, etc.  Posting now in case people want to review it before I merge.


db/container.go
```diff
@@ -16,6 +16,7 @@ type Container struct {
 	Pid      int
 	IP       string
 	Mac      string
+	Minion   string
```
ZhongtianWang: Since `Minion` actually refers to minion's private IP, I think it would be nice to change the name to make it less confusing.

minion/etcd/minion.go
```diff
@@ -0,0 +1,121 @@
...
+	conn.Transact(func(view db.Database) error {
+		dbms, sms := filterSelf(view.SelectFromMinion(nil), storeMinions)
+		del, add := diffMinion(dbms, sms)
+
+		for _, m := range del {
+			view.Remove(m)
+		}
+
+		for _, m := range add {
+			minion := view.InsertMinion()
```
kklin: Could this be
```
m.id = minion.ID
view.Commit(m)
```

We do this in a lot of places, wonder if it'd be worth making `InsertMinion` take minions to insert.

ejj:
Merged

PR #505, opened 2016-08-06
azure: Properly return errors when setting up the Azure provider

kklin:
We weren't properly returning an error when attempting to load the Azure
credentials. This caused `cluster` to think that the Azure provider was usable,
when it really wasn't.


ejj:
sigh . . . this is why we need unit tests  . . .

PR #506, opened 2016-08-07
Godep: Fix conflicting grpc on godep restore

sirspinach:
The grpc upgrade patch successfully compiled, but a created a
Godep conflict that prevented `godep restore` from working
properly.


ejj:
Should this patch update the vendor directory as well, or just Godep.json?

sirspinach:
Godep.json only. I think I incorrectly chose an old commit hash rather than the new when I was rebasing.

PR #507, opened 2016-08-08
stitch: Check `provider` and `size` argument types

kklin:
The implementations for `provider` and `size` casted their arguments to
strings without checking if they were really strings. This would cause
panics when bad arguments were passed in. Now, we nicely return an error
instead.


ejj:
Merged

PR #509, opened 2016-08-11
etcd: Remove oldMaster variable from elector logic

sirspinach:
This variable, and the if block that uses it, have no effect.


ejj:
Merged

PR #518, opened 2016-08-23
work in progress: quilt driver

sirspinach:
We don't want containers to begin running before they are connected to a working network.

We can solve this problem by creating a custom docker network driver, and associate this network driver with the docker 'quilt' network.

When the scheduler `Run`s a new container and attaches it to the `quilt` network, docker calls the CreateEndpoint and AttachNetwork methods from our driver. In this pull request's implementation, CreateEndpoint simply blocks until `view.MinionSelf().NetworkReady` is true. Since docker will not start the new container until CreateEndpoint is finished, the container is guaranteed to wait for NetworkReady to be true.


sirspinach:
CreateEndpoint currently leads to a deadlock because it `Transact`s for `view.MinionSelf().NetworkReady`, while at the same time the scheduler calls `dk.Run(container)` from within another `Transact`.

sirspinach:
Next Steps:

- Driver probably shouldn't be its own package. It can go with scheduler.
- CreateEndpoint could do more than just be a dummy waiting function. We could, for example, create the endpoints in the CreateEndpoint method. According to Ethan, `NetworkReady` should be defined as: 1. container veths are created and attached at both ends and 2. `/etc/hostnames` is defined. Right now `NetworkReady` is defined here: https://github.com/NetSys/quilt/pull/518/commits/2f118ce8587286726e8e4fac1220ff403317829f#diff-58ffc76eab31b80b47f7c0f82cd18684R169

ejj:
@mmussomele is working on a complete version. Closing for now

PR #519, opened 2016-08-26
Add scale tester

mmussomele:
Still have 1 thing to do (support for booting an N+1th container), but the code is ready for review, and those changes should be minimal.


quilt-tester/util/util.go
```diff
@@ -0,0 +1,26 @@
...
+	"errors"
+	"time"
+)
+
+// Sleep is stored in a variable so it can be mocked out in the unit tests
+var Sleep = time.Sleep
+
+// WaitFor simply waits until `pred` returns true. If it times out before `pred` returns
+// true, it will return an error.
+func WaitFor(pred func() bool, timeout time.Duration) error {
```
kklin: Looks like some of these changes are from `exec` pr.. Maybe you should just include my `quilt-tester: Speed up `waitFor` unit tests` commit in your PR?

quilt-tester/scale/container.go
```diff
@@ -0,0 +1,104 @@
...
+	"minion":         {},
+}
+
+type scaleContainer struct {
+	ip    string
+	image string
+	name  string
+}
+
+func getContainers(minionList []string) []scaleContainer {
```
kklin: This is pretty :)

quilt-tester/scale/container.go
```diff
@@ -0,0 +1,104 @@
...
+
+			container.ip = host
+			out <- container
+		}
+	}()
+
+	return out
+}
+
+func mergeContainers(channels ...chan scaleContainer) chan scaleContainer {
```
kklin: Why not just `[]chan`?

quilt-tester/scale/util.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+func ssh(hostIP string, command ...string) *exec.Cmd {
```
kklin: Hm, my exec PR also introduces a similar function. Could you just define this in the `quiltctl/command` package?

quilt-tester/scale/scale.go
```diff
@@ -0,0 +1,221 @@
...
+		usage()
+	}
+
+	if outputFile == "" {
+		log.Error("No output file specified")
+		usage()
+	}
+
+	log.Info("Starting the quilt daemon.")
+	cmd := exec.Command("quilt", "-log-file=/tmp/quilt-scale-output", "daemon")
```
kklin: Kind funky place to put the log file. Maybe take the output as an arg, and output to /dev/null by default?

quilt-tester/scale/scale.go
```diff
@@ -0,0 +1,221 @@
...
+	if err := cmd.Start(); err != nil {
+		log.WithError(err).Fatal("Failed to start quilt")
+	}
+	time.Sleep(5 * time.Second) // Give quilt a while to start up
+
+	log.Info("Getting local client")
+	localClient, err := client.New(api.DefaultSocket)
+	if err != nil {
+		cmd.Process.Kill()
```
kklin: Less error prone to just `defer` the `Kill` and `cleanup` functions once.

quilt-tester/scale/scale.go
```diff
@@ -0,0 +1,221 @@
...
+	log.Info("Machines have booted successfully")
+
+	log.Info("Loading main stitch")
+	flatSpec, err := loadSpec(specPath, namespace)
+	if err != nil {
+		cleanup(cmd, namespace, localClient)
+		log.WithError(err).WithField("spec", specPath).Fatal("Failed to load")
+	}
+
+	// Gather information the expected state of the system from the main stitch
```
kklin: "Gather information on.." (?)

quilt-tester/scale/scale.go
```diff
@@ -0,0 +1,221 @@
...
+	log.Info("Querying stitch containers")
+	expectedContainers, err := queryStitch(flatSpec)
```
kklin: I think `queryStitchContainers` would be a better name.

quilt-tester/scale/scale.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		cleanup(cmd, namespace, localClient)
+		log.WithError(err).WithField("spec", flatSpec).Fatal("Failed to get " +
+			"Stitch from spec")
```
kklin: Maybe "Failed to get expected containers from spec"?

quilt-tester/scale/scale.go
```diff
@@ -0,0 +1,221 @@
...
+	startTimestamp := time.Now().In(utc)
+
+	workerIPs, err := getWorkerIPs(localClient)
+	if err != nil {
+		cleanup(cmd, namespace, localClient)
+		log.WithError(err).Fatal("Failed to get worker IPs")
+	}
+
+	log.Info("Waiting for containers to boot")
+	err = testUtil.WaitFor(containersBooted(workerIPs, expCounts), time.Hour)
```
kklin: Would be nice if this timeout were a const somewhere, I wouldn't be surprise if we tweaked it.

quilt-tester/scale/scale.go
```diff
@@ -0,0 +1,221 @@
...
+		return
+	}
+
+	log.Info("Waiting for machines to shutdown")
+	time.Sleep(time.Minute)
+}
+
+func usage() {
+	flag.Usage()
+	fmt.Println("\nPreboot spec, main spec and output file are required flags.")
```
kklin: I think usage messages are normally edited by modifying the `flag.Usage` function itself.

quilt-tester/scale/sleep/preboot-sleep.spec
```diff
@@ -0,0 +1,7 @@
+(define AdminACL (list "local"))
+
+(let ((cfg (list (provider "Amazon")
+         (size "m4.4xlarge")
+		 (githubKey "mmussomele"))))
```
kklin: Use the SSH keypair as `quilt-tester`? This way different users don't have to edit this spec.

kklin: Now that I think about it, this wouldn't work for me at all because my GitHub key is at `~/.ssh/github` and the scale tester only uses the default private key.

quilt-tester/scale/timestamp.go
```diff
@@ -0,0 +1,111 @@
...
+	"strings"
+	"sync"
+	"time"
+)
+
+func getLastTimestamp(workers []string) time.Time {
+	// Everything is after the zero time
+	latestTimestamp := time.Time{}
+	containers := getContainers(workers)
+	containerCount := float64(len(containers))
```
kklin: More clear to do the type conversion when you take the average.

quilt-tester/scale/timestamp.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	channels := []chan time.Time{}
+	for ip, containers := range hostMap {
+		channels = append(channels, queryTimestamps(ip, containers))
+	}
+
+	out := mergeTimestamps(channels...)
+
+	timestampsCollected := 0.0
```
kklin: Does this need to be a float?

quilt-tester/scale/timestamp.go
```diff
@@ -0,0 +1,111 @@
...
+			for _, container := range containers {
+				if _, ok := collected[container]; ok {
+					continue
+				}
+
+				cmdStr := fmt.Sprintf(cmdTemplate, container)
+				args := strings.Fields(cmdStr)
+				output, err := ssh(minion, args...).Output()
+				if err != nil {
+					continue
```
kklin: log?

quilt-tester/scale/timestamp.go
```diff
@@ -0,0 +1,111 @@
...
+				out <- time.Unix(seconds, 0)
+				collected[container] = struct{}{}
+			}
+		}
+	}()
+
+	return out
+}
+
+func mergeTimestamps(channels ...chan time.Time) chan time.Time {
```
kklin: Why not just `[]chan`?

quilt-tester/scale/wait.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+
+	"github.com/NetSys/quilt/api/client"
+	"github.com/NetSys/quilt/db"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+func containersBooted(minionList []string, expCounts map[string]int) func() bool {
+	expectedTotal := 0.0
```
kklin: Same here re types

quilt-tester/scale/wait.go
```diff
@@ -0,0 +1,66 @@
...
+			if count, ok := counts[image]; !ok || count != expCount {
+				return false
+			}
+		}
+
+		return true
+	}
+}
+
+func machinesBooted(localClient client.Client) func() bool {
```
kklin: Isn't this shared with `quilt-tester`?

quilt-tester/scale/sleep/Dockerfile
```diff
@@ -0,0 +1,4 @@
+From ubuntu
+RUN apt-get update && apt-get -y install iputils-ping
```
kklin: Trailing whitespace

quilt-tester/scale/sleep/scale-sleep.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh
+
+timestamp() {
+    until ping -q -c1 localhost > /dev/null 2>&1; do
+        sleep 0.5
+    done
+    echo "quilt_timestamp=$(date -u +%s)"
+}
+
+timestamp
```
kklin: Trailing whitespace

api/client/client.go
```diff
@@ -16,7 +16,7 @@ import (
 
 const (
 	// The timeout for connecting to the daemon.
-	timeout = 5 * time.Second
+	timeout = time.Minute
```
kklin: This also increases the timeout for when you're first making a connection to the daemon. So if the daemon isn't even running, users wouldn't get feedback for an entire minute.
Should just have multiple timeout constants, I guess.

kklin:
Code LGTM, just some small nits. I'll try running it tomorrow.

quilt-tester/scale/scale.go
mmussomele: I did this originally, but log.Fatal calls os.Exit, so deferred functions aren't run.

kklin: Hm, I guess you could just `log.Error` and exit in the defer function?

quilt-tester/util/util.go
mmussomele: I think @ejj is going to merge your PR first, so this will resolve then

quilt-tester/scale/wait.go
mmussomele: Ah, I suppose it is. I'll move this into util then

mmussomele:
Fixes made to everything with a +1, rest to come

quilt-tester/tester_test.go
```diff
@@ -43,19 +47,19 @@ func TestWaitFor(t *testing.T) {
 		}
 		return false
 	}
-	err := waitFor(callThreeTimes, 5)
+	err := util.WaitFor(callThreeTimes, 5*time.Second)
 	if err != nil {
```
secant: `if err := ...; err != nil`

api/server/server.go
```diff
@@ -55,6 +56,21 @@ func Run(conn db.Conn, listenAddr string) error {
...
+	go func() {
+		for spec := range apiServer.stitchChan {
+			compiledSpec, err := stitch.New(spec)
+			if err != nil {
+				log.WithError(err).WithField("spec",
+					spec).Error("Failed to compile")
+			}
+
+			err = engine.UpdatePolicy(apiServer.dbConn, compiledSpec)
+			if err != nil {
```
secant: `if err := ...; err != nil`

quilt-tester/scale/scale.go
```diff
@@ -0,0 +1,272 @@
...
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// Reserved namespace for the scale tester. If you are developing the scale tester,
+// use the `-namespace=<your_namespace>` command line flag to ensure you don't interfere
+// with the tester.
+// We want the namespace to be deterministic so the user can use `quilt stop` to halt the
+// namespace if the scale tester fails for some reason.
+const scaleNamespace = "bd89e4c89f4d384e7afb155a3af99d8a6f4f5a06a9fecf0b6d220eb66e156dbe"
```
secant: maybe have some indication that this is the scale tester?

api/server/server.go
mmussomele: It's more than 89 characters

quilt-tester/scale/sleep/preboot-sleep.spec
mmussomele: The sleep spec is just an example spec, the scale tester is general enough to work with any valid stitch. If you were running scale tests, you would just change the specs to work for your keys.
This is a little confusing though, so I'll update the specs to be more clear that they're to be changed.

mmussomele:
This is obsolete, resubmitted on paper branch

PR #525, opened 2016-09-11
Change stitch extension to .js

secant:
so importing doesn't break - importing `.spec` will now break though


kklin:
This can be closed, I merged it in with my geo PR.

PR #527, opened 2016-09-13
Scale test

mmussomele:
This is all the changes I've made while developing the scale tester so far. A lot of these commits should probably be squashed after the deadline before we merge into master and some will probably be deleted.


mmussomele:
Not ready for merge, I need to clean this up first

PR #529, opened 2016-09-27
fix vagrant ssh command concatenation / multi-line string

tejasmanohar:
currently, this code looks like a mix of concatenation w/ string literals in a multi-line string which doesn't work. had to fix to get vagrant provider to run.


ejj:
Wow does it actually work?  We've been having problems with vagrant for months, but we've been so busy with other issues that we haven't gotten around to fixing it yet...

I think the only thing that needs to be changed is the commit message.  I'm somewhat particular about their format (see the contributing documentation, or other commits in the history).  Go ahead and fix that up and I'll merge.

tejasmanohar:
@ejj After making the PR'd changes to the Vagrant image, I was able to `quilt daemon` && `quilt run sparkPI.js` (changed provider in code... as well as diskSize so I still have some free space on my system :P) without getting any errors about the provider going unavailable, failing to boot, etc.

That said, I haven't actually figured out where to see the output of the Spark job yet so I'm not 100% sure that the Vagrant provider works though I'm fairly sure that this change does since there's a clear before/after (will look into integration tests). I expect to see results in STDOUT of spark-submit command but didn't find out where that was being run last night (or which machine was master from Quilt logs). I'll continue looking tonight, but let me know if you have any tips!

ejj:
Have a look at this:  it should tell you how to find the results (assuming it's working).

https://github.com/NetSys/quilt/tree/master/specs/spark

tejasmanohar:
@ejj Ahh, will try that later tonight! Thanks.

PR #531, opened 2016-09-28
Assorted Fixes

kklin:
These are the small tweaks from the paper branch, many of which are dependencies of future PR's.

PR #532, opened 2016-09-28
Lint: Fixed lint

alvinwan:
- Whenever AWS fails to connect, add all the details of the error to the log.
- Fixed lint errors of the form "ineffectual assignment to..."

depends on #531


cluster/provider/awsSpot.go
```diff
@@ -63,7 +64,7 @@ func (clst *amazonCluster) Connect(namespace string) error {
 	clst.namespace = namespace
 
 	if _, err := clst.List(); err != nil {
-		return errors.New("AWS failed to connect")
+		return fmt.Errorf("AWS failed to connect: %v", err)
```
mmussomele: This change should be its own commit.

cluster/provider/awsSpot.go
```diff
@@ -513,7 +514,7 @@ func (clst *amazonCluster) SetACLs(acls []string) error {
 
 		if !groupIngressExists {
 			log.Debug("Amazon: Add intragroup ACL")
-			_, err = session.AuthorizeSecurityGroupIngress(
+			_, _ = session.AuthorizeSecurityGroupIngress(
```
mmussomele: why not just `session.AuthorizeSecurityGroupIngress(`?

ejj:
I think this patch is obsolete now. If you think it should still be merged, go ahead and re-open it.

PR #536, opened 2016-09-28
vagrant: log stderr when vagrant shell command fails

tejasmanohar:
With logs, we can (try to) figure out why the Vagrant provider is "unavailable", "unable to boot", etc. without adding one-off debugging code and/or running commands outside of Quilt.


cluster/provider/vagrant_api.go
```diff
@@ -166,8 +167,17 @@ func (api vagrantAPI) Shell(id string, commands string) ([]byte, error) {
 	}
 	chdir = fmt.Sprintf(chdir, vdir+id)
 	shellCommand := chdir + strings.Replace(commands, "%s", id, -1) + ")"
-	output, err := exec.Command(shCmd, []string{"-c", shellCommand}...).Output()
-	return output, err
+
+	var outbuf, errbuf bytes.Buffer
```
kklin: Hm, I think it might be reasonable for this command to return both `stdout` and `stderr`, and to let the caller log if they want to.

tejasmanohar: 👍  039c086

tejasmanohar:
>Only other nit is that that letter after the colon should be capitalized:
vagrant: Log..

Fixed. 2c23be6

cluster/provider/vagrant_api.go
kklin: Sorry, what I meant was that I don't think the `Shell` function should log, it should just give sufficient information to the caller for the caller to log. A specific place where this is better is https://github.com/NetSys/quilt/blob/master/cluster/provider/vagrant.go#L17. The way that we tell whether the Vagrant provider is usable for booting machines is by trying to add the box. If it fails, it's not necessarily an error (so we shouldn't log at Error):  the user may just be using the Amazon provider and not care that they don't have Vagrant installed.

Not sure if this is the best way, but maybe instead of returning `stderr` explicitly, we should return `stdout, errors.New(stderr)`..

tejasmanohar: Fair. I suggest we explicitly return `stdout, stderr, err` rather than `stdout, err(... stderr...)` since `err` can give different information like status code. Will fix.

That said, if you'd prefer to embed it in the err (let me know), I'd suggest implementing the `Err() string` interface on `type ShellError struct{}` but also providing `Stderr() []bytes` and Stdout() []bytes` functions (callee can assert interface and call functions if passes).

kklin: Both sound fine to me, up to you. I just wanted to move that log :)

ejj:
Fwiw I like the idea of returning stdout, street, and an error code

Ethan (Phone)

tejasmanohar:
agreed. will re-open against master.

PR #541, opened 2016-09-29
Update code checkout command

akashgangil:
The repo is no longer private, a non-ssh git clone (https) would work.

### DIFF
--- a/docs/Dev.md
+++ b/docs/Dev.md
@@ -84,10 +84,9 @@ We currently require go version 1.3 or later.  Ubuntu 15.10 uses this version
 by default, so you should just be able to apt-get install golang to get
 started.
 
-Since the repository is private, you can't use "go get" to checkout the source
-code, so you'll have to do so manually:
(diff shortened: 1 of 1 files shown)


ejj:
Akash! Thanks so much for the contribution!  The code itself looks great to me.

One tiny nit . . . the title should begin with a short one word subject heading.  I.E. `docs: Update code checkout command`

I'd be happy to make this change myself before merging if you'd prefer.  Or feel free to resubmit the PR.

Thanks again Akash

ejj:
Ah one more thing.  I actually do think that the recommend way to get the code is actually to use `go get` as suggested in the original documentation.  Cloning it manually should be more of a secondary option for advanced users.

The code that's there is somewhat unclear on this point though. I think perhaps the best thing to do is merge your patch, and then I'll do another patch on top of it that clarifies things a bit.

ejj:
Made a slight tweak to the commit message and merged.

PR #542, opened 2016-10-02
quiltctl: Add Logs command

tejasmanohar:
Depends on [kklin/exec](https://github.com/kklin/quilt/tree/exec).

This is being pushed up for initial review but isn't 100% ready for merge, as I haven't been able to test it locally yet due to problems retrieving StitchIDs locally (though I tested the SSH commands independently and wrote unit tests for parsing etc so fairly confident).
```
$ quilt containers
ERROR [Oct  2 00:22:16.299] Error connecting to leader.              error=no leader found
```
I dived in some, and it seems  `client.QueryMachines()` is returning neither results (`[]`) nor errors [here](https://github.com/kklin/quilt/blob/b94c71c3375df4d5425595ae4f2afb294c7f4e16/quiltctl/command/command.go#L42). Other commands like `quilt stop <namespace>` work so I imagine it's something beyond the connection layer. I'll investigate further (likely tomorrow), but let me know if you have any ideas in the meantime (had a spec running on AWS while testing FYI). Thanks!

// cc @kklin


quiltctl/command/log.go
```diff
@@ -0,0 +1,130 @@
...
+func (lCmd *Log) createFlagSet() {
+	flags := flag.NewFlagSet("logs", flag.ExitOnError)
+
+	flags.StringVar(&lCmd.host, "H", api.DefaultSocket,
+		"the host to query for machine information")
+	flags.StringVar(&lCmd.privateKey, "i", "",
+		"the private key to use to connect to the host")
+	flags.StringVar(&lCmd.sinceTimestamp, "since", "", "show logs since timestamp")
+	flags.BoolVar(&lCmd.shouldTail, "f", false, "follow log output")
+	flags.BoolVar(&lCmd.showTimestamps, "t", false, "show timestamps")
```
tejasmanohar: needs tests

quiltctl/command/log.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (lCmd *Log) createFlagSet() {
+	flags := flag.NewFlagSet("logs", flag.ExitOnError)
+
+	flags.StringVar(&lCmd.host, "H", api.DefaultSocket,
+		"the host to query for machine information")
+	flags.StringVar(&lCmd.privateKey, "i", "",
+		"the private key to use to connect to the host")
+	flags.StringVar(&lCmd.sinceTimestamp, "since", "", "show logs since timestamp")
+	flags.BoolVar(&lCmd.shouldTail, "f", false, "follow log output")
```
tejasmanohar: needs tests

quiltctl/command/log.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (lCmd *Log) createFlagSet() {
+	flags := flag.NewFlagSet("logs", flag.ExitOnError)
+
+	flags.StringVar(&lCmd.host, "H", api.DefaultSocket,
+		"the host to query for machine information")
+	flags.StringVar(&lCmd.privateKey, "i", "",
+		"the private key to use to connect to the host")
+	flags.StringVar(&lCmd.sinceTimestamp, "since", "", "show logs since timestamp")
```
tejasmanohar: needs tests

quiltctl/command/log.go
```diff
@@ -0,0 +1,130 @@
...
+		lCmd.targetContainer)
+	if err != nil {
+		log.WithError(err).
+			Error("Error getting the host on which the container is running.")
+		return 1
+	}
+
+	containerClient, err := getClient(api.RemoteAddress(containerHost))
+	if err != nil {
+		log.WithError(err).Error("Error connecting to leader.")
```
kklin: This should be something like "Error connecting to container client" (I know this error is from the exec command, I'll fix it there myself).

quiltctl/command/log.go
```diff
@@ -0,0 +1,130 @@
...
+		log.WithError(err).Error("Error running the logs command.")
+		return 1
+	}
+
+	return 0
+}
+
+// Usage prints command usage info.
+func (lCmd *Log) Usage() {
+	fmt.Println("usage: quilt logs [-H=<daemon_host>] [-i=<private_key>] " +
```
kklin: The main usage string in `quilt.go` should also be updated.

quiltctl/command/log_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	logsCmd.Run()
+
+	if !sshCalled {
+		t.Errorf("Never tried SSHing")
+	}
+}
+
+func checkLogParsing(t *testing.T, args []string, expContainer int, expKey string,
```
kklin: I we should just make the command parsing checks generic:
`func checkParsing(t *testing.T, args []string, tgt SubCommand, exp interface{})`

Would probably also be easier to define another helper for `checkParsingError`.

kklin:
Oh, it's kind of scary that we print an error message if you ctrl-C out of the `quilt logs -f. Not sure what the best way to fix this is though.

tejasmanohar:
>Oh, it's kind of scary that we print an error message if you ctrl-C out of the `quilt logs -f. Not sure what the best way to fix this is though.

I'll look into this once I get my local setup together. If the error is due to not tearing down properly, we can (and should since it's long-running) handle the signal.

tejasmanohar:
>BTW, I was able to use the quilt logs command fine. Are you using a custom minion? Do quilt machines and quilt containers work?

@kklin `quilt machines` works ->
```
Machine-2{Master, Amazon us-west-1 m4.large, sir-iwpid22j, PublicIP=54.153.58.245, PrivateIP=172.31.1.40, Disk=32GB, Connected}
Machine-6{Worker, Amazon us-west-1 m4.large, sir-1kpid7sh, PublicIP=54.67.36.216, PrivateIP=172.31.2.10, Disk=32GB, Connected}
Machine-5{Worker, Amazon us-west-1 m4.large, sir-2sbiegpj, PublicIP=52.53.157.179, PrivateIP=172.31.7.126, Disk=32GB, Connected}
Machine-4{Worker, Amazon us-west-1 m4.large, sir-4yfgdmth, PublicIP=52.53.158.89, PrivateIP=172.31.3.13, Disk=32GB, Connected}
Machine-3{Worker, Amazon us-west-1 m4.large, sir-gp68d6cj, PublicIP=54.153.14.112, PrivateIP=172.31.13.131, Disk=32GB, Connected}
```

I haven't messed with `minion`s yet. Here's the error I'm getting in `getLeaderClient` (via `getLeaderIP`)--
```
ERROR [Oct  2 16:19:26.018] Unable to get leader IP.                 error=rpc error: code = 2 desc = unrecognized table: db.Etcd
```

I SSH'd into the Master VM and checked on etcd via `docker ps` -> `docker logs` but don't see anything troubling there. Searching the code for more info about what's supposed to be here, but let me know if you know what's missing in the meantime!

tejasmanohar:
>FYI, I fixed the build in my exec branch.

What was the problem with build/tests, btw (didn't see any new commits, guess you rebased them together?)? Were they timing out? I saw `exit status 1` when coverage was running without any details in the CI output. Wish you could SSH into Travis container like you can with Circle :\

kklin:
@tejasmanohar Not sure how much you've dived into the architecture yet, but, in a nutshell, Quilt has two parts, and the `exec` series changed the remote part (the minion). See https://github.com/NetSys/quilt/blob/master/docs/Dev.md#developing-the-minion for instructions on running development code for the minion.

The build was failing in Travis because `golint` had some failures -- it was kind of buried in the Travis output. I rebased the commit that fixes the failures to be the first in my series.

tejasmanohar:
@kklin Ahh, I'll read through rest of `docs/` tonight. I bet the problem is that I'm not using your changes to `minion/` locally.

tejasmanohar:
That was it! Docs are useful :P. I'm up & running now. I'll incorporate feedback here & ping back.

tejasmanohar:
>Oh, it's kind of scary that we print an error message if you ctrl-C out of the `quilt logs -f. Not sure what the best way to fix this is though.

@kklin I've handled this via putting the subprocess in a new group and handling the signal, but it's not perfect / there are definitely remaining side-effects (e.g. SSH client may not be properly closed). As I noted in the commit, I think we should refactor this to use a Go SSH client rather than the `ssh` program directly. What do you think? If you agree, I'd be happy to do so.

kklin:
@tejasmanohar Good idea, I think that'd be much better!

kklin:
`quilt logs` (and `quilt exec`) just hangs for me:
```
kklin@Kevins-MBP-6 ~quilt % quilt logs 1                                               [remotes/tejas/logs_cmd][0]
^Ctcsetattr: Interrupted system call
Killed by signal 1.
```
Works fine if I set `Setpgid: false`.

quiltctl/command/log_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	logsCmd.Run()
+
+	if !sshCalled {
+		t.Errorf("Never tried SSHing")
+	}
+}
+
+func checkLogParsing(t *testing.T, args []string, exp *Log, expErr error) {
```
kklin: why *Log?

quiltctl/command/log_test.go
```diff
@@ -0,0 +1,142 @@
...
+			}
+			return
+		}
+
+		t.Errorf("Unexpected error when parsing exec args: %s", err.Error())
+		return
+	}
+
+	if logsCmd.targetContainer != exp.targetContainer {
+		t.Errorf("Expected exec command to parse target container %d, but got %d",
```
kklin: "Expected log command ...", same for the other couple errors.

local.mk
```diff
@@ -0,0 +1 @@
+REPO=tejasmanohar
```
kklin: Don't think you meant to commit this :)

quiltctl/command/ssh.go
```diff
@@ -118,8 +119,13 @@ var ssh = func(host string, args []string) *exec.Cmd {
 		"-o", "StrictHostKeyChecking=no"}
 
 	cmd := exec.Command("ssh", append(baseArgs, args...)...)
+
+	// run command in own process group
```
kklin: Comments should generally be a sentence (start with a capital, and end with a period? Something like that).

quiltctl/command/log.go
```diff
@@ -105,6 +107,16 @@ func (lCmd *Log) Run() int {
 	}
 	if lCmd.shouldTail {
 		dockerCmd += " --follow"
+
+		// handle interrupt signal since this is a long-running process
```
kklin: Seems nicer to handle this in the `ssh` function:
- This same problem is present if you have a long running `exec`
- `Exit`ing directly here prevents any potential cleanup work

Perhaps if we catch the signal in `ssh`, we just return `nil`. What do you think?

quiltctl/command/log_test.go
tejasmanohar: mostly out of habit. though it doesn't matter for a small struct like this, why would you prefer passing by value instead of reference? (totally fine changing it if you'd like)

quiltctl/command/log.go
tejasmanohar: That means the `ssh` function would call `Run()` as well, which could be fine-- just something to consider. That said, I don't see why this is necessary in the first place so maybe it isn't and I missed something-- will check this evening. I believe Go will naturally exit w/ `130` if an interrupt signal is sent so, as long as the signal does not propagate to the child process, we should not need to do anything special here.

quiltctl/command/ssh.go
```diff
@@ -113,6 +114,25 @@ func (sCmd *SSH) Usage() {
 	sCmd.flags.Usage()
 }
 
+func runCommand(cmd *exec.Cmd) error {
+	cancel := make(chan os.Signal, 1)
+	signal.Notify(cancel, os.Interrupt)
+
+	exit := make(chan error, 1)
+	go func() {
+		exit <- cmd.Run()
```
tejasmanohar: When `exit <- cmd.Run()` is executed inside the goroutine (`go func(){}()`) (like it is now), `cmd.Stdout`, `cmd.Stderr`, etc. doesn't seem to pipe to `os.Stdout`, `os.Stderr`, etc. (assigned in `ssh()`) and the command exits immediately (`Run()` returns `nil`) though when it's executed outside the goroutine--
```go
exit := make(chan error, 1)
exit <- cmd.Run()

var err error
select {
// ...
```
it works fine (except the sub-command blocks so the interrupt signal is not handled of course). Any ideas what may be causing this? Pretty stumped.

tejasmanohar: Handled in 8d063d6fa18023be972a52009abbbe086ea62797 w/o modifying `ssh()`

quiltctl/command/log.go
tejasmanohar: Handled in 8d063d6fa18023be972a52009abbbe086ea62797 w/o making `ssh()` run

tejasmanohar:
Incorporated comments, @kklin. Should be ready for review!

quiltctl/command/log_test.go
kklin: I have a slight preference for by value because seeing the pointer forced me to think a little harder.. I guess there were more possible errors with `*exp`, so I had to read the code more closely.

ejj: I have a strong preference for passing things by value.  I generally like the immutability it provides.

tejasmanohar:
Did some debugging with @kklin, and it turned out `logs` command was working on my computer and not the rest of yours (@ejj, @kklin) due to the `ControlPath` setting in my `~/.ssh/config`. Not sure I entirely understand _why_ that is yet, but I decided it'd be much simpler if we migrated to Go's SSH library since there are likely a number of settings that could conflict.

ejj:
FWIW, I think the migration patch should be first in the series, and then the logs patch should build on top of that.  That way we don't have a patch in the history where logs is implemented, but broken for most users.

tejasmanohar:
>That way we don't have a patch in the history where logs is implemented, but broken for most users.

@ejj The `logs` command only failed when being run in a separate process group where we tried to handle the interrupt signals. I removed that from the first commit of this PR so that it works but ^C etc is not handled properly (alike `quilt exec` on master).

ejj:
Ah, kk

quiltctl/command/log.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	containerClient, err := getClient(api.RemoteAddress(containerHost))
+	if err != nil {
+		log.WithError(err).Error("Error connecting to container client.")
+		return 1
+	}
+	defer containerClient.Close()
+
+	container, err := getContainer(containerClient, lCmd.targetContainer)
```
ejj: btw, could some of these above three calls be pulled into the library and shared with exec?

quiltctl/command/log_test.go
ejj: Going to change this?

quiltctl/command/ssh.go
```diff
@@ -118,8 +118,10 @@ var ssh = func(host string, args []string) *exec.Cmd {
 		"-o", "StrictHostKeyChecking=no"}
 
 	cmd := exec.Command("ssh", append(baseArgs, args...)...)
+
```
ejj: Shouldn't be in this patch.

quiltctl/command/log.go
tejasmanohar: i agree that our `quiltctl/ssh` package could take in container ID and host and figure out how to connect but i'd like to do that in another PR as it requires making an interface around getClient etc for testability.

tejasmanohar: could also make a separate helper function to just return containerHost & dockerID but they're pretty unrelated so doesn't seem like a good abstraction

tejasmanohar:
Should be ready for merge

PR #543, opened 2016-10-02
quiltctl: Add exec command

kklin:


quiltctl/command/ssh.go
```diff
@@ -118,3 +111,15 @@ func (sCmd *SSH) Run() int {
 func (sCmd *SSH) Usage() {
 	sCmd.flags.Usage()
 }
+
+// Stored in a variable so we can mock it out for unit tests.
+var ssh = func(host string, args []string) *exec.Cmd {
```
tejasmanohar: This feels leaky. I think we're better off encapsulating with structs or interfaces.

kklin: @tejasmanohar Yeah, I agree. I think using the SSH library will fix this, though.

PR #547, opened 2016-10-09
Minion subnet

mmussomele:
This patch changes IP assignment so that each minion gets it's own subnet and assigns IPs locally.

Unit tested and works on sparkPI.spec


minion/etcd/network.go
```diff
@@ -125,6 +142,50 @@ func runNetwork(conn db.Conn, store Store) {
...
+		time.Sleep(5 * time.Second)
+	}
+}
+
+func generateSubnet(store Store) net.IP {
+	// XXX: Replace this with a less random selection process. i.e. select a random
+	// element from a set of untaken subnets
+	for {
+		subnet := randomMinionSubnet()
+		err := store.Create(subnetStore+strconv.Itoa(int(subnet)), "", 0)
```
kklin: Could we have `randomMinionSubnet` return a `net.IP`, and just store the String representation in etcd?

minion/etcd/network.go
```diff
@@ -7,6 +7,7 @@ import (
 	"math/rand"
```
ejj: Commit title reads a bit weird to me.  Also I think it's too long.  Perhaps just "Make workers assign container IPs"

minion/etcd/network.go
```diff
@@ -22,11 +23,25 @@ const (
 	minionDir      = "/minion"
 	labelToIPStore = minionDir + "/labelIP"
 	containerStore = minionDir + "/container"
+	subnetStore    = minionDir + "/subnets/"
+
+	privateMaskBits  = 8
+	privateSubnetInt = 0x0a000000
```
ejj: I don't much like the names of these variables.

I think it might be clearer if you had `quiltPrefix`, `quiltMask`, `subPrefix`, `subMask` and possibly `labelPrefix`.

Also I'd use the `net` utilities to express these rather than writing out the hex.  Also I think I'd story the prefix as a mask instead of as `20` or whatever.

minion/etcd/network.go
```diff
@@ -76,6 +91,7 @@ func wakeChan(conn db.Conn, store Store) chan struct{} {
 }
 
 func runNetwork(conn db.Conn, store Store) {
+	subnet := generateSubnet(store)
```
ejj: Should the masters be generating a subnet for themselves?

minion/etcd/network.go
ejj: +1

ejj: I also think that doing a simple Create isn't sufficient here.  We really need this subnet key to disappear from the etcd in the event the minion dies (so we don't end up leaking subnets).

mmussomele:
Changes made and tested.

minion/etcd/etcd.go
```diff
@@ -20,6 +20,7 @@ type Store interface {
 	Create(path, value string, ttl time.Duration) error
 	Update(path, value string, ttl time.Duration) error
 	Set(path, value string, ttl time.Duration) error
+	Refresh(path, value string, ttl time.Duration) error
```
ejj: Eh . . . I think you actually uncovered a rather serious bug in the etcd elector code. `minion/etcd/elector` uses the Update() function to refresh the leader election key which is technically wrong.  If, for whatever reason, some other node has taken control over the key, we shouldn't overwrite it, but instead fail to update as your Refresh function here does.

Since that's such a serious bug, I think we should call it out with it's own commit.  Specifically, I think the addition of the Refresh function, and the associated fix to minion/etcd/elector.go should be in it's own commit before the rest of the series.  Should be easy, just moving code around.

ejj: Bonus points if you can write a unit test that catches the problem btw =)

minion/etcd/minion.go
```diff
@@ -11,12 +15,30 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
-const timeout = 30
+const (
+	minionTimeout = 30
+	subnetStore   = minionDir + "/subnets"
+	subnetTTL     = 30 * time.Minute
```
ejj: That's a bit excessive . . .  Maybe 5 minutes?

minion/etcd/minion.go
```diff
@@ -11,12 +15,30 @@ import (
...
+)
+
+var (
+	// Store in a variable so we can change it for unit tests
+	subnetAttempts = 1000
+
+	quiltPrefix = net.IPv4(10, 0, 0, 0)
```
ejj: Minor nit, but it'd be nice to have a comment above this saying "10.0.0.0/8".  The CIDRMask function is a bit odd looking so having the comment would clarify a lot.

minion/etcd/minion.go
```diff
@@ -26,7 +48,7 @@ func runMinionSync(conn db.Conn, store Store) {
 func readMinion(conn db.Conn, store Store) {
 	tree, err := store.GetTree("/minion/nodes")
 	if err != nil {
-		log.WithError(err).Warning("Failed to get minions form Etcd.")
+		log.WithError(err).Warning("Failed to get minions from Etcd.")
```
ejj: Separate patch.

minion/etcd/minion.go
```diff
@@ -119,7 +141,90 @@ func writeMinion(conn db.Conn, store Store) {
...
 
 	key := "/minion/nodes/" + minion.PrivateIP
-	if err := store.Set(key, string(js), timeout*time.Second); err != nil {
+	if err := store.Set(key, string(js), minionTimeout*time.Second); err != nil {
 		log.Warning("Failed to update minion node in Etcd: %s", err)
 	}
 }
+
+func syncSubnet(conn db.Conn, store Store) {
+	minion, err := conn.MinionSelf()
```
ejj: Random unrelated side note.  I wonder if MinionSelf() should panic if there is no self.  Would save a lot of useless error checking for something that can only occur if we have a bug anyways.

minion/etcd/minion.go
```diff
@@ -119,7 +141,90 @@ func writeMinion(conn db.Conn, store Store) {
...
+		time.Sleep(time.Minute)
+	}
+
+	conn.Transact(func(view db.Database) error {
+		view.Commit(minion)
+		return nil
+	})
+}
+
+func generateSubnet(store Store, minion db.Minion) (string, error) {
```
ejj: Minor nit, but the file would read a bit more easily if this function came before the sync function.  Logically this code makes more sense to read first.

minion/etcd/minion.go
```diff
@@ -119,7 +141,90 @@ func writeMinion(conn db.Conn, store Store) {
...
+	// XXX: Replace this with a less random selection process. i.e. select a random
```
ejj: Meh.  I think it's fine.

Though here's an idea for an algorithm that's simple if you really care.
```
cur = randomSubnet()
For subnetAttempts:
    check(cur)
    if failed:
         cur = (cur + 1) % range
```

ejj: To be clear, I think you should either remove the comment or implement the new algorithm.  I don't care which.

minion/etcd/minion.go
```diff
@@ -119,7 +141,90 @@ func writeMinion(conn db.Conn, store Store) {
...
+		subnet := randomMinionSubnet()
+		err := store.Create(subnetKey(subnet), minion.PrivateIP, subnetTTL)
+		if err == nil {
+			return subnet, nil
+		}
+		log.WithError(err).WithField("subnet",
+			subnet).Debug("Subnet taken, trying again.")
+	}
+
+	return "", errors.New("failed to create subnet")
```
ejj: I like the word `allocate` instead of `create` here.

minion/etcd/minion.go
```diff
@@ -119,7 +141,90 @@ func writeMinion(conn db.Conn, store Store) {
...
+}
+
+func randomMinionSubnet() string {
+	minionBits, _ := subMask.Size()
+	quiltBits, _ := quiltMask.Size()
+	subnetBits := minionBits - quiltBits
+	maxIP := uint32(0x7fffffff >> uint(31-subnetBits))
```
ejj: I think all of this would be simpler if you just choose a random 32bi `x` and then did `randomSubnet = x & (~mask) | prefix`

minion/etcd/minion_test.go
```diff
@@ -160,6 +161,35 @@ func TestFilter(t *testing.T) {
 	}
 }
 
+// Test that GenerateSubnet generates valid subnets, and passively test that it can
+// generate enough unique subnets by generating 3500 out of the possible 4095
+func TestGenerateSubnet(t *testing.T) {
```
ejj: Test for sync subnet?

ejj: Basically all the new code in this file should have coverage IMO.

minion/etcd/network.go
```diff
@@ -24,9 +24,13 @@ const (
...
 
-// We store rand.Uint32() in a variable so it's easily mocked out by the unit tests.
-// Nondeterminism is hard to test.
-var rand32 = rand.Uint32
+var (
+	// We store rand.Uint32() in a variable so it's easily mocked out by the unit
+	// tests. Nondeterminism is hard to test.
+	rand32 = rand.Uint32
+
+	labelPrefix = net.IPv4(10, 0, 0, 0) // Labels get their own /20
```
ejj: Can we put all of the prefixes in some central constants package?  I'm worried about it getting out of sync.

minion/etcd/network.go
```diff
@@ -173,48 +199,44 @@ func updateEtcdContainer(s Store, etcdData storeData, containers []db.Container)
...
-	score := func(left, right interface{}) int {
-		return containerJoinScore(left.(storeContainer), right.(storeContainer))
-	}
-	pairs, _, _ := join.Join(dbContainers, etcdData.containers, score)
+	dbContainerSlice := dbSliceToStoreSlice(containers)
 
-	newIPMap := map[string]string{}
-	for _, c := range dbContainers {
-		newIPMap[string(c.StitchID)] = ""
+	// Even if we aren't assigning IPs, we still need to sync up with the IPs
```
ejj: Meh, I don't like this comment.  A random reader of the code in the future won't get the "Even if we aren't assigning IPs reference".

minion/etcd/network.go
```diff
@@ -332,6 +356,66 @@ func updateWorkerDBC(view db.Database, self db.Minion, etcdData storeData) {
 
 		view.Commit(dbc)
 	}
+
+	if self.Subnet == "" {
+		return etcdData, nil
+	}
+
+	oldContainers := view.SelectFromContainer(nil)
```
ejj: I don't really like that each worker is attempting to update the etcd container's file.  (1) it leads to this rather awkward merging code which is somewhat error-prone and (2) having all the nodes hit this one file is going to cause either failed writes, or workers over-writing each others updates.

I think it would be much better if we had a directory which contained a key per worker that had a json map from stitch ID to Ip for those containers running on that worker.  Then each worker, when populating it's local database, would read all the keys in that directory and do a join to assign the Ips locally.

minion/etcd/network.go
```diff
@@ -411,63 +495,70 @@ func containerJoinScore(left, right storeContainer) int {
 // syncIPs takes a map of IDs to IPs and creates an IP address for every entry that's
 // missing one.
 func syncIPs(ipMap map[string]string, prefixIP net.IP) {
```
ejj: Is there any way we could somehow pull this Ip management code out into a library that is used by bot the containers and labels?

minion/constants/constants.go
```diff
@@ -0,0 +1,24 @@
+package constants
```
ejj: Should this be in the previous, or it's own patch?

stitch/stitch.go
```diff
@@ -91,6 +93,11 @@ func toStitch(parsed []ast) (Stitch, error) {
 		return Stitch{}, err
 	}
 
+	if tooManyWorkers(*ctx.machines) {
```
ejj: @kklin, what do you think of this?  I'm not sure this should be in stitch as it's very QRI specific.  I might be inclined to drop the patch entirely.  Specifying workers is in some ways like specifying too many containers, we'll make a best effort, but may fail.

ejj: Sure that feels much better to me.

minion/etcd/minion.go
mmussomele: Minions won't have a self until they receive their initial config, so they'll be running first.

minion/constants/constants.go
mmussomele: There's no reason for them to be exported until this patch

minion/etcd/network.go
mmussomele: It seems to make sense to put them with the IP management library for the minion.

mmussomele:
Changed made and code tested. Confirmed to work on sparkPI.

mmussomele:
Note: There's a performance issue with adding TTL support to etcd_mock, I'm working on diagnosing it.

minion/etcd/minion.go
mmussomele: Didn't quite do this, but I found a way that I think is simpler

mmussomele:
Ready for review.

minion/etcd/network.go
```diff
@@ -166,48 +185,55 @@ func updateEtcdContainer(s Store, etcdData storeData, containers []db.Container)
...
 
-	ip.Sync(newIPMap, net.IPv4(10, 0, 0, 0))
-	for i := range dbContainers {
-		dbContainers[i].IP = newIPMap[string(dbContainers[i].StitchID)]
+	dbContainerSlice := dbSliceToStoreSlice(containers)
+	for i := range dbContainerSlice {
+		id := dbContainerSlice[i].StitchID
+		dbContainerSlice[i].IP = ipMap[strconv.Itoa(id)]
 	}
+	sort.Sort(storeContainerSlice(dbContainerSlice))
```
ejj: I'd consider putting the sort in the writeDiff function.  That way it's a little bit more clear when you read the writeDiff why the string comparison is supposed to work

minion/etcd/network.go
```diff
@@ -325,6 +353,78 @@ func updateWorkerDBC(view db.Database, self db.Minion, etcdData storeData) {
 
 		view.Commit(dbc)
 	}
+
```
ejj: I'm sure this code works, but I think it's a bit difficult to reason about the fact that `etcdc` has an IP field given that the worker should be completely ignoring it in this code.  I think in this patch, we should remove the IP from etcdc just for clarity -- plus it removes any possibility of etcdc and the dbc from disagreeing on the IP for a particular container, and that causing confusion.

minion/etcd/network.go
```diff
@@ -325,6 +353,78 @@ func updateWorkerDBC(view db.Database, self db.Minion, etcdData storeData) {
...
+	subnet := ip.Parse(self.Subnet, ip.QuiltPrefix, ip.QuiltMask)
+
+	oldContainers := view.SelectFromContainer(nil)
+	newIPMap, err := updateContainerIP(etcdData, oldContainers, subnet, self, store)
+	if err != nil {
+		return
+	}
+
+	for _, c := range oldContainers {
+		if ipAddr, ok := newIPMap[strconv.Itoa(c.StitchID)]; ok && ipAddr != "" {
```
ejj: Why would this ever happen?  Isn't the source of truth on the IP the database code? I.E. shouldn't a particular container's IP be fixed by the code that's doing the assignment?

minion/etcd/network.go
```diff
@@ -383,6 +483,19 @@ func updateDBLabels(view db.Database, etcdData storeData) {
...
+	}
+
+	for k, v := range first {
+		if val, ok := second[k]; !ok || v != val {
+			return false
+		}
+	}
+	return true
+}
+
```
ejj: Btw, I think containerJoinScore should change (possibly in an earlier patch).

Basically, I think we should add a patch before this one that says a given containers Ip address should never change. Thus the join score shouldn't take into account the IP address, (or rather the score should be -1 if the IP is different).

minion/etcd/network.go
ejj: Ah, I guess it would happen because we're assigning IPs in updateContainer IP?  So this code would go away later?  I think I'd add a comment to that affect at least.

minion/etcd/network.go
```diff
@@ -325,6 +353,78 @@ func updateWorkerDBC(view db.Database, self db.Minion, etcdData storeData) {
...
+		return oldIPMap, nil
+	}
+
+	jsonData, err := json.Marshal(newIPMap)
+	if err != nil {
+		log.WithError(err).Error("Failed to marshal minion container IP map")
+		return oldIPMap, err
+	}
+
+	err = store.Set(path.Join(minionIPStore, self.PrivateIP), string(jsonData), 0)
```
ejj: What if the minion dies . . . do we leak this key or does it get cleaned up later?

I wonder if the minion key in minion/etcd/minion.go should be a directory instead?

Sigh, an etcd3 lease would simplify a lot of this.  Too bad we didn't get the update done.

minion/etcd/network.go
```diff
@@ -383,6 +483,19 @@ func updateDBLabels(view db.Database, etcdData storeData) {
 	}
 }
 
+func strMapEquals(first, second map[string]string) bool {
```
ejj: util has an implementation of this.

minion/ip/ip.go
```diff
@@ -15,56 +15,93 @@ var (
...
+func FromInt(ip32 uint32) net.IP {
+	b := make([]byte, 4)
+	binary.BigEndian.PutUint32(b, ip32)
+	return net.IP(b)
+}
+
+// Random selects a random IP address in the given prefix and mask that isn't already
+// present in conflicts.
+// Returns 0 on failure.
+func Random(conflicts map[string]struct{}, pre net.IP, subnetMask net.IPMask) net.IP {
```
ejj: Probably belongs in a separate patch, but I think since we're working with a smaller range here, we should ditch this Random allocation logic.  My preference would be to search the entire subnet for an IP linearly starting from a random location.

minion/etcd/network.go
mmussomele: Actually this is the only place it's used, I think I'm gonna ditch it

ejj: Don't you need the sort for the diff to work?

minion/etcd/network.go
mmussomele: So storeContainers still need the IP field since it makes the master logic simpler. I'll just have the worker ignore it and overwrite it if the ipMap updateContainerIPs returns has a conflicting value.

minion/etcd/network.go
mmussomele: so this code changed slightly because of the above comment, I think this is clear:
```
for _, c := range oldContainers {
                 ipAddr, ok := newIPMap[strconv.Itoa(c.StitchID)]
                 if ok && ipAddr != c.IP {
                         c.IP = ipAddr
                         c.Mac = ip.ToMac(ipAddr)
                         view.Commit(c)
                 }
         }
```

minion/ip/ip.go
mmussomele: hmmm, problem is the Master updates the storeContainers and they'll all have "" for an IP until it syncs up, so all the container will just be destroyed if we do that (none will join up)

minion/etcd/network.go
mmussomele: hmmm, problem is the Master updates the storeContainers and they'll all have "" for an IP until it syncs up, so all the container will just be destroyed if we do that (none will join up)

mmussomele: ah you know what, I'll only return -1 if they disagree and the etcd container actually has an IP address

mmussomele:
Changes made, unit tested, and sparkPI tested.

mmussomele:
@ejj ready for another review

minion/etcd/network.go
ejj: Yeah I don't quite understand this point still.  Perhaps we should talk about it on slack or in person.  If it really does make the master simpler I think it's fine, but I'd like to understand it a bit better.

minion/etcd/minion.go
```diff
@@ -118,8 +155,106 @@ func writeMinion(conn db.Conn, store Store) {
...
+	err := store.Refresh(subnetKey(minion.Subnet), minion.PrivateIP,
+		subnetTTL)
+	if err == nil {
+		return minion
+	}
+	log.WithError(err).Infof("Failed to refresh subnet '%s', "+
+		"generating a new one.", minion.Subnet)
+
+	// Invalidate the subnet until we get a new one.
+	conn.Transact(func(view db.Database) error {
```
ejj: This logic repeats just below.  I think you could pull it out into some sort of setMinionSubnet function.

minion/etcd/minion.go
```diff
@@ -118,8 +155,106 @@ func writeMinion(conn db.Conn, store Store) {
...
+	})
+
+	// If we failed to refresh, someone took our subnet or we never had one.
+	for {
+		minion.Subnet, err = generateSubnet(store, minion)
+		if err == nil {
+			break
+		}
+
+		log.WithError(err).Error("Failed to allocate subnet, " +
```
ejj: Warn.

minion/etcd/minion.go
```diff
@@ -2,29 +2,65 @@ package etcd
...
+	minionTimeout = 30
+	subnetStore   = minionDir + "/subnets"
+	selfNode      = "self"
+)
+
+var (
+	// Store in variables so we can change them for unit tests
+	subnetAttempts = 1000
+	syncWait       = 15 * time.Second
+	ttlWait        = time.Minute
```
ejj: I'm not sure we really need ttlWait at all.

minion/etcd/minion.go
```diff
@@ -118,8 +155,106 @@ func writeMinion(conn db.Conn, store Store) {
...
+	// If we failed to refresh, someone took our subnet or we never had one.
+	for {
+		minion.Subnet, err = generateSubnet(store, minion)
+		if err == nil {
+			break
+		}
+
+		log.WithError(err).Error("Failed to allocate subnet, " +
+			"trying again in a minute")
+		sleep(ttlWait)
```
ejj: Not clear why we need to sleep this long.  If it failed to to a race condition, I think we should just wait a second or two and be fine.

minion/etcd/minion.go
```diff
@@ -2,29 +2,65 @@ package etcd
...
+const (
+	minionTimeout = 30
+	subnetStore   = minionDir + "/subnets"
+	selfNode      = "self"
+)
+
+var (
+	// Store in variables so we can change them for unit tests
+	subnetAttempts = 1000
+	syncWait       = 15 * time.Second
```
ejj: I also generally like these wait timers to be expressed as a fraction of the TTL.  I.e. synWait should be subnetTTL / 4 or something.

PR #548, opened 2016-10-12
stitch: Get github sshkeys with get request instead of github api

mmzhao:
This patch makes a get request to github for sshkeys instead of using their api. This should solve the github api key rate error problem as you can curl their site for the keys seemingly without limit.


secant:
Make sure you run `gofmt`on your edits (or `make check`)
The build failed because your import statements are not alphabetically ordered

mmzhao:
Thanks for the help. I just realized that as well.

PR #549, opened 2016-10-12
stitch: Get github sshkeys with get request instead of github api

mmzhao:
This patch should solve the github api key rate error problem as you can curl their site for the keys seemingly without limit.

### DIFF
--- a/stitch/key.go
+++ b/stitch/key.go
@@ -1,6 +1,10 @@
 package stitch
 
-import "github.com/google/go-github/github"
+import (
+	"io/ioutil"
+	"net/http"
+	"strings"
+)
 
 type key interface {
 	keys() ([]string, error)
@@ -9,13 +13,14 @@ type key interface {
 }
 
 var githubCache = make(map[string][]string)
+var httpGet = http.Get
 
 func (githubKey astGithubKey) keys() ([]string, error) {
 	username := string(githubKey)
 	if keys, ok := githubCache[username]; ok {
 		return keys, nil
 	}
-	keys, err := getGithubKeys(username)
+	keys, err := getGithubKeys("https://github.com/" + username + ".keys")
 	if err != nil {
 		return nil, err
 	}
@@ -27,20 +32,17 @@ func (plaintextKey astPlaintextKey) keys() ([]string, error) {
 	return []string{string(plaintextKey)}, nil
 }
 
-// Stored in a variable so we can mock it out for the unit tests.
-var getGithubKeys = func(username string) ([]string, error) {
-	usersService := github.NewClient(nil).Users
-	opt := &github.ListOptions{}
-	keys, _, err := usersService.ListKeys(username, opt)
-
+func getGithubKeys(keyURL string) ([]string, error) {
+	res, err := httpGet(keyURL)
 	if err != nil {
 		return nil, err
 	}
-
-	var keyStrings []string
-	for _, key := range keys {
-		keyStrings = append(keyStrings, *key.Key)
+	keyBytes, err := ioutil.ReadAll(res.Body)
+	defer res.Body.Close()
+	if err != nil {
+		return nil, err
 	}
-
+	keys := strings.TrimSpace(string(keyBytes))
+	keyStrings := strings.Split(keys, "\n")
 	return keyStrings, nil
 }
--- a/stitch/stitch_test.go
+++ b/stitch/stitch_test.go
@@ -1,7 +1,10 @@
 package stitch
 
 import (
+	"bytes"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"reflect"
 	"regexp"
 	"strings"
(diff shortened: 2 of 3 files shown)


stitch/key.go
```diff
@@ -29,18 +33,16 @@ func (plaintextKey astPlaintextKey) keys() ([]string, error) {
...
 // Stored in a variable so we can mock it out for the unit tests.
 var getGithubKeys = func(username string) ([]string, error) {
-	usersService := github.NewClient(nil).Users
-	opt := &github.ListOptions{}
-	keys, _, err := usersService.ListKeys(username, opt)
-
+	res, err := http.Get("https://github.com/" + username + ".keys")
 	if err != nil {
 		return nil, err
 	}
```
mmussomele: slightly less error prone to `defer res.Body.Close()` here

mmzhao:
Changes made.

stitch/key.go
```diff
@@ -1,6 +1,13 @@
...
 
-import "github.com/google/go-github/github"
+import (
+	"io/ioutil"
+	"net/http"
+	"path"
+	"strings"
+)
+
+var githubURL = "https://github.com/"
```
mmussomele: `const`?

stitch/key.go
```diff
@@ -1,6 +1,13 @@
 package stitch
```
mmussomele: Not sure how important this is, but the detailed part of the commit message should describe the context of this commit. i.e. it could help to briefly describe what the github api key rate error problem was, and what you did to fix it (quick 1-2 sentence overview)

mmussomele: Sorry to keep coming back, but a few more things about the commit message. 
1) It needs to be wrapped to 70 characters (reviewing [Contributing.md](https://github.com/NetSys/quilt/blob/master/Contributing.md#coding-style) might help)
2) This is really picky, but something seems off about the commit title. I think it tries to say too much. Maybe something along the lines of `stitch: Fetch Github keys using HTTP GET request` and letting the rest of the message explain the 'instead' part (just a suggestion)

stitch/key.go
```diff
@@ -29,18 +36,16 @@ func (plaintextKey astPlaintextKey) keys() ([]string, error) {
 
 // Stored in a variable so we can mock it out for the unit tests.
 var getGithubKeys = func(username string) ([]string, error) {
-	usersService := github.NewClient(nil).Users
-	opt := &github.ListOptions{}
-	keys, _, err := usersService.ListKeys(username, opt)
-
+	res, err := http.Get(path.Join(githubURL, username+".keys"))
```
mmussomele: Actually turns out using path.Join doesn't work, it will clean the https://github.com into https:/github.com, my bad

mmussomele: Doing what you did before is probably fine, but I'd still keep the github constant

mmzhao:
Changes made. Thanks for all the feedback!

stitch/key.go
```diff
@@ -29,18 +35,16 @@ func (plaintextKey astPlaintextKey) keys() ([]string, error) {
 
 // Stored in a variable so we can mock it out for the unit tests.
 var getGithubKeys = func(username string) ([]string, error) {
-	usersService := github.NewClient(nil).Users
-	opt := &github.ListOptions{}
-	keys, _, err := usersService.ListKeys(username, opt)
-
+	res, err := http.Get(githubURL + username + ".keys")
```
ejj: I personally don't see much point in `githubURL` being a variable.  You've still go this very specific `username + ".keys"` stuff going on so it's not like you're going to swap out the value for some other service transparently.  I think I'd just encode the string directly.

kklin: Maybe store `const githubKeyURLFmt := "https://github.com/%s.keys"`? Don't feel strongly either way.

mmzhao:
Changes made and unit tests added. Also changed the TestKeys test in `stitch_test.go` since the arguments of `getGithubKeys()` were changed from username to github URL.

stitch/stitch_test.go
```diff
@@ -691,8 +691,8 @@ func TestMachineAttribute(t *testing.T) {
 }
 
 func TestKeys(t *testing.T) {
-	getGithubKeys = func(username string) ([]string, error) {
-		return []string{username}, nil
+	getGithubKeys = func(keyUrl string) ([]string, error) {
```
ejj: Btw, did you consider mocking out `http.Get()` instead?  That way we can cover getGithubKeys as well?

mmzhao: Didn't consider it, but it seems like a better strategy.

mmzhao:
Ready for review.

stitch/key.go
```diff
@@ -28,19 +33,17 @@ func (plaintextKey astPlaintextKey) keys() ([]string, error) {
 }
 
 // Stored in a variable so we can mock it out for the unit tests.
-var getGithubKeys = func(username string) ([]string, error) {
-	usersService := github.NewClient(nil).Users
-	opt := &github.ListOptions{}
-	keys, _, err := usersService.ListKeys(username, opt)
-
+var getGithubKeys = func(keyUrl string) ([]string, error) {
```
ejj: Should just be a normal fun now, also we don't need the comment above.

mmzhao:
Changes made.

PR #550, opened 2016-10-15
Fix public internet NAT rules

kklin:


kklin:
Made `publicInterface` a global variable instead of storing within the minion.

PR #551, opened 2016-10-15
Pre-language modifications

kklin:


stitch/get.go
```diff
@@ -16,13 +17,16 @@ import (
 	"github.com/spf13/afero"
 )
 
-// Break out the download and create functions for unit testing
-var download = func(repo *vcs.RepoRoot, dir string) error {
-	return repo.VCS.Download(dir)
+// ImportGetter provides functions for working with imports.
+type ImportGetter struct {
```
ejj: I'm somewhat surprised this isn't an interface.  Perhaps with a download() function (that could be a noop) and a resolve function that returns the path or a file handle.  Wouldn't that be the more idiomatic way to handle this?

inspect/main.go
```diff
@@ -9,8 +9,6 @@ import (
 	"github.com/NetSys/quilt/stitch"
 )
 
-const quiltPath = "QUILT_PATH"
```
ejj: I don't think the commit message `Make QUILT_PATH a constant` is sufficiently descriptive.   It already was a constant . . .  Also why are we doing this change?

kklin: It seemed appropriate to pull the multiple hard-coded references into a variable..

stitch/get.go
kklin: You mean, make the VCS functions an interface, and `ImportGetter` would have have an object that implements that interface?

That makes sense, I didn't do it because it wasn't my goal in this patch, but it'd clean up the code a lot.

inspect/main.go
ejj: Sure, what I'm saying is that the commit message isn't clear enough about that.  I think the code itself is fine.

stitch/get.go
ejj: Yep, that's what I'm suggesting

kklin:
Reworded commit message and refactored VCS interaction.

kklin:
@ejj The language patch is ready, but blocking on the builds being setup for the new Dockerfiles (I'd like to test the specs)

PR #554, opened 2016-10-17
quiltctl: Fix output redirection for 'quilt logs'

stevenlin598:
Before, output/input from the ssh connection were being copied using
threads in io.Copy. Output from connection became truncated in 'quilt
logs' as quilt terminated without checking if the threads had finished
sending and receiving data. To fix, simply set the output/input of the
connection to be stdin/stdout of os which avoids threads altogether.

### DIFF
--- a/quiltctl/ssh/native.go
+++ b/quiltctl/ssh/native.go
@@ -3,7 +3,6 @@ package ssh
 import (
 	"errors"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


quiltctl/ssh/native.go
```diff
@@ -48,24 +47,9 @@ func (c *NativeClient) Connect(host string, keyPath string) error {
...
-	}
-	go io.Copy(os.Stdout, stdout)
-
-	stderr, err := c.session.StderrPipe()
-	if err != nil {
-		return err
-	}
-	go io.Copy(os.Stderr, stderr)
-
+	c.session.Stdout = os.Stdout
```
kklin: I think it's slightly more readable to set the output of the command in the `Run()` function

quiltctl/ssh/native.go
```diff
@@ -3,7 +3,6 @@ package ssh
 import (
```
kklin: In your commit message, I think you meant thread instead of channel.

kklin: Also, your commit message should be specific that this is fixing a problem seen in `quilt logs`. Remember to keep the short message short (under 70 chars I think? I use the `vim-fugitive` to remind me)

stevenlin598:
Also, stdout and stderr become mixed when printing from quilt logs, which is a bug that still needs to be fixed.

PR #555, opened 2016-10-20
Migrate language from the Lisp to Javascript

kklin:


ejj:
Where's the 5.5% coverage decrease coming from?  That's a huge drop

kklin:
I think it's because the Stitch implementation is now much shorter, and it was very well covered before.

specs/haproxy/haproxy.js
```diff
@@ -0,0 +1,15 @@
+var image = "quilt/haproxy"
+var cfg = "/usr/local/etc/haproxy/haproxy.cfg";
+
+function Haproxy(n, hosts) {
```
tejasmanohar: removed the connectToPublic method?

specs/spark/sparkPI.js
```diff
@@ -0,0 +1,39 @@
...
+    provider: "Amazon",
+    region: "us-west-1",
+    size: "m4.large",
+    diskSize: 32,
+    sshKeys: githubKeys("ejj"), // Replace with your GitHub username.
+});
+deployment
+    .deploy(baseMachine.asMaster())
+    .deploy(baseMachine.asWorker().replicate(nWorker + 1))
+    .deploy(sprk);
```
tejasmanohar: nit: i'd avoid chaining whenever you're not building off the returned value from the last call. e.g. chaining .map .filter etc makes sense since they return new values but for things that mutate (or don't even and just return the object which the method is defined on), it's weird

tejasmanohar:
(Finished reviewing. I'll send my MEAN PR after this is merged / finalized.)

stitch/bindings.js
```diff
@@ -0,0 +1,452 @@
+// The default deployment object. createDeployment overwrites this.
+deployment = new Deployment({});
```
tejasmanohar: why do we need this global?

stitch/bindings.js
```diff
@@ -0,0 +1,452 @@
...
+containerIDCounter = 0;
+
+// Overwrite the deployment object with a new one.
+function createDeployment(deploymentOpts) {
+    deployment = new Deployment(deploymentOpts)
+    return deployment;
+}
+
+function Deployment(deploymentOpts) {
+    this.maxPrice = deploymentOpts.maxPrice || 0.0;
```
tejasmanohar: 0.0 === 0 in JS, all `Number`s are floats

stitch/bindings.js
```diff
@@ -0,0 +1,452 @@
...
+        return this.connectToPublic(range);
+    }
+    this.connections.push(new Connection(range, to));
+    return this;
+}
+
+// publicInternet is an object that looks like another label that can be
+// connected to or from. However, it is actually just syntactic sugar to hide
+// the connectToPublic and connectFromPublic functions.
+publicInternet = {
```
tejasmanohar: scope with `var`?

specs/spark/spark.js
```diff
@@ -0,0 +1,56 @@
+var image = "quilt/spark";
+
+function setImage(newImage) {
```
kklin: I agree, but I don't think it's worth the time to address it now.

stitch/bindings.js
kklin: Changed all globals to be scoped with `var`.

tejasmanohar: Why do we need it scoped outside of a function in general though? Don't we pass `deployment` into all functions that use it?

stitch/get.go
```diff
@@ -124,121 +127,115 @@ func (getter ImportGetter) resolveSpecImports(folder string) error {
...
 
-	for _, ast := range asts {
-		name := parseImport(ast)
-		if name == "" {
-			newAsts = append(newAsts, ast)
-			top = false
-			continue
-		}
+func (getter *ImportGetter) requireImpl(call otto.FunctionCall) (otto.Value, error) {
+	if len(call.ArgumentList) < 1 {
```
ejj: Why isn't this `!= 1`?  I.E. what does it mean to import more than one file in a require?

stitch/stitch.go
```diff
@@ -1,17 +1,24 @@
...
 import (
+	"encoding/json"
 	"fmt"
-	"strings"
-	"text/scanner"
 
-	log "github.com/Sirupsen/logrus"
+	"github.com/robertkrimen/otto"
+
+	// Automatically import the Javascript underscore utility-belt library into
```
ejj: Is this the right thing to do?

stitch/stitch.go
```diff
@@ -85,287 +92,225 @@ func (stitchr Range) Accepts(x float64) bool {
...
+	return ctx, err
 }
 
-func parseKeys(rawKeys []key) []string {
-	var keys []string
-	for _, val := range rawKeys {
-		key, ok := val.(key)
-		if !ok {
-			log.Warnf("%s: Requested []key, found %s", key, val)
+func (ctx *evalCtx) createPortRules() {
```
ejj: I think this function deserves a comment . . . not entirely clear what it's for.

stitch/stitch.go
```diff
@@ -375,3 +320,24 @@ func (cs ConnectionSlice) Get(ii int) interface{} {
 func (cs ConnectionSlice) Len() int {
 	return len(cs)
 }
+
+func stitchError(vm *otto.Otto, err error) otto.Value {
+	return vm.MakeCustomError("StitchError", err.Error())
+}
+
+func toOttoFunc(fn func(otto.FunctionCall) (otto.Value, error)) func(
```
ejj: This is a bit mysterious, could use a function comment

stitch/stitch.go
tejasmanohar: not conventional... but since we don't support NPM etc, maybe it's best

specs/etcd/etcd-example.js
```diff
@@ -0,0 +1,14 @@
+var etcd = require("github.com/NetSys/quilt/specs/etcd/etcd");
+
+var nWorker = 3;
+
+var deployment = createDeployment({});
+
+var baseMachine = new Machine({
+    provider: "Amazon",
+    sshKeys: githubKeys("ejj"), // Replace with your GitHub username.
```
kklin: Even though it returns a list of keys?

stitch/stitch.go
ejj: ack sounds good

PR #557, opened 2016-10-21
cluster: Automatically create machine ACLs for public connections

kklin:
Before, the only notion of machine ACLs were Admin ACLs, which allowed
all traffic to the cluster. Making applications publicly accessible
using Admin ACLs posed a security risk because sensitive ports such as
those for SSH, Docker Swarm, and our Minion GRPC were also made open.

We now automatically generate and enforce ACLs based on the public
connections defined in the stitch.


cluster/provider/awsSpot.go
```diff
@@ -284,47 +284,47 @@ func (clst amazonCluster) List() ([]Machine, error) {
 				Provider: db.Amazon,
```
ejj: How thoroughly have you tried this out?  I think I didn't include them for a reason, but I can't quite remember what it is.

Anyways, if you're confident it works, I'm fine with it.

cluster/provider/awsSpot.go
```diff
@@ -468,7 +469,7 @@ func (clst *amazonCluster) GetCreateSecurityGroup(session EC2Client) (
 	return *csgResp.GroupId, nil, nil
 }
 
-func (clst *amazonCluster) SetACLs(acls []string) error {
+func (clst *amazonCluster) SetACLs(acls []ACL) error {
```
ejj: Is this the right patch for this change?  Might be cleaner to do it one earlier.

cluster/provider/awsSpot.go
kklin: We used to include them, but this patch removed them while trying to decrease indentation I think: https://github.com/NetSys/quilt/commit/19d955efd62032014336dd8d7e5af573bbaf7578#diff-97341d2e1fc06dde22fa710e89b4c1a4L281

cluster/provider/awsSpot.go
```diff
@@ -476,95 +477,44 @@ func (clst *amazonCluster) SetACLs(acls []string) error {
...
-					_, err = session.RevokeSecurityGroupIngress(
-						options)
-					if err != nil {
-						return err
-					}
-				} else {
-					groupIngressExists = true
-				}
+		if len(rangesToAdd) != 0 {
+			logACLs(true, rulesToRemove)
```
kklin: Not sure when I fixed this, but it looks correct now..

cluster/provider/awsSpot.go
kklin: So, have a new commit that changes the interface to be `SetACLs([]ACL)`, and make the final commit just create the machine ACLs automatically?

ejj: Yep that's exactly what I mean.

ejj:
Also this merge conflicts with master.

kklin:
Addressed comments

PR #560, opened 2016-10-22
quilt ssh should fail for partially booted machines

tejasmanohar:
Resolves [quilt ssh should fail for partially booted machines (Trello)](https://trello.com/c/zzPrg9MR/314-quilt-ssh-should-fail-for-partially-booted-machines).


cluster/provider/cloud_config.go
```diff
@@ -112,14 +112,14 @@ date >> /var/log/bootscript.log
 
 export DEBIAN_FRONTEND=noninteractive
 
+ssh_keys="%[2]s"
+setup_user quilt "$ssh_keys"
```
kklin: Doesn't `setup_user` add the user to the docker group, which doesn't exist until docker is installed?

tejasmanohar: Dropped that commit since it's no longer necessary with 47250d8

tejasmanohar: actually may be good anyways so ssh is available asap, will correct

cluster/foreman.go
```diff
@@ -108,9 +109,17 @@ func (fm *foreman) runOnce() {
...
+
 		if connected && !m.connected {
+			// Check that SSH service has started.
+			addr := fmt.Sprintf("%s:ssh", m.machine.PublicIP)
+			conn, err := net.Dial("tcp", addr)
+			if err != nil {
+				log.WithError(err).Debugf("No service running at %s", addr)
+				return
+			}
+			defer conn.Close()
```
kklin: No need to defer

cluster/provider/cloud_config.go
tejasmanohar: fixed

tejasmanohar:
@kklin Addressed concerns.

tejasmanohar:
ah, missed the tests on this fn. will re-open when ready

PR #561, opened 2016-10-23
quilt ssh should fail for partially booted machines

tejasmanohar:
Continuation of #560. Opened new PR, as GitHub doesn't let you re-open after force-pushing for some reason. Slackbot assigned @mmzhao as reviewer.


db/machine.go
```diff
@@ -133,3 +135,32 @@ func (ms MachineSlice) Get(ii int) interface{} {
...
+	if m.machine.PublicIP == "" {
+		return errors.New("machine must have non-zero PublicIP")
+	}
+
+	conn, err := net.Dial("tcp", fmt.Sprintf("%s:ssh", m.machine.PublicIP))
+	if err != nil {
+		return err
+	}
+	conn.Close()
+	return nil
```
mmussomele: `return conn.Close()`?

db/machine.go
```diff
@@ -133,3 +135,32 @@ func (ms MachineSlice) Get(ii int) interface{} {
 func (ms MachineSlice) Len() int {
 	return len(ms)
 }
+
+// MachineClient is an interface over machine clients.
+type MachineClient interface {
```
tejasmanohar: fair point, changed

tejasmanohar:
Resolved comments, @mmussomele.

ejj:
First patch seems fine.  Second patch seems like a huge hack.  Do we really need it?  Wouldn't it make more sense to delay saying the machine is Connected until we can actually talk to the minion over GRPC (i.e. we successfully query it's state)?  That should happen way after ssh comes up right?

ejj:
@tejasmanohar I merged the first patch and filed a trello card on the second.  If you're currently working on it, please assign yourself to the trello car and put in "in progress".  If you haven't started it yet, I don't think it's all that urgent.  I suspect the first patch will go a long way.

https://trello.com/c/6eVAh1tL/386-stricter-connected-semantics

tejasmanohar:
sg, thanks

PR #562, opened 2016-10-24
quiltctl: Remind user to start daemon before running spec

mmzhao:
Before, when a user tried to run a spec before starting the daemon,
they would only get an error message reporting that they were unable
to connect to the daemon. Now, they will also receive a reminder to
start the daemon before attempting to run a spec.

### DIFF
--- a/quiltctl/command/error.go
+++ b/quiltctl/command/error.go
@@ -11,6 +11,7 @@ type DaemonConnectError struct {
 }
 
 func (err DaemonConnectError) Error() string {
-	return fmt.Sprintf("Unable to connect to the Quilt daemon at %s: %s.",
-		err.host, err.connectError.Error())
+	return fmt.Sprintf("Unable to connect to the Quilt daemon at %s: %s. "+
+		"Is the quilt daemon running? If not, you can start it with "+
+		"`quilt daemon`.", err.host, err.connectError.Error())
 }


quiltctl/command/run.go
```diff
@@ -66,6 +66,8 @@ func (rCmd *Run) Run() int {
 	c, err := getClient(rCmd.host)
 	if err != nil {
 		log.Error(err)
+		fmt.Println("Is the quilt daemon running? Please start the daemon " +
```
secant: I think placing this in `quiltctl/command/error.go` would be more appropriate.

kklin: +1, it'd be nice if we wrote this message for other subcommands that talk to the daemon, e.g. `quilt machines`

mmzhao: Would it be better to say "Is the quilt daemon running? Please start the daemon with `quilt daemon` before trying to run a subcommand." then? Or should I still name the specific subcommand (which I believe would require me to pass it through `getClient`)?

secant: I think just saying subcommand is less complex and gets the same message. (not 100% sure, but my hunch is) not being able to reach the daemon should hit that error message regardless of the subcommand.

mmzhao:
Changes made.

quiltctl/command/error.go
```diff
@@ -11,6 +11,8 @@ type DaemonConnectError struct {
 }
 
 func (err DaemonConnectError) Error() string {
-	return fmt.Sprintf("Unable to connect to the Quilt daemon at %s: %s.",
+	return fmt.Sprintf("Unable to connect to the Quilt daemon at %s: %s. "+
+		"Is the quilt daemon running? Please start the daemon with "+
+		"`quilt daemon` before trying to use a subcommand.",
```
ejj: I don't much like the mention of a "subcommand".  That's going to sound scary to a user.  I think I'd just say "Is the quilt daemon running?  If not, you can start it with `quilt daemon`".  Something like that

mmzhao: Sure, just made the change.

quiltctl/command/error.go
```diff
@@ -11,6 +11,7 @@ type DaemonConnectError struct {
 }
 
 func (err DaemonConnectError) Error() string {
-	return fmt.Sprintf("Unable to connect to the Quilt daemon at %s: %s.",
-		err.host, err.connectError.Error())
+	return fmt.Sprintf("Unable to connect to the Quilt daemon at %s: %s. "+
+		"Is the quilt daemon running? If not, you can start it with "+
+		"`quilt daemon`.", err.host, err.connectError.Error())
```
secant: Just FYI, we wrap to 89 characters, but I'm ok with how this looks :)

PR #567, opened 2016-10-29
Send compiled deployment object to QRI

kklin:


kklin:
@mmzhao is reviewing.

stitch/stitch.go
```diff
@@ -215,6 +179,29 @@ func New(specStr string, getter ImportGetter) (Stitch, error) {
 	return spec, nil
 }
 
+// FromJavascriptString gets a Stitch handle from a string containing Javascript code.
+func FromJavascriptString(specStr string, getter ImportGetter) (Stitch, error) {
```
ejj: I think I'd prefer the name `FromJavascript`.  It's clear from the argument types that it's a string.

stitch/stitch.go
```diff
@@ -17,8 +16,7 @@ import (
 
 // A Stitch is an abstract representation of the policy language.
 type Stitch struct {
```
ejj: Did you consider just defining the type instead of making this a struct? I.E. `type Stitch *evalCtx`?

kklin: Would it be OK to just ditch `evalCtx` and move its fields into `Stitch`? Also, what do you think of getting rid of the `QueryFoo` methods?

kklin:
@ejj Made changes. I think you should read it again before merging as I introduced a commit before to simplify the Stitch struct.

kklin:
@ejj Would appreciate it if you upped the priority on reviewing this -- blocks some changes to the module system.

ejj:
Ack.  I want to get this docker exec change in first.  I'll look at this next (probably this evening or tomorrow)

PR #569, opened 2016-10-30
Pre-TLS tweaks

kklin:


kklin:
@secant ping

Not very high priority, TLS probably won't be ready for another week. Just wanted to make sure this was on your radar.

PR #570, opened 2016-10-30
Run tests from the main controller

kklin:


quilt-tester/tests/spark/check_spark.go
```diff
@@ -0,0 +1,41 @@
...
+	"regexp"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+func main() {
+	containers, err := exec.Command("quilt", "containers").Output()
+	if err != nil {
+		log.WithError(err).Fatal("Unable to get containers.")
```
LuiseV: Is there a reason why we log a Fatal error when `quilt containers` fail, but panic in `check_logs.go` when `quilt machines` fail?

quiltctl/ssh/native.go
```diff
@@ -54,6 +57,40 @@ func (c *NativeClient) Connect(host string, keyPath string) error {
 	return nil
 }
 
+var defaultKeys = []string{"id_rsa", "id_dsa", "id_ecdsa", "id_ed25519"}
+
+// Gets the signers for the default private key locations if possible
+func defaultSigners() []ssh.Signer {
```
LuiseV: Do we want a unit test for this?

quilt-tester/tests/spark/check_spark.go
kklin: No good reason, I'll change the panic.

quiltctl/ssh/native.go
kklin: I'll add one!

kklin:
@LuiseV Made changes.

Steven's commit is in here because without it, `quilt logs` doesn't work, and thus the spark test wouldn't work.

PR #572, opened 2016-11-03
scheduler: Make exclusion constraints bidirectional

mmussomele:
Before, when enforcing exclusion placement constraints, they were
only enforced in one direction. That is, if you had a placement
constraint that had TargetLabel = red and OtherLabel = blue, then
blue containers could be placed on a minion with red containers.
Visually, the constraint was Red -> Blue.

The correct behavior is for this constraint to work both ways,
so that blue containers could not be placed on a minion with red
containers. This patch corrects that behavior.


tejasmanohar:
the anonymous function is a code smell IMO but rest looks good

mmussomele:
I'm not a huge fan of it, but I think it would be much messier without it; not sure it's worth the time to make really pretty.

ejj:
I agree with tejas, I think the code is ugly.  I think we should refactor the function.

ejj:
I'll also note, I think to some degree we're premature optimizing here.  It's likely fine to write something that's a tad slower (i.e. recomputes state more often), but more clear.  My main complaint here is I have to think really hard to convince myself that the code is correct.

ejj:
I would also be very happy if we could break validPlacement up into 2 or more functions.  Perhaps one that provides pre-computed state, and then one that uses that state to check the constraints.  Not sure.

PR #575, opened 2016-11-07
specs: Exclusively place proxies for MEAN example

LuiseV:
With this update, we place haproxy containers on all Worker VMs in the MEAN app
example deployment. This allows us to access the web app via any Worker VM,
which simplifies the instructions in DeployMEANapp.md.


tejasmanohar:
FWIW, I think the API for exclusive placement rules is unintuitive (e.g. `true`)

specs/mean/example.js
```diff
@@ -18,7 +18,12 @@ namespace.deploy(baseMachine.asWorker().replicate(3));
 
 var mongo = new Mongo(3);
 var app = new App(3, 8080, { MONGO_URI: mongo.uri("mean-example") });
-var haproxy = new HaProxy(2, app.services(), 8080);
+var haproxy = new HaProxy(3, app.services(), 8080);
+
+// Places all haproxy containers on separate Worker VMs.
+// This is just for convenience for the example instructions, as it allows us to
+// access the web application by using the IP address of any Worker VM.
+haproxy.service.place(new LabelRule(true, haproxy.service));
```
tejasmanohar: Not sure I follow. HAP runs on 80 so can't you always use the public IP of any VM?

LuiseV:
@tejasmanohar I think we all agree on that. Placement is going to go through a major redesign at some point when we know what the right design is.

specs/mean/example.js
LuiseV: Nope, you can only access the website through VMs that have a HAP container

tejasmanohar: oh this makes it placed _everywhere_. hm, not sure it's a great default. i'd rather explain quilt containers & machines to the user

tejasmanohar: once floating IPs are in, it won't matter as much

LuiseV: Yeah, floating IPs should eliminate this step. Also, I agree that this is not a good production default, which is why I left the comment saying that this is just to simplify our guide. For the purpose of the brief introduction to Quilt, I think it is fine. Finding the right machine by using `quilt containers` is not intuitive right now, and I don't think it is crucial for the purpose of the getting started guide. As soon as floating IPs are in, I we should be able to scratch this :)

kklin: @LuiseV do you have any idea how we should improve `quilt containers`? It seems like we need some sort of unified view with the machine it's scheduled on, etc. Also prettier formatting and a better way to reference items than the database ID.
Might be something @stevenlin598 could do.

tejasmanohar: machine ID seems like a good addition

tejasmanohar: or maybe a "quilt status" that shows machines & containers in one place

PR #577, opened 2016-11-07
travis: Build the production Dockerfile

ejj:
By building the production Dockerfile as part of CI, we'll catch
errors before they get into the docker build system.  This would have
caught a rather annoying error much earlier in the development
process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: go
 
+sudo: required
+
+services:
+      - docker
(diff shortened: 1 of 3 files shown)


ejj:
Ideally, this should fail travis as docker is currently broken.  Once we've fixed docker, travis should be happy and we should mere.

ejj:
Build failed for the right reason.

PR #579, opened 2016-11-08
refactor mean spec

tejasmanohar:
incorporates comments from original PR


README.md
```diff
@@ -34,14 +34,21 @@ web proxy:
...
-    var App = require("github.com/NetSys/quilt/specs/mean/app");
+    var Node = require("github.com/NetSys/quilt/specs/node/node");
     var HaProxy = require("github.com/NetSys/quilt/specs/haproxy/haproxy").Haproxy;
     var Mongo = require("github.com/NetSys/quilt/specs/mongo/mongo");
 
     // Create 3 replicated instances of each service.
     var mongo = new Mongo(3);
-    var app = new App(3, 8080, { MONGO_URI: mongo.uri("mean-example") });
-    var haproxy = new HaProxy(3, app.services(), 8080);
+    var app = new Node({
```
LuiseV: Is it common practice to pass a config object rather than separate arguments? I see that it is a nice way to name the arguments, so I'm just curious.

specs/mean/example.js
```diff
@@ -17,8 +17,15 @@ namespace.deploy(baseMachine.asMaster());
 namespace.deploy(baseMachine.asWorker().replicate(3));
 
 var mongo = new Mongo(3);
-var app = new App(3, 8080, { MONGO_URI: mongo.uri("mean-example") });
-var haproxy = new HaProxy(3, app.services(), 8080);
+var app = new Node({
+  nWorker: 3,
+  image: "quilt/mean-service",
```
LuiseV: I like that we pass the image to the constructor. We should update docs/DeployMEANapp.md accordingly. Feel free to change it yourself or ping me, if you'd like me to do it.

specs/node/node.js
```diff
@@ -0,0 +1,42 @@
...
+  var hosts = this._app.children();
+  for (var i = 0; i < containers.length; i++) {
+    containers[i].setEnv("HOST", hosts[i]);
+  }
+
+  this._app.connect(this._port, this._app);
+};
+
+Node.prototype.deploy = function(deployment) {
+  deployment.deploy([this._app]);
```
LuiseV: Maybe `this.services()` instead of `[this._app]`?

README.md
tejasmanohar: not a "standard" or something a linter would suggest but pretty common

tejasmanohar: [slightly nicer in es6](https://medium.com/@tejasmanohar/an-argument-for-named-parameters-e2ba3920f313)

specs/mean/example.js
tejasmanohar: i'll let you change that... i think this can make the guide more library-oriented vs copy-paste

README.md
```diff
@@ -34,14 +34,21 @@ web proxy:
 [//]: # (b1)
 ```javascript
     // `App` is a Node.js application using Express, AngluarJS, and MongoDB.
```
tejasmanohar: I moved the comment down to `var app =` since the imported Node constructor can be used for any type of Node app

tejasmanohar:
cc @ejj for review

README.md
```diff
@@ -40,7 +40,13 @@ web proxy:
 
     // Create 3 replicated instances of each service.
     var mongo = new Mongo(3);
-    var app = new App(3, 8080, { MONGO_URI: mongo.uri("mean-example") });
+    var app = new App({
+      nWorker: 3,
+      port: 8080,
+      env: {
+        MONGO_URI: mongo.uri("mean-example")
```
ejj: This isn't the right patch to fix it, but why don't we call mongo db "mongo"?

PR #582, opened 2016-11-08
spark: Fix incorrect instruction in README.md

ejj:
The Spark README.md suggested that we run `quilt exec -it` when `quilt
exec` alone is correct.


mmussomele:
Looks good

PR #584, opened 2016-11-11
docs: Update to Quilt.js and current console formatting

LuiseV:


docs/GettingStarted.md
```diff
@@ -36,9 +36,10 @@ follow the instructions
 ## Configure A Cloud Provider
 
 Below we discuss how to setup Quilt for Amazon EC2.  Other providers are
-supported as well, including [Vagrant](Vagrant.md), and Google Compute Engine.
-Since Quilt deploys systems consistently across providers, the details of the
-rest of this document will apply no matter what provider you choose.
+supported as well, including [Vagrant](Vagrant.md), Google Compute Engine, and
+soon also DigitalOcean. Since Quilt deploys systems consistently across
```
ejj: I don't like words like "soon", the problem is as the project changes it's very easy for those words to get stale.  I.E. once we support digital ocean, we'll forgot to remove this word.

Also, I don't want to say we support Vagrant cause it's broken

docs/GettingStarted.md
```diff
@@ -52,62 +53,49 @@ aws_secret_access_key = <YOUR_SECRET_KEY>
...
+run specs/example.js`. Quilt will set up several Ubuntu VMs on your cloud
 provider as Workers, and these Workers will host Nginx Docker containers.
 
 
 ### Accessing the Worker VM
 It will take a while for the VMs to boot up, for Quilt to configure the network,
-and for Docker containers to be initialized. "New connection" message in the console
-output indicates that new VM is fully booted and has began communicating with
-Quilt.
+and for Docker containers to be initialized. When a machine is labelled
```
ejj: "labeled".  Apparently both forms are acceptable, but one l is preferred in north america.  At any rate, the rest of the project uses one l.

I also might change the word to "marked".  Label is a loaded term in the project.

docs/GettingStarted.md
```diff
@@ -52,62 +53,49 @@ aws_secret_access_key = <YOUR_SECRET_KEY>
...
 The public IP of the Worker VM can be deduced from the console output. The
-following output shows the Worker VM's public IP to be 52.39.213.45:
+following output shows the Worker VM's public IP to be 52.53.179.179:
 ```
-INFO [Jun  7 17:24:43.660] New connection.
-machine=Machine-3{Role=Worker, Provider=Amazon, Region=us-west-2, Size=m3.medium, DiskSize=32, CloudID=sir-03hg1gw1, PublicIP=52.39.213.45, PrivateIP=172.31.6.254}
-INFO [Jun  7 17:24:43.660] New connection.
-machine=Machine-2{Role=Master, Provider=Amazon, Region=us-west-2, Size=m3.medium, DiskSize=32, CloudID=sir-03hdcezk, PublicIP=52.25.237.134, PrivateIP=172.31.1.198}
+INFO [Nov  9 00:02:27.535] db.Machine:
+Machine-2{Master, Amazon us-west-1 c4.large, sir-4yj8fa2g, PublicIP=52.53.214.121, PrivateIP=172.31.11.109, Disk=32GB, Connected}
```
ejj: Not the patch to fix this, but I wish our specs defaulted to m4.large.  c4 is more expensive for no apparent reason.  Maybe file a trello card?

docs/GettingStarted.md
```diff
@@ -132,9 +120,13 @@ other and your local computer.
 
 ### Loading the Nginx Webpage
```
ejj: Also probably better for a separate patch, but we should probably rewrite this to be in terms of a trivial node server or something.  (i.e. the node app without mongo or ha proxy).  Not sure

docs/GettingStarted.md
LuiseV: So would you rather not mention DigitalOcean at all for now?

docs/GettingStarted.md
LuiseV: They actually do default to m4.large now. This output is from when the spec was run in the summer. I'll update it.

docs/GettingStarted.md
ejj: Yeah I think so.

LuiseV:
Updated, @ejj

PR #586, opened 2016-11-13
Network driver

mmussomele:


cluster/provider/cloud_config.go
```diff
@@ -99,6 +100,9 @@ export DEBIAN_FRONTEND=noninteractive
 ssh_keys="{{.SSHKeys}}"
 setup_user quilt "$ssh_keys"
 
+sudo mkdir /run/docker/plugins
+sudo chmod -R /run 0666
```
mmussomele: This is one thing I'm not too certain about, would appreciate feedback.

minion/ip/ip.go
```diff
@@ -39,32 +40,96 @@ var (
...
+// WithIPs returns a new Pool with the combined IP set of p's IPs and the given IPs
+func (p Pool) WithIPs(ips []string) Pool {
+	ipSet := map[string]struct{}{}
+	for _, ip := range ips {
+		ipSet[ip] = struct{}{}
+	}
+
+	for ip := range p.ipSet {
+		ipSet[ip] = struct{}{}
+	}
```
mmussomele: I thought about this and I think it's fine, since those IPs won't be allocated by the pool anyway. It's not harmful, but I suppose there could be benefit to giving the user feedback if they pass in a set that has an IP outside the subnet. I'll put a comment in for now

mmussomele: Scratch that, I'll add a warning

minion/docker/docker.go
```diff
@@ -280,30 +317,42 @@ func (dk Client) list(filters map[string][]string, all bool) ([]Container, error
...
 	if err != nil {
 		return Container{}, err
 	}
 
 	env := make(map[string]string)
-	for _, value := range c.Config.Env {
+	for _, value := range dkc.Config.Env {
 		e := strings.Split(value, "=")
 		if len(e) > 1 {
 			env[e[0]] = e[1]
```
mmussomele: The docker client returns the `Env` as a `[]string`

minion/scheduler/worker.go
```diff
@@ -54,17 +67,48 @@ func runWorker(conn db.Conn, dk docker.Client, myIP string) {
...
+
+	return good, bad
+}
+
+func syncWorker(dbcs []db.Container, dkcs []docker.Container, pool ip.Pool) (
+	changed []db.Container, toBoot, toKill []interface{}) {
 
 	pairs, dbci, dkci := join.Join(dbcs, dkcs, syncJoinScore)
 
 	for _, i := range dkci {
```
mmussomele: The next line does a type assertion to turn this into a docker container and names it dkc, it's just a placeholder

mmussomele:
I've addressed @LuiseV's comments. The patch isn't ready for merge; needs to be tested with the network waits removed, but in the meantime a preemptive review of the design and such would be good @ejj

specs/haproxy/docker-entrypoint.sh
```diff
@@ -1,13 +1,7 @@
...
 set -e
 
-timestamp() {
-    until ping -q -c1 localhost > /dev/null 2>&1; do
-        sleep 0.5
-    done
-    date -u +%s > /tmp/boot_timestamp
-}
-timestamp &
+date -u +%s > /tmp/boot_timestamp
```
tejasmanohar: oops, this file can actually be deleted since scripts/run.go is used now

specs/redis/run.go
```diff
@@ -51,8 +40,6 @@ func main() {
 			os.Exit(1)
 		}
 
-		resolveHostname(master)
```
tejasmanohar: is the master env var still needed? maybe we can remove that validation too & stop passing it from the spec?

tejasmanohar: (same for other specs)

minion/ip/ip.go
```diff
@@ -39,32 +40,112 @@ var (
...
+		prefix: ToInt(prefix),
+		mask:   MaskToInt(mask),
+		ipSet:  map[string]struct{}{},
+	}
+}
+
+// WithIPs returns a new Pool with the combined IP set of p's IPs and the given IPs. If
+// the set contains an IP outside the pool's subnet, an error is returned as a warning,
+// but the ips are still added to the new pool.
+func (p Pool) WithIPs(ips []string) (Pool, error) {
```
ejj: Why not just have WithIPs modify `p` instead of returning a new pool?  Seems a tad more clear no?

I get the functional style is cool, but I don't see the benefit in this case given that Pool is more-or-less a map like thing.

minion/ip/ip.go
```diff
@@ -39,32 +40,112 @@ var (
 	MaxMinionCount = int(math.Pow(2, float64(minionMaskBits-quiltMaskBits))+0.5) - 1
 )
 
+// Pool represents a set of IP addresses that can be given out to requestors.
+type Pool struct {
+	prefix uint32
```
ejj: Probably Cleaner to use an `ip.IPNet`.  You could even embed it if it's clean

minion/ip/ip.go
```diff
@@ -39,32 +40,112 @@ var (
...
+
+	return Pool{
+		prefix: p.prefix,
+		mask:   p.mask,
+		ipSet:  ipSet,
+	}, err
+}
+
+// Request selects a random IP address from the pool that hasn't already been given out.
+func (p Pool) Request() (net.IP, error) {
```
ejj: I think I like the name "Allocate" better.

Btw, it's weird that WithIPs creates a new pool, but Request modifies `p`.  Another argument in favor of tweaking `WithIPs`

minion/ip/ip.go
```diff
@@ -39,32 +40,112 @@ var (
...
+	randStart := Rand32() & ^p.mask
+	randIP := randStart
+	for {
```
ejj: Why not use a standard for loop?  I.E.
`for randIP = randStart + 1; randIP != randStart; randIP = (randIP + 1) & ^p.mask` {`

minion/ip/ip.go
```diff
@@ -39,32 +40,112 @@ var (
...
+		if randIP == randStart {
+			return net.IPv4zero, errors.New("IP pool exhausted")
+		}
+	}
+}
+
+// Free releases the given IP, allowing the Pool to allocate it again from Request.
+func (p Pool) Free(ip net.IP) error {
+	if _, ok := p.ipSet[ip.String()]; !ok {
+		return fmt.Errorf("IP %s not present in pool", ip.String())
```
ejj: I think this should likely be a panic instead.  It's not an error that can occur during normal operation, instead it's an exceptional condition that's only the result of a bug.  Thus it should panic for the same reason you would do an assertion.

minion/ip/ip.go
```diff
@@ -39,32 +40,96 @@ var (
...
-			ipSet[ip.String()] = struct{}{}
+			ips = append(ips, ipString)
 		} else {
 			unassigned = append(unassigned, k)
 		}
 	}
 
 	// Don't assign the IP of the default gateway
-	ipSet[network.GatewayIP] = struct{}{}
+	ips = append(ips, network.GatewayIP)
```
ejj: @mmussomele thoughts on @LuiseV's comment?  I'm not sure either way.

minion/ip/ip.go
```diff
@@ -39,32 +40,112 @@ var (
...
+		return fmt.Errorf("IP %s not present in pool", ip.String())
+	}
+
+	delete(p.ipSet, ip.String())
+	return nil
+}
+
 // Sync takes a map of IDs to IPs and creates an IP address for every entry that's
 // missing one.
 func Sync(ipMap map[string]string, prefixIP net.IP, mask net.IPMask) {
```
ejj: My other thought here, I don't think Sync should be in this library.  It's too weird, application specific logic.  Should probably pull it into the etcd code.

I also have a general sense that it could be simplified a lot if it's reorganized a bit.  Not sure, think about it

ejj:
for `minion: Make ip package ignorant of the default gateway`.  I think we should ditch the patch and pull the logic into the user instead.  That function is so wonky, I don't think it should be in the library basically.

minion/scheduler/worker.go
```diff
@@ -54,17 +67,48 @@ func runWorker(conn db.Conn, dk docker.Client, myIP string) {
...
-		toKill = append(toKill, i.(docker.Container))
+		dkc := i.(docker.Container)
+		err := pool.Free(net.ParseIP(dkc.IP))
```
ejj: And actually, come to think of it.  I'm not sure we need the `ip.Pool` struct at all.  If it's just used internally, in this library, would be simpler to use `net.IPNet` and a set of used IPs.  Would simplify things a lot.

ejj: Also note, `ip.Sync` isn't used anymore with this patch.  And thus `ip.WithIPs` isn't used anymore either.

cluster/provider/cloud_config.go
kklin: Actually 0755 might be more standard.

minion/ip/ip.go
mmussomele: A few things, `randStart+1` is not guaranteed to be within the subnet, and also if it happens that `randStart` is the only IP left, we'll say that the pool is exhausted this way. Not a huge deal, but I prefer this way

minion/ip/ip.go
```diff
@@ -39,32 +40,112 @@ var (
...
+// but the ips are still added to the new pool.
+func (p Pool) WithIPs(ips []string) (Pool, error) {
+	subnet := FromInt(p.prefix)
+	mask := net.IPMask(FromInt(p.mask))
+
+	var err error
+	ipSet := map[string]struct{}{}
+	for _, ip := range ips {
+		addr := Parse(ip, subnet, mask)
+		if err == nil && addr.Equal(net.IPv4zero) {
```
mmussomele: It's simpler to just use `net.ParseIP` and check for inclusion using the pool's IPNet

minion/docker/docker.go
```diff
@@ -87,19 +92,27 @@ func New(sock string) Client {
 }
 
 // Run creates and starts a new container in accordance RunOptions.
-func (dk Client) Run(opts RunOptions) (string, error) {
+func (dk Client) Run(opt RunOptions) (string, error) {
```
mmussomele: I wanted the dk.create line to fit on one line, kinda silly, I'll put it back

minion/docker/docker.go
```diff
@@ -234,30 +271,42 @@ func (dk Client) list(filters map[string][]string, all bool) ([]Container, error
...
+		Path:   dkc.Path,
+		Args:   dkc.Args,
+		Pid:    dkc.State.Pid,
 		Env:    env,
-		Labels: c.Config.Labels,
-	}, nil
+		Labels: dkc.Config.Labels,
+	}
+
+	if network != "" {
```
mmussomele: Ah yes I like that better

ejj:
Skimming it now, looks a lot better.  Working on a detailed review.  I think the code coverage could be a lot better for the new code though.  We're adding a lot of uncovered lines that would be easy to test.

minion/network/plugin/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil, errors.New("no IP address provided")
+	}
+
+	mac := req.Interface.MacAddress
+	if mac != "" {
+		return nil, fmt.Errorf("expected no mac in request, got mac=%s", mac)
+	}
+
+	d.Lock()
+	defer d.Unlock()
```
ejj: I'd tweak this so you aren't holding the lock past when you need it.

minion/network/plugin/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+
+// Join creates a Veth pair for the given endpoint ID, returning the interface info.
+func (d *driver) Join(req *dnet.JoinRequest) (*dnet.JoinResponse, error) {
+	if !d.endpointExists(req.EndpointID) {
+		return nil, fmt.Errorf("no such endpoint: %s", req.EndpointID)
+	}
+
+	// We just need to create the Veth and tell Docker where it should go; Docker
+	// will take care of moving it into the container and renaming it.
+	tempPeer, err := network.AddVeth(req.EndpointID)
```
ejj: Isn't there a race condition here?  I.E, it might no longer exist at this point right?

minion/network/plugin/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+)
+
+var (
+	networkSocket = NetworkName + ".sock"
+	pluginSocket  = filepath.Join(pluginDir, networkSocket)
+)
+
+type driver struct {
+	sync.RWMutex
+	endpointMap map[string]struct{}
```
ejj: See slack, not sure we should have this at all.  Instead we could define the veth as the endpoint and use the OS system calls to interface with it.

minion/network/plugin/plugin_test.go
```diff
@@ -0,0 +1,309 @@
...
+		t.Fatal("Leave should not error if endpoint is paired")
+	}
+
+	err = d.DeleteEndpoint(&dnet.DeleteEndpointRequest{EndpointID: "0"})
+	if err != nil {
+		t.Fatal("Delete endpoint should not error if endpoint is allocated")
+	}
+}
+
+func TestCreateNetwork(t *testing.T) {
```
ejj: I might combine all of these into a single "TestNoop", also they should all use assert.

Also might say the tests can be Parllel

minion/docker/docker.go
```diff
@@ -99,7 +110,18 @@ func (dk Client) Run(opts RunOptions) (string, error) {
...
+	if opts.IP != "" {
+		nc = &dkc.NetworkingConfig{
+			EndpointsConfig: map[string]*dkc.EndpointConfig{
+				opts.NetworkMode: {IPAddress: opts.IP},
+			},
+		}
+	}
+
+	id, err := dk.create(opts.Name, opts.Image,
+		opts.Args, opts.Labels, env, &hc, nc)
```
mmussomele: huh, I guess not (was one character over before)

minion/docker/docker.go
```diff
@@ -112,6 +134,19 @@ func (dk Client) Run(opts RunOptions) (string, error) {
 	return id, nil
 }
 
+// ConfigureNetwork makes a request to docker to create a network running on driver with
+// the given subnet.
+func (dk Client) ConfigureNetwork(driver string, subnet net.IPNet) error {
+	_, err := dk.CreateNetwork(dkc.CreateNetworkOptions{
+		Name:   driver,
+		Driver: driver,
+		IPAM: dkc.IPAMOptions{
```
mmussomele: So you can't stop docker from trying to supply IP addresses, so the only thing we can do is tell it which ones to supply. If the IP address we give docker isn't part of the subnet it chose, it will reject it, so this is how we tell docker the subnet too.

minion/ip/ip.go
```diff
@@ -39,33 +40,80 @@ var (
...
+	return nil
+}
+
+// Allocate selects a random IP address from the pool that hasn't already been given out.
+func (p Pool) Allocate() (net.IP, error) {
+	prefix := ToInt(p.IP)
+	mask := MaskToInt(p.Mask)
+
+	randStart := Rand32() & ^mask
+	for offset := uint32(0); offset < ^mask+1; offset++ {
```
sirspinach: I think `offset <= ^mask` is better here because `^mask + 1` could overflow

minion/ip/ip.go
```diff
@@ -39,33 +40,80 @@ var (
...
+func NewPool(prefix net.IP, mask net.IPMask) Pool {
+	return Pool{
+		IPNet: net.IPNet{IP: prefix, Mask: mask},
+		ipSet: map[string]struct{}{},
+	}
+}
+
+// AddIP adds the given IP to this Pool's IP set.
+// It is an error to attempt to add an IP outside the Pool's subnet.
+func (p Pool) AddIP(ip string) error {
```
sirspinach: Since this func is the opposite of Free, `Block` might be a better name for this guy

minion/network/plugin/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+	innerVeth = "eth"
+)
+
+var (
+	networkSocket = NetworkName + ".sock"
+	pluginSocket  = filepath.Join(pluginDir, networkSocket)
+)
+
+type driver struct {
+	sync.RWMutex
```
mmussomele: not relevant now that the driver has no state

minion/network/plugin/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+	// exists.
+	for {
+		if _, err := os.Stat(pluginSocket); err == nil {
+			return
+		}
+		time.Sleep(500 * time.Millisecond)
+	}
+}
+
+func newDriver() *driver {
```
mmussomele: Yea I think it exists from when the driver was more complicated to init, I'll delete it

minion/docker/docker.go
ejj: Can we talk about this next time you're in?  Sounds reasonable, but I just want to make sure I understand the docker model for dealing with IPs

minion/docker/docker.go
```diff
@@ -112,6 +134,19 @@ func (dk Client) Run(opts RunOptions) (string, error) {
 	return id, nil
 }
 
+// ConfigureNetwork makes a request to docker to create a network running on driver with
+// the given subnet.
+func (dk Client) ConfigureNetwork(driver string, subnet net.IPNet) error {
```
mmussomele: It conflicts with the docker-client `CreateNetwork` function

minion/etcd/network_test.go
```diff
@@ -283,27 +283,26 @@ func testUpdateWorkerDBC(t *testing.T, view db.Database) {
 	minion.Self = true
 	view.Commit(minion)
 
-	for id := 1; id < 4; id++ {
+	for id := 1; id < 3; id++ {
```
ejj: Why this change?

minion/network/plugin/plugin.go
```diff
@@ -0,0 +1,207 @@
...
+		return err
+	} else if ok {
+		return fmt.Errorf("endpoint %s already exists", eid)
+	}
+
+	return nil
+}
+
+func endpointExists(eid string) (bool, error) {
+	_, name := network.VethPairNames(eid)
```
ejj: The one question I have ... will this work when docker renames the veth?

minion/etcd/network_test.go
mmussomele: Because of the way UpdateWorker matches on things, if the container with ID 3 was committed, it would do something different.

minion/network/plugin/plugin.go
mmussomele: Docker only renames the inner veth, which we already did before this patch

PR #587, opened 2016-11-14
Floating IPs

tejasmanohar:
v0 of Floating IPs. They're implemented as machine rules, which are at the label/service, not container-level, so that's a bit awkward since 1 IP : 1 container, but I think we should fix this in a refactor of placement rules (and maybe under the hood-- e.g. assign a rule of ["eip-1", "eip-2"] to a label with 2 containers).


cluster/provider/amazon_test.go
```diff
@@ -475,3 +476,70 @@ func TestStop(t *testing.T) {
 		},
 	)
 }
+
+func TestAssociateFloatingIP(t *testing.T) {
```
tejasmanohar: do we prefer to just test public functions (e.g. Boot)? since there are multiple calls to the same AWS APIs and they all accept pointer arguments (making it hard to specify specific input (hence the use of mock.Anything)), I find it harder to read but can figure something out if that's preferred // cc @kklin

tejasmanohar:
@kklin is reviewing.

tejasmanohar:
oops, need to update minion tests

tejasmanohar:
Updated spec examples & code to allow users to enter their floating IP's (GCE NAT / AWS Elastic IP) public address instead of ID as per @ejj's suggestion.

cluster/cluster.go
```diff
@@ -283,6 +286,8 @@ func syncDB(cloudMachines []provider.Machine, dbMachines []db.Machine) (
 			return -1
 		case m.Region != "" && dbm.Region != m.Region:
 			return -1
+		case m.FloatingIP != "" && dbm.FloatingIP != m.FloatingIP:
```
kklin: Can we just do `dbm.FloatingIP != m.FloatingIP`? I think the `m.Region != "" && ...` etc is leftover from before regions and sizes were implemented in all providers.

Also, this won't disassociate floating IP's right? What if you want to switch the IP on a machine for whatever reason?

stitch/bindings.js
```diff
@@ -420,6 +422,9 @@ function MachineRule(exclusive, optionalArgs) {
     if (optionalArgs.region) {
         this.region = optionalArgs.region;
     }
+    if (optionalArgs.floatingIp) {
+      this.floatingIp = optionalArgs.floatingIp;
```
kklin: I thought we were going to error if someone applies a floatingIP placement rule with a service that has multiple containers, and accepts connections from `publicInternet`?

cluster/provider/awsSpot.go
```diff
@@ -385,6 +411,84 @@ func (clst *amazonCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
 	return pickBestSize(constants.AwsDescriptions, ram, cpu, maxPrice)
 }
 
+func (clst amazonCluster) associateFloatingIPs(ips []floatingIP) error {
```
kklin: Not saying we should do that before this commit, btw. I do think the code can be simplified, though. Will think about it later.

stitch/bindings.js
```diff
@@ -262,7 +262,8 @@ Service.prototype.getQuiltPlacements = function() {
             otherLabel: placement.otherLabel || "",
             provider: placement.provider || "",
             size: placement.size || "",
-            region: placement.region || ""
+            region: placement.region || "",
+            floatingIp: placement.floatingIp || ""
```
tejasmanohar: i've always seen acronyms treated as words in JS camelCase (unlike Go), but I don't feel strongly. Let me know if you think we should still change this.

cluster/cluster.go
tejasmanohar: yep, i think you're right. changed.

if you're also concerned about disassociating IPs on AWS, I already tested that we don't need to add anything to Stop(). next associate call just unassigns & re-assigns.

cluster/provider/awsSpot.go
tejasmanohar: agreed. gave it a 2nd look & made some changes. let me know what you think!

specs/example-ip-gce.js
```diff
@@ -0,0 +1,21 @@
+// https://cloud.google.com/compute/docs/configure-instance-ip-addresses#reserve_new_static
```
tejasmanohar: done

stitch/bindings.js
tejasmanohar: added

stitch/bindings.js
kklin: Ah, didn't realize that was the standard in Javascript. Let's leave it camelcase then.

cluster/provider/awsSpot.go
```diff
@@ -455,6 +541,75 @@ OuterLoop:
 	return errors.New("timed out")
 }
 
+// getInstanceIDs builds a map of spot IDs to instance IDs from spotIDs
+// for a given region.
+func (clst amazonCluster) getInstanceIDs(region string,
```
kklin: I think we could simplify `getInstanceIDs` and `waitOnInstances`. They're pretty similar.

What if we defined a (non-blocking) `getInstances([]awsID) (map[awsID]*instance, error)`? `associateFloatingIPs` would then wait until all instances were `!= nil && running`.. We might also define some sort of `waiter` function, not sure.

What do you think?

tejasmanohar: I started on this path, but I found the control flow more awkward. Since all AWS actions require grouping by region, I think it's much cleaner to group by region at the very beginning of the `associateFloatingIPs` function. Otherwise, you group aws IDs on every action (e.g. after getting instances from `getInstances`, you have to group results again to associate floating IPs-- which is also weird with the waiter function)

tejasmanohar: To illustrate, here's a (rough, definitely non-functional) version that takes that approach-- https://gist.github.com/tejasmanohar/50d26d768f09b8ae7536920991b59abb

kklin: I'm not saying we get rid of grouping by region first, but that we just define a single `getInstances` function rather than having both `getInstanceIDs` and `waitOnInstances`.

```
func wait(pred func() bool) error

for region, floatingIPs := range toAssociate {
    wait(func() bool {
        for _, inst := range getInstances(region, spotIDs) {
            if inst == nil || inst.state != running {
                return false
            }
        }
        return true
    }

    instanceMap := getInstances(region, spotIDs)
    for _, ip := range floatingIPs {
        associate(instanceMap[ip.spotID].instanceID, ip.address)
    }
}
```

tejasmanohar: ah ok, ^ that would mean we retry all spot IDs everytime... but i guess that's fine to start.

tejasmanohar: anyways, i'll do that! but probably without the wait() fn and just a loop

kklin: It's the same runtime because we're still iterating over all the current SpotIDRequests right? Either way, I'd happily trade more obviously correct code for something less efficient until it shows up as a bottleneck.

btw, the reason why I proposed the `wait()` function is because of `clst.wait`. Not sure how much cleaner the function would make that code, though.

tejasmanohar:
Ready for 2nd review @kklin

stitch/bindings.js
tejasmanohar: just how i've seen it & how we do things at work at least, i only starting writing `floatingIP` with the capital `P` because of gometalinter :P

cluster/cluster.go
```diff
@@ -283,6 +286,8 @@ func syncDB(cloudMachines []provider.Machine, dbMachines []db.Machine) (
 			return -1
 		case m.Region != "" && dbm.Region != m.Region:
```
tejasmanohar: saw loops of `Attempting to Boot machines`, `Attempting to halt machines` after removing the != "" checks so requires more investigation (probably outside of this PR?)

tejasmanohar: nevermind, that seems unrelated. strange enough, after attempting to boot/halt a bunch, it still prints the same spot instance request IDs

kklin: Why do you say it's unrelated? Could you post the log? But yeah, no need to spend too much time on it, could you just file a card on the issue?

Also, it's possible that it created and destroyed other spot requests, and kept the same ones in the DB because of our join score.

stitch/stitch_test.go
```diff
@@ -73,6 +73,20 @@ func TestMachine(t *testing.T) {
 			},
 		},
 	)
+
+	checkMachines(t, `var baseMachine = new Machine({
+	  provider: "Amazon",
+	  floatingIp: "eipalloc-123"
```
kklin: Same here re making the test use the IP address

cluster/provider/awsSpot.go
```diff
@@ -385,6 +427,115 @@ func (clst *amazonCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+			}
+
+			for _, inst := range instances {
+				if inst == nil ||
+					*inst.State.Name != ec2.InstanceStateNameRunning {
+					return true
+				}
+			}
+			return false
+		}, 100, 10*time.Second)
```
kklin: I think it'd be much more intuitive if `Wait` took a timeout rather than number of times to try.

cluster/provider/awsSpot.go
```diff
@@ -385,6 +427,115 @@ func (clst *amazonCluster) ChooseSize(ram stitch.Range, cpu stitch.Range,
...
+		if err != nil {
+			return err
+		}
+
+		for _, ip := range floatingIPs {
+			_, err := session.AssociateAddress(&ec2.AssociateAddressInput{
+				InstanceId: aws.String(
+					*instances[ip.spotID].InstanceId,
```
kklin: Is this already a *string? If so, we can just assign it directly InstanceId

stitch/stitch_test.go
tejasmanohar: fixed

minion/scheduler/master_test.go
```diff
@@ -75,6 +75,7 @@ func TestCleanup(t *testing.T) {
 			Exclusive:   true,
 			TargetLabel: "1",
 			Region:      "Region1",
+			FloatingIP:  "eipalloc-123",
```
tejasmanohar: fixed

cluster/provider/awsSpot.go
tejasmanohar: fixed

cluster/provider/awsSpot.go
tejasmanohar: fixed, set 15m arbitrarily for now

tejasmanohar:
Ready for review by @ejj.

util/util.go
```diff
@@ -144,3 +146,25 @@ func StrStrMapEqual(x, y map[string]string) bool {
 	}
 	return true
 }
+
+// Wait retries a function up to X times until it returns false with Y constant
+// backoff, erroring if it never completes.
+func Wait(f func() bool, backoff time.Duration, timeout time.Duration) error {
```
tejasmanohar: done

tejasmanohar:
@ejj Rebased & ready for review again. Thanks!

ejj:
Eh, sorry @tejasmanohar could you rebase again?  Seems to be conflicting with some new changes.

cluster/provider/provider.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"fmt"
+
+	"github.com/NetSys/quilt/db"
+	"github.com/NetSys/quilt/stitch"
+)
+
+// Machine represents an instance of a machine booted by a Provider.
+type Machine struct {
```
ejj: This looks like a rebasing error.  The machine struct is already in the machine package.

cluster/provider/provider.go
```diff
@@ -0,0 +1,121 @@
...
+	FloatingIP string
+	Size       string
+	DiskSize   int
+	SSHKeys    []string
+	Provider   db.Provider
+	Region     string
+}
+
+// ACL represents allowed traffic to a machine.
+type ACL struct {
```
ejj: Same with the ACL struct

ejj: Yeah most of this file should be deleted.

engine/engine.go
```diff
@@ -155,6 +156,9 @@ func machineTxn(view db.Database, stitch stitch.Stitch) error {
 			return -1
 		case dbMachine.Size != "" && stitchMachine.Size != dbMachine.Size:
 			return -1
+		case dbMachine.FloatingIP != "" &&
```
ejj: Could you join it and just change the floating IP, or do you have to delete the machine and re-create it to set it's floating IP

engine/engine.go
```diff
@@ -177,7 +181,6 @@ func machineTxn(view db.Database, stitch stitch.Stitch) error {
 
 	for _, bootSet := range bootList {
 		bootSet := bootSet.(db.Machine)
-
```
ejj: Useless diff

stitch/bindings.js
```diff
@@ -91,13 +91,24 @@ Deployment.prototype.vet = function() {
...
             var otherLabel = plcm.otherLabel;
             if (otherLabel !== undefined && !labelMap[otherLabel]) {
                 throw service.name + " has a placement in terms of an " +
                     "undeployed service: " + otherLabel;
             }
         });
+
+        if (hasFloatingIp && service.incomingPublic.length
+            && service.containers.length > 1) {
+            throw service.name + " has a floating IP and multiple containers. " +
```
ejj: File a trello card on this, or is it fixed in this series?

tejasmanohar: I think there's a card about reworking placement rules to be in terms of containers, not labels. Will note there.

tejasmanohar:
When I rebased again, cluster/provider/* went away so guess it was incomplete before.

stitch/bindings.js
tejasmanohar: Opened card

engine/engine.go
tejasmanohar: removed

engine/engine.go
tejasmanohar: yep, seems possible but noted as low-priority. should we do that in this iteration? i'd add set/unsetFloatingIP methods to cluster.provider{}

tejasmanohar:
Ready for next round of review @ejj (maybe @kklin too since the last commit is a large refactor). Addressed https://github.com/NetSys/quilt/pull/587#discussion_r90688737.

cluster/cluster.go
```diff
@@ -36,6 +37,23 @@ type cluster struct {
 	providers map[db.Provider]provider
 }
 
+type syncSet struct {
```
tejasmanohar: some of this probably "should" have been in its own series... but it was really hard to read the code when adding even more arguments and return values (especially in tests)

cluster/provider/provider.go
ejj: ping ...

tejasmanohar:
Rebased but for some reason, when removing an IP, the cluster now removes it properly... and then tries to re-add it so one of the tests started failing

```
time="2016-12-17T17:48:51-08:00" level=info msg="Attempt to remove floating IPs machines." count=1
time="2016-12-17T17:48:51-08:00" level=info msg="Successfully removed floating IPs from machines"
time="2016-12-17T17:48:51-08:00" level=info msg="db.Machine:\n\tMachine-2{Master, FakeAmazon  m4.large, 1, FloatingIP=another.ip}\n\tMachine-4{Master, FakeVagrant  vagrant.large, 1}\n\tMachine-5{Master, FakeAmazon  m4.large, 3, FloatingIP=ip}"
time="2016-12-17T17:48:51-08:00" level=info msg="db.Machine:\n\tMachine-2{Master, FakeAmazon  m4.large, 1, FloatingIP=another.ip}\n\tMachine-4{Master, FakeVagrant  vagrant.large, 1}\n\tMachine-5{Master, FakeAmazon  m4.large, 3, FloatingIP=ip}"
time="2016-12-17T17:48:51-08:00" level=error msg="Failed to get ACLs" error="no ACL rows found"
time="2016-12-17T17:48:51-08:00" level=info msg="Attempt to assign floating IPs machines." count=1
time="2016-12-17T17:48:51-08:00" level=info msg="Successfully assigned floating IPs to machines"
```

I think it has something to do with join being called twice in a loop. Will keep looking but wanted to post an update here in case someone has any ideas what could've changed to cause this

kklin:
@tejasmanohar I took a look, looks like it's because of this conditional https://github.com/NetSys/quilt/pull/587/files#diff-f2792de55ee0d606be19884f969036cdR261. We don't update the DB to reflect that the floating IP was removed.

tejasmanohar:
derp :( - that makes sense.

tejasmanohar:
@LuiseV Ready for review!

specs/example-ip.js
```diff
@@ -0,0 +1,24 @@
...
+
+var baseMachine = new Machine({
+  provider: "Amazon",
+  size: "m4.large",
+  region: "us-west-2",
+  sshKeys: githubKeys("ejj")
+});
+deployment.deploy(baseMachine.asMaster());
+
+baseMachine.floatingIp = floatingIp;
```
LuiseV: We should probably also think about how we'd create multiple machines with floating IPs. If we `replicate` the machines, the current solution would involve some ugly array indexing, I think... There should be a better solution, but I don't think you should care about that now.

cluster/amazon/amazon.go
```diff
@@ -117,14 +126,28 @@ func (clst Cluster) Boot(bootSet []machine.Machine) error {
 			awsIDs = append(awsIDs, awsID{
 				spotID: *request.SpotInstanceRequestId,
 				region: br.region})
+
+			if br.floatingIP != "" {
+				floatingIPs = append(floatingIPs, floatingIP{
+					awsID: awsID{
+						spotID: *request.SpotInstanceRequestId,
```
tejasmanohar: not sure i follow... https://play.golang.org/p/UcJ_gtIbGp

cluster/vagrant/vagrant.go
```diff
@@ -1,6 +1,7 @@
 package vagrant
```
tejasmanohar: IMO, no. e58bc39 introduces new methods on the `provider` interface to make "Floating IPs join()'able in-place", and this just makes Vagrant _satisfy_ the interface. Wouldn't want an in-between commit with compile errors on master :P

cluster/amazon/amazon.go
```diff
@@ -314,6 +352,113 @@ func (clst Cluster) getClient(region string) client {
...
+			}
+
+			for _, inst := range instances {
+				if inst == nil ||
+					*inst.State.Name != ec2.InstanceStateNameRunning {
+					return false
+				}
+			}
+			return true
+		}, 10*time.Second, 15*time.Minute)
```
tejasmanohar: kept it ~equal to what the timeout was for other stuff in this package, but I agree. what do you think is more reasonable? idk if things slow down as you boot more machines in parallel in one account etc (though I wouldn't expect that)

cluster/provider/provider.go
ejj: Right, I'm saying the cluster machines moved to cluster/machine/machine.go.

@tejas, are you going to fix the rebasing error?

cluster/provider/provider.go
```diff
@@ -0,0 +1,121 @@
...
+	case "Vagrant":
+	default:
+		panic(fmt.Sprintf("Unknown Cloud Provider: %s", m.Provider))
+	}
+
+	m.Region = region
+	return m
+}
+
+func resolveString(ptr *string) string {
```
tejasmanohar: (ignoring)

cluster/provider/provider.go
```diff
@@ -0,0 +1,121 @@
...
+// they don't particularly care where a system is placed.
+func DefaultRegion(m db.Machine) db.Machine {
+	if m.Region != "" {
+		return m
+	}
+
+	region := ""
+	switch m.Provider {
+	case "Amazon":
+		region = "us-west-1"
```
tejasmanohar: (ignoring)

cluster/provider/provider.go
```diff
@@ -1,121 +0,0 @@
-package provider
```
tejasmanohar: (ignoring)

cluster/amazon/amazon.go
```diff
@@ -314,6 +352,113 @@ func (clst Cluster) getClient(region string) client {
 	return clst.clients[region]
 }
 
+func (clst Cluster) associateFloatingIPs(requests []floatingIP) error {
+	if len(requests) == 0 {
+		return nil
+	}
+
+	// toAssociate is a map of region to spot ID to floating IP.
+	toAssociate := map[string]map[string]floatingIP{}
```
tejasmanohar: (outdated commit)

specs/example-ip.js
tejasmanohar: nope, no reason!

and yes, I agree. will think on that

specs/example-ip.js
```diff
@@ -0,0 +1,24 @@
...
+publicInternet.connect(80, webTier);
+
+webTier.place(new MachineRule(false, {floatingIp: floatingIp}));
+deployment.deploy(webTier);
+
+var baseMachine = new Machine({
+  provider: "Amazon",
+  size: "m4.large",
+  region: "us-west-2",
+  sshKeys: githubKeys("ejj")
```
tejasmanohar: sure!

tejasmanohar:
Updated @LuiseV

cluster/provider/provider.go
tejasmanohar: weird, yes

cluster/provider/provider.go
tejasmanohar: fixed rebase error

cluster/amazon/amazon.go
```diff
@@ -210,6 +233,17 @@ func (clst Cluster) List() ([]machine.Machine, error) {
...
 		}
 
+		addrResp, err := client.DescribeAddresses(nil)
+		if err != nil {
+			return nil, err
+		}
+		ipMap := map[string]*ec2.Address{}
+		for _, ip := range addrResp.Addresses {
+			if ip.InstanceId != nil {
+				ipMap[*ip.InstanceId] = ip
```
LuiseV: Oh, wait, my bad. You are dereferencing the member, not the struct.

cluster/amazon/amazon.go
LuiseV: Yeah, I'd say something between 5 and 10 minutes.

LuiseV:
Ready for @ejj

specs/example-ip.js
```diff
@@ -0,0 +1,23 @@
...
+webTier.place(new MachineRule(false, {floatingIp: floatingIp}));
+deployment.deploy(webTier);
+
+var baseMachine = new Machine({
+  provider: "Amazon",
+  size: "m4.large",
+  region: "us-west-2",
+  // AWS: http://docs.aws.amazon.com/cli/latest/reference/ec2/allocate-address.html
+  // Google: https://cloud.google.com/compute/docs/configure-instance-ip-addresses#reserve_new_s
+  floatingIp: "xxx.xxx.xxx.xxx (CHANGE ME)",
```
kklin: I don't think setting it here works:  this will cause the master and worker to both get the same floating IP right?

kklin: Also, looks like the Google link is cutoff.

LuiseV: Good point about setting the floating IP. Sorry for missing that.
I guess you could do something like
```javascript
var worker = baseMachine.asWorker();
worker.floatingIp = "xxx.xxx.xxx.xxx (CHANGE ME)";
deployment.deploy(worker);```
Not sure if you saw the above comment on this, @kklin, but we should probably think about a nice way to assign floating IPs to many different VMs. I imagine it won't be super clean with `replicate()`.

tejasmanohar: eek, yeah- that's why I set it separately earlier :( (changed back)

kklin: @tejasmanohar I think it deserves a comment too. Code that looks buggy is basically as bad as buggy code IMO :)

cluster/amazon/amazon.go
tejasmanohar: ah, i see where you got confused

ejj:
Is there a rebase error in this patch `util: Refactor quilt-tester and awsSpot to use util.WaitFor`?  I see you delete the code and tests, but I don't see you add it back

PR #591, opened 2016-11-15
quiltctl: Lookup `run` argument in QUILT_PATH

kklin:
This patch allows users to pass an import path to the run command. This
is nice because it matches the semantics of the `get` command:

```
ethan:quilt-demo|master$ quilt get github.com/ejj/quilt-demo/mean
INFO [Nov 12 17:55:18.008] Updating github.com/ejj/quilt-demo in /Users/ethan/.quilt/github.com/ejj/quilt-demo
Successfully installed import.
ethan:quilt-demo|master$ quilt run github.com/ejj/quilt-demo/mean
Successfully started run.
```


tejasmanohar:
Feels weird to me, but maybe that's just because neither Go nor Node has this (AFAIK).

kklin:
@tejasmanohar I don't feel strongly about it, but it was a feature request by @ejj so you guys can discuss.

tejasmanohar:
Oh ok. I also don't tbh. +0.

kklin:
Will assign reviewer once build starts passing again.

ejj:
Btw, my interpretation is that go more-or-less does let you do this.  The syntax for `go get`ing a package is the same as the import syntax for that package.

PR #593, opened 2016-11-17
api: Don't allow switching namespaces

kklin:
This patch rejects requests to deploy specs that would switch namespaces
because we currently have a bug that could cause machines in the old
namespace to be destroyed.


api/server/server.go
```diff
@@ -101,6 +102,18 @@ func (s server) Run(cts context.Context, runReq *pb.RunRequest) (*pb.RunReply, e
 			"machines", ip.MaxMinionCount)
 	}
 
+	var clusters []db.Cluster
+	s.dbConn.Transact(func(view db.Database) error {
+		clusters = view.SelectFromCluster(nil)
+		return nil
+	})
+	if len(clusters) != 0 && clusters[0].Namespace != stitch.QueryNamespace() {
```
LuiseV: Maybe it is clearer to use `GetCluster()` since we only check `clusters[0]` anyways?

kklin: That function throws an error when there's no clusters, so I think it's more straightforward to just test it ourselves rather than inspect the error returned.

LuiseV: Fine with me!

LuiseV:
Ready for @ejj

PR #594, opened 2016-11-18
docker: Stop using docker exec

ejj:
We're currently dealing with a docker bug in which the docker daemon
deadlocks occasionally.  While looking through stack traces we've
noticed that the docker daemon appears to be processing exec calls.
Additionally, a month or two ago we noticed that the minion was making
an exec call at the time of the deadlock.  This patch deals with the
issue by removing all exec calls from the minion.

A happy side effect of the approach this patch takes, is slightly
cleaner code.  All of the exec calls before this patch were to access
OVS utilities that are now bundled with the minion container directly.

### DIFF
--- a/minion/network/worker.go
+++ b/minion/network/worker.go
@@ -19,7 +19,6 @@ import (
 	"github.com/NetSys/quilt/join"
 	"github.com/NetSys/quilt/minion/docker"
 	"github.com/NetSys/quilt/minion/ovsdb"
-	"github.com/NetSys/quilt/minion/supervisor"
 	"github.com/NetSys/quilt/stitch"
 	"github.com/NetSys/quilt/util"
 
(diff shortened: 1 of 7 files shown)


minion/network/worker.go
```diff
@@ -1478,48 +1472,33 @@ func deleteRoute(namespace string, r route) error {
...
-	return nil
-}
 
-func deleteOFRules(dk docker.Client, flows []interface{}) error {
-	flowCommands := []string{}
 	for _, f := range flows {
 		flow := f.(OFRule)
-		flowCommands = append(flowCommands, fmt.Sprintf("%s,%s",
-			flow.table, flow.match))
+		stdin.Write([]byte(fmt.Sprintf("%s,%s,actions=%s\n",
```
kklin: Nice!

ejj:
I'll do it as a separate patch, all those other constants should be unexported too.

Thanks!

PR #595, opened 2016-11-18
docs: Remove network dependency in readme_test

kklin:
The README has code that resolves GitHub keys.


docs/readme_test.go
```diff
@@ -112,6 +115,12 @@ func TestReadme(t *testing.T) {
 }
 
 func checkConfig(content string, quiltPath string) error {
+	stitch.HTTPGet = func(url string) (*http.Response, error) {
+		resp := http.Response{
+			Body: ioutil.NopCloser(bytes.NewBufferString("")),
+		}
+		return &resp, nil
+	}
```
mmussomele: restore `stitch.HTTPGet` after?

mmussomele:
Other than the one comment, looks good.

docs/readme_test.go
kklin: Eh, tests run in isolation, I think it'd just make the test slightly harder to follow.

kklin: Actually, Matt swayed me otherwise.

PR #597, opened 2016-11-18
scripts: Remove build-push script

ejj:
It's not really doing anything useful.


ejj:
I don't think anyone necessarily has to review this.  Just pushing to test travis.

PR #599, opened 2016-11-19
[RFC] provider: Support for DigitalOcean

secant:
Still need to:
- [ ] support floating IPs
- [ ] get test coverage


tejasmanohar:
We should probably use Digital Ocean's Block Storage (alike AWS EBS).

secant:
@tejasmanohar [That is what I'm using](https://github.com/NetSys/quilt/pull/599/commits/16efcaf2dc96569332b3a17a8a4c6b127dd71336#diff-4908ea42fa76029aebdad7bf617013bbR197).

cluster/provider/digitalocean.go
```diff
@@ -0,0 +1,287 @@
...
+		return nil
+	}
+	if len(bootSet) <= 0 {
+		return nil
+	}
+
+	done := make(chan error, len(bootSet))
+	errors := []error{}
+	for _, m := range bootSet {
+		go clst.createAndAttach(m, done)
```
kklin: I think it makes more sense for `createAndAttach` to not know about the parallelism.

```
go func(m machine) {
    defer wg.Done()
    if err := clst.createAndAttach(m) ; err != nil {
        errChan <- err
    }
}
```

BTW, I think it'd be reasonable to just return any of the multiple errors. So the error channel would be of size 1, and we'd try a non-blocking write if there's an error.

cluster/provider/digitalocean.go
```diff
@@ -0,0 +1,287 @@
...
+		return
+	}
+
+	_, _, err = clst.client.StorageActions.Attach(v.ID, d.ID)
+	if err != nil {
+		done <- err
+		return
+	}
+
+	for len(d.VolumeIDs) == 0 {
```
kklin: What if this loops forever?

kklin: Same for the above Status loop.

I'd write some sort of waiter function that tries every `x` seconds, and handles timing out.

cluster/provider/digitalocean.go
```diff
@@ -0,0 +1,287 @@
...
+func (clst doCluster) Stop(machines []Machine) error {
+	if len(machines) <= 0 {
+		return nil
+	}
+
+	done := make(chan error, len(machines))
+	errors := []error{}
+
+	for _, m := range machines {
+		go clst.deleteAndWait(m.ID, done)
```
kklin: Same comment here re parallelism.

cluster/provider/digitalocean.go
```diff
@@ -0,0 +1,287 @@
...
+		return
+	}
+	d, _, _ := clst.client.Droplets.Get(id)
+	volIds := d.VolumeIDs
+	_, err = clst.client.Droplets.Delete(id)
+	if err != nil {
+		done <- err
+		return
+	}
+	for d != nil {
```
kklin: Same here re waiting.

cluster/provider/digitalocean.go
kklin: Also easier to unit test this way.

cluster/provider/digitalocean.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+
+		opt.Page = page + 1
+	}
+	return machines, nil
+}
+
+// Boot will boot every machine in a goroutine, and wait for the machines to come up.
+func (clst doCluster) Boot(bootSet []Machine) error {
+	if first {
```
ejj: what's this all about?  Seems weird

cluster/provider/digitalocean.go
kklin: BTW, @tejasmanohar added a generic waiter in his PR.

cluster/provider/digitalocean.go
secant: leftovers from when `List` wasn't working and it would boot many many times

PR #600, opened 2016-11-19
scheduler: Fix bug in scheduler

LuiseV:
The scheduler deletes containers as it loops over them. This could before cause
incorrect placements and going out of bounds. With this patch, we iterate over
the list of containers in reverse order to avoid these issues.


LuiseV:
Cool, I agree that the commit message was pretty vague. Do you like the new one better?

Also, I'll use `github.com/stretchr/testify` next time!

kklin:
Yep, looks good. Ready for @ejj

LuiseV:
@ejj , as you suggested, I updated to keep a separate array of validly placed containers and assigning it to the minion at the end of the loop. I kind of don't like passing around that extra argument, though. Using a set of containers instead of an array would simplify the implementation of the check/delete. Might be better?

LuiseV:
Actually, I take that back. The set would presumably make testing more annoying, since scheduling becomes non-deterministic

PR #601, opened 2016-11-19
Table locks

mmussomele:
This patch reduces the lock granularity on the db to the table level, allowing transactions that operate on independent table sets to run concurrently.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -22,19 +22,17 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
-type server struct {
-	dbConn db.Conn
-}
+type server struct{}
 
 // Run accepts incoming `quiltctl` connections and responds to them.
-func Run(conn db.Conn, listenAddr string) error {
+func Run(listenAddr string) error {
 	proto, addr, err := api.ParseListenAddress(listenAddr)
 	if err != nil {
 		return err
 	}
 
 	var sock net.Listener
-	apiServer := server{conn}
+	apiServer := server{}
 	for {
 		sock, err = net.Listen(proto, addr)
 
@@ -63,9 +61,21 @@ func Run(conn db.Conn, listenAddr string) error {
 	return nil
 }
 
-func (s server) Query(cts context.Context, query *pb.DBQuery) (*pb.QueryReply, error) {
+func (s server) Query(cts context.Context, query *pb.DBQuery) (reply *pb.QueryReply,
+	err error) {
+	defer func() {
+		// If the queried table is not valid, db.Open will panic.
+		if r := recover(); r != nil {
+			reply = nil
+			err = fmt.Errorf("unrecognized table: %s", query.Table)
+		}
+	}()
+
 	var rows interface{}
-	err := s.dbConn.Transact(func(view db.Database) error {
+	// Only one table type will ever be accessed, so we can just open a connection
+	// to that table.
+	conn := db.Open(db.TableType(query.Table))
+	conn.Transact(func(view db.Database) error {
 		switch db.TableType(query.Table) {
 		case db.MachineTable:
 			rows = view.SelectFromMachine(nil)
@@ -80,13 +90,10 @@ func (s server) Query(cts context.Context, query *pb.DBQuery) (*pb.QueryReply, e
 		case db.ClusterTable:
 			rows = view.SelectFromCluster(nil)
 		default:
-			return fmt.Errorf("unrecognized table: %s", query.Table)
+			panic("not reached")
 		}
 		return nil
 	})
-	if err != nil {
-		return nil, err
-	}
 
 	json, err := json.Marshal(rows)
 	if err != nil {
@@ -109,7 +116,7 @@ func (s server) Deploy(cts context.Context, deployReq *pb.DeployRequest) (
 			"machines", ipdef.MaxMinionCount)
 	}
 
-	err = s.dbConn.Transact(func(view db.Database) error {
+	err = db.Open(db.ClusterTable).Transact(func(view db.Database) error {
 		cluster, err := view.GetCluster()
 		if err != nil {
 			cluster = view.InsertCluster()
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -40,11 +40,10 @@ var sleep = time.Sleep
 
 // Run continually checks 'conn' for cluster changes and recreates the cluster as
 // needed.
-func Run(conn db.Conn) {
+func Run() {
 	var clst *cluster
-	clusterConn := conn.Restrict(db.ClusterTable, db.MachineTable, db.ACLTable)
-	for range clusterConn.TriggerTick(30).C {
-		clst = updateCluster(conn, clst)
+	for range db.TriggerTickOn(30, db.ClusterTable, db.MachineTable, db.ACLTable).C {
+		clst = updateCluster(clst)
 
 		// Somewhat of a crude rate-limit of once every five seconds to avoid
 		// stressing out the cloud providers with too many API calls.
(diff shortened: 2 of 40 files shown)


cluster/cluster.go
```diff
@@ -53,10 +54,10 @@ func Run(conn db.Conn) {
 }
 
 func newCluster(conn db.Conn, namespace string) *cluster {
+	triggerConn := conn.Restrict(db.ClusterTable, db.MachineTable)
```
kklin: ACLTable was removed?

minion/etcd/elector.go
```diff
@@ -15,6 +15,7 @@ const leaderKey = dir + "/leader"
 
 // Run blocks implementing leader election.
 func runElection(conn db.Conn, store Store) {
+	conn = conn.Restrict(db.EtcdTable, db.MinionTable)
```
kklin: Do we gain anything by restricting here? My gut feeling is that we should restrict as late as possible to avoid inadvertent panics.

minion/supervisor/supervisor.go
```diff
@@ -59,7 +59,7 @@ type supervisor struct {
 
 // Run blocks implementing the supervisor module.
 func Run(conn db.Conn, dk docker.Client) {
-	sv := supervisor{conn: conn, dk: dk}
+	sv := supervisor{conn: conn.Restrict(db.MinionTable, db.EtcdTable), dk: dk}
```
kklin: Same here re `Restrict`ing early

db/db.go
```diff
@@ -41,16 +40,38 @@ func New() Conn {
...
-	cn := Conn{db: db, lock: &sync.Mutex{}}
+	cn := Conn{db: db}
 	cn.runLogger()
 	return cn
 }
 
+// Restrict creates a new connection connected to the same database, but with restricted
+// access to only the given tables.
+// While this function returns a new Conn, the Conn it uses should be used in as limited
+// a scope as possible. The reaon for this is that if the newly restricted Conn is
```
kklin: "reason"

db/db.go
```diff
@@ -41,16 +40,38 @@ func New() Conn {
...
+// passed to another function, that function has no way of knowing what tables the Conn
+// has access to. The suggested usage is as follows:
+//
+//	conn := db.New()
+//	/* Potentially other code */
+//	conn.Restrict(db.MachineTable, db.ContainerTable).Transact(...)
+func (cn Conn) Restrict(tables ...TableType) Conn {
+	// Same database data, just a subset of the tables
```
kklin: Comments should be sentence-like (add a period, and make it less fragment-y).

db/db.go
```diff
@@ -41,16 +40,38 @@ func New() Conn {
...
+	db := Database{make(map[TableType]*table), cn.db.idAlloc}
+	for _, t := range tables {
+		db.tables[t] = cn.db.accessTable(t)
+	}
+
+	conn := Conn{db: db}
+	return conn
```
kklin: Combine these two lines?

db/db.go
```diff
@@ -148,6 +192,29 @@ func (db Database) nextID() int {
 	return *db.idAlloc
```
kklin: Do we need to lock this now? We could have an `InsertFoo` happen concurrently on two different tables right?

db/db.go
```diff
@@ -109,21 +131,43 @@ func (cn Conn) TriggerTick(seconds int, tt ...TableType) Trigger {
...
+	}
+}
+
 // Stop a running trigger thus allowing resources to be deallocated.
 func (t Trigger) Stop() {
 	close(t.stop)
 }
 
 func (db Database) insert(r row) {
-	table := db.tables[getTableType(r)]
```
kklin: Have you thought through whether these functions are thread-safe? A comment explaining why they're safe would be nice.

db/db_test.go
```diff
@@ -124,6 +124,179 @@ func TestMachineString(t *testing.T) {
...
+	conn := New()
+	conn.Transact(func(view Database) error {
+		view.InsertEtcd()
+		view.InsertLabel()
+		view.InsertMinion()
+		view.InsertMachine()
+		view.InsertCluster()
+		view.InsertPlacement()
+		view.InsertContainer()
+		view.InsertConnection()
```
kklin: Missing the `InsertACL`

db/db_test.go
```diff
@@ -124,6 +124,179 @@ func TestMachineString(t *testing.T) {
...
+		}
+	}
+}
+
+// returns false when the transactions deadlock
+func checkRestrictedNoDeadlock() bool {
+	transactOneStart := make(chan struct{})
+	transactTwoDone := make(chan struct{})
+	done := make(chan struct{})
+	doneRoutines := make(chan struct{}, 2)
```
kklin: I _think_ you could just have this be `make(chan struct{})` and `close` it to notify all threads.

db/db_test.go
```diff
@@ -124,6 +124,179 @@ func TestMachineString(t *testing.T) {
...
+	// condition
+	for i := 0; i < 10; i++ {
+		if !checkRestrictedNoDeadlock() {
+			t.Fatal("Transacts deadlocked")
+		}
+	}
+}
+
+// returns false when the transactions deadlock
+func checkRestrictedNoDeadlock() bool {
```
kklin: I don't really get how this test works. Aren't you `restrict`ing on completely different tables?

It'd be nice if we could randomly generate this test.

api/server/server.go
```diff
@@ -65,7 +65,9 @@ func Run(conn db.Conn, listenAddr string) error {
 
 func (s server) Query(cts context.Context, query *pb.DBQuery) (*pb.QueryReply, error) {
 	var rows interface{}
-	err := s.dbConn.Transact(func(view db.Database) error {
+	err := s.dbConn.Restrict(db.MachineTable, db.EtcdTable,
```
kklin: Maybe dynamically `Restrict` here based on `query.Table`? We only ever use one of the tables in a single transaction.

db/db.go
mmussomele: Yea they are, since each conn has a separate db map, so they are reading from different maps, and they can only modify and read tables they have a lock on. Although this makes me realize the incrementing of the DB id isn't thread safe, I'll fix that up

db/db.go
mmussomele: Ah just saw this comment. Yea I realized this too, it's fixed now.

db/db_test.go
mmussomele: Yea, it tests that two restricted connections can run concurrently without deadlocking when their table sets have no overlap. And yea, I could randomize the table sets.

db/db_test.go
mmussomele: yup good point

minion/etcd/elector.go
mmussomele: Good point, no need to do it there

api/server/server.go
mmussomele: 👍 ended up making the code simpler

mmussomele:
@kklin ready for another look, or to pass on

kklin:
Just a note, once this gets merged, we should remember to update PR's to use restricted connections.

db/db_test.go
kklin: Why are we only testing if there's no overlap? Shouldn't we avoid deadlock even if there is overlap?

db/db.go
```diff
@@ -30,27 +32,49 @@ type row interface {
...
+// a scope as possible. The reason for this is that if the newly restricted Conn is
+// passed to another function, that function has no way of knowing what tables the Conn
+// has access to. The suggested usage is as follows:
+//
+//	conn := db.New()
+//	/* Potentially other code */
+//	conn.Restrict(db.MachineTable, db.ContainerTable).Transact(...)
+func (cn Conn) Restrict(tables ...TableType) Conn {
+	// This Conn has the same database data, just a subset of the tables.
+	db := Database{make(map[TableType]*table), cn.db.idAlloc, cn.db.lock}
```
kklin: Could there be some sort of idAlloc object that contains the lock? Seems slightly smelly that we have to remember to copy both here.

db/db_test.go
mmussomele: Ah ok, I see what you mean. The test is poorly named I suppose. What this tests is that two transacts on different tables sets can run concurrently instead of sequentially. The test in particular is constructed to deadlock iff the transactions run sequentially. I'll figure out a better name.

db/db.go
mmussomele: good idea, I'll add one

mmussomele:
All comments corrected

ejj:
In the first commit message you say you want to avoid "unneeded deadlocks".  I don't like that phrase, a deadlock is a circular lock dependency that's irrecoverable (something that doesn't happen for us).  Maybe say "unneeded blocking" instead?

ejj:
So here's my review:

I think the first two patches are ready ... I think you should resubmit the PR without the last patch, so I can go ahead and merge them and we get them on master.

The last patch I think is a good idea, but I think you should make at tweak.  I actually think you should support *both* approaches.  I.E. you have a New() function which creates and returns a brand new database and an Open() function which returns a default database stored as a private variable in the db package.  That way you don't have to change the unit tests as drastically (also it seems clean)

mmussomele:
After discussion offline, the series has changed a lot and has been re-submitted here: https://github.com/NetSys/quilt/pull/657

PR #608, opened 2016-11-20
quiltctl: Explictly request PTY allocation in `exec`

kklin:
Before, the `exec` subcommand would failed when run in contexts that
couldn't allocate a pseudo-terminal:
```
# This failed unless we enable psudo-terminal allocation with `-t`.
% ssh localhost quilt exec 1 cat /etc/hosts
```


tejasmanohar:
or, should we just add a `-t` flag for PTY like docker exec does?

kklin:
@tejasmanohar I think setting it automatically is easier for users.. although I could be missing a reason why Docker doesn't just do it automatically too.

tejasmanohar:
@kklin I think it's just a precaution. Commands you run via `<docker/quilt> exec` may act differently when run in a pseudo-terminal or not (e.g. different output format).

kklin:
@tejasmanohar Good point, I'll make us use a flag then.

kklin:
@tejasmanohar ready for review

Makefile
```diff
@@ -32,7 +32,7 @@ COV_SKIP= /minion/pb /minion/pprofile /api/pb /constants /scripts /quilt-tester
 		  /quilt-tester/tests/100-basic /quilt-tester/tests/100-basic/check_docker.go \
 		  /quilt-tester/tests/100-basic/check_logs.go \
 		  /quilt-tester/tests/spark /quilt-tester/tests/spark/check_spark.go \
-		  /quiltctl/testutils /cluster/provider/mocks
+		  /quiltctl/testutils /cluster/provider/mocks /api/client/mock
```
tejasmanohar: may as well name the folder "mocks" everywhere to be consistent

api/client/getter/getter.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+// New returns an implementation of the Getter interface.
+func New() client.Getter {
+	return clientGetterImpl{addrClientGetterImpl{}}
+}
+
+// We create a separate interface for getting a client given an address so that
+// ContainerClient and LeaderClient can be unit tested.
+type addrClientGetter interface {
```
tejasmanohar: not super happy about the concept of a client getter interface but don't have a better drop-in idea rn

tejasmanohar:
(merge conflicts btw)

kklin:
Ready for @ejj review.

tejasmanohar:
Maybe rename PR title

tejasmanohar:
oh, I meant the commit message & PR should probably mention the `-t` flag

kklin:
@tejasmanohar Oh, that makes a lot more sense :P

api/client/getter/getter.go
```diff
@@ -0,0 +1,144 @@
...
+	client.Client, error) {
+
+	machines, err := localClient.QueryMachines()
+	if err != nil {
+		return nil, fmt.Errorf("unable to query machines: %s", err.Error())
+	}
+
+	// Try to figure out the lead minion's IP by asking each of the machines
+	// tracked by the local daemon.
+	for _, m := range machines {
```
ejj: Could we just check the master machines?

api/client/getter/getter.go
```diff
@@ -0,0 +1,144 @@
...
+		log.WithError(err).Debug("Unable to get leader IP.")
+	}
+
+	return nil, errors.New("no leader found")
+}
+
+func (getter clientGetterImpl) ContainerClient(localClient client.Client,
+	stitchID int) (client.Client, error) {
+
+	leaderClient, err := getter.LeaderClient(localClient)
```
ejj: Btw, I think we should be able to ask any master where the containers are scheduled not just the leader.  Don't know if that actually works, but it might make sense to implement it, should be trivial.

api/util/util.go
```diff
@@ -0,0 +1,25 @@
+package util
```
ejj: Meh, I don't much l like util packages, could this go somewhere else?

That said, I don't feel that strongly about it, if you think this is the best place for it, then fine.

quiltctl/command/exec.go
```diff
@@ -73,6 +78,11 @@ func (eCmd *Exec) Parse(args []string) error {
 
 // Run finds the target continer, and executes the given command in it.
 func (eCmd *Exec) Run() int {
+	if eCmd.allocatePTY && !isTerminal() {
+		log.Error("Cannot allocate pseudo-terminal without a terminal.")
```
ejj: No periods in log messages

quiltctl/command/exec.go
```diff
@@ -100,16 +110,27 @@ func (eCmd *Exec) Run() int {
...
+		if err = eCmd.SSHClient.RequestPTY(); err != nil {
+			log.WithError(err).Info("Error requesting pseudo-terminal")
+			return 1
+		}
 	}
 
-	command := fmt.Sprintf("docker exec -it %s %s", container.DockerID, eCmd.command)
+	var flags string
+	if eCmd.allocatePTY {
+		flags = "-it"
```
ejj: Seems a bit suspicious to me that docker treats `i` and `t` separately, but we're combining them is this the right long term strategy, or a temporary thing?

api/client/getter/getter.go
kklin: Just to confirm, just check whether the master machine is the leader?

api/util/util.go
kklin: It's used both by `api/client/getter` and `quiltctl/command`. Not sure where else to put it, not against moving it though.

api/client/getter/getter.go
ejj: Actually I don't think my comment makes sense, nevermind.

api/util/util.go
ejj: fine where it is

api/client/getter/getter.go
kklin: @ejj would that be cleaner if it was in a separate commit? This commit is just moving the old implementation into an interface.

kklin: Hmm now that I look into it more, we only use the `LeaderClient` to query the cluster's container information. So I think this is better done in a separate PR where we get rid of `LeaderClient` altogether. It'd simplify the logic quite a bit. I don't really want to do it in this series though as this is blocking some other patches, and the changes would be done in a separate commit anyways. What do you think?

PR #610, opened 2016-11-21
quilt-tester: Deploy services in `basic` test

kklin:
@ejj it really is too easy to forget to deploy something..


kklin:
@secant That sounds reasonable, could you file a card?

PR #612, opened 2016-11-22
quiltctl: Confirm changes before deploying new specs

kklin:
Before, it was very easy for users to make destructive changes to their
existing cluster by deploying a new spec. For example, the new spec may
have different machine types, which would result in the old machines
being destroyed.

This patch makes `quilt run` display a diff of the deployment object
and obtain user confirmation before actually deploying the proposed
spec.


kklin:
@LuiseV will review as this is really user-facing

ejj:
FWIW, we should write a trello card on making the json diff prettier.  Would be a good starter project for someone

PR #613, opened 2016-11-22
Network integration test

kklin:


quilt-tester/tests/75-network/check_network.go
```diff
@@ -0,0 +1,276 @@
...
+		container.IP: {},
+	}
+	for _, label := range container.Labels {
+		for _, toLabelName := range tester.connectionMap[label] {
+			toLabel := tester.labelMap[toLabelName]
+			for _, ip := range append(toLabel.ContainerIPs, toLabel.IP) {
+				expReachable[ip] = struct{}{}
+			}
+		}
+		// We can ping our label. E.g. 1.yellow.q can ping yellow.q
```
secant: `ping 1.yellow.q`

quilt-tester/tests/75-network/check_network.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	return unreachable, unauthorized
+}
+
+// ping `target` from within container `id` with 3 packets, with a timeout of
+// 1 second for each packet.
+func ping(id int, target string) (string, bool) {
+	outBytes, err := exec.Command(
+		"quilt", "exec", strconv.Itoa(id), "ping", "-c", "3", "-W", "1", target).
```
secant: move the `.` to the newline

quilt-tester/tests/75-network/check_network.go
```diff
@@ -0,0 +1,276 @@
...
+
+	return pingResults
+}
+
+func (tester networkTester) test(container db.Container) (
+	unreachable []string, unauthorized []string) {
+
+	// We should be able to ping ourselves.
+	expReachable := map[string]struct{}{
+		container.IP: {},
```
secant: `struct{}{}`?

Makefile
```diff
@@ -32,6 +32,7 @@ COV_SKIP= /minion/pb /minion/pprofile /api/pb /constants /scripts /quilt-tester
 		  /quilt-tester/tests/100-basic /quilt-tester/tests/100-basic/check_docker.go \
 		  /quilt-tester/tests/100-basic/check_logs.go \
 		  /quilt-tester/tests/spark /quilt-tester/tests/spark/check_spark.go \
+		  /quilt-tester/tests/75-network /quilt-tester/tests/75-network/check_network.go \
```
kklin: It lets us set the order that tests will run in. 75 will make the network test run after spark (which defaults to 50), and before `basic` (which checks for any errors in the logs, and so should be run last)

quilt-tester/tests/75-network/check_network.go
kklin: Don't really see why, I think it looks cleaner this way.

quilt-tester/tests/75-network/check_network.go
kklin: It's a compiler error if I move it

quilt-tester/tests/75-network/check_network.go
kklin: It actually should be `yellow.q`. What I meant is that containers can ping their overarching label, but they can't ping the other containers in that label. Updated the comment to make the distinction clearer.

quilt-tester/tests/75-network/check_network.go
```diff
@@ -0,0 +1,276 @@
...
+		log.WithError(err).Fatal("FAILED, couldn't get leader client")
+	}
+
+	tester, err := newNetworkTester(leader)
+	if err != nil {
+		log.WithError(err).Fatal("FAILED, couldn't initialize network tester")
+	}
+
+	containers, err := leader.QueryContainers()
+	if err != nil {
```
ejj: Isn't the right patch for it, but at some point, it would be cool if we had an assert framework for the integration tests similar to the testify assert.  Would be very hard to role our own, and it would simplify the code a lot.  What do you think?

quilt-tester/tests/75-network/check_network.go
```diff
@@ -0,0 +1,276 @@
...
+					unreachable:  unreachable,
+					unauthorized: unauthorized,
+				}
+			}
+		}(testChan)
+	}
+
+	// Feed the worker threads until we've run all the tests.
+	for len(containers) != 0 {
+		for i := len(containers) - 1; i >= 0; i-- {
```
ejj: We've been burned by this sort of fancy slice logic before.  I'd just as soon just build a new slice that omits the elements we'd like to delete.

kklin: Done.

Makefile
```diff
@@ -33,6 +33,7 @@ COV_SKIP= /minion/pb /minion/pprofile /api/pb /constants /scripts /quilt-tester
...
```
ejj: Btw I wonder if it makes sense for the network test to run before the spark test.  I.E. if the network is broken then spark is definitely going to fail, so may as well test that directly first, what do you think?

Shouldn't block this patch at any rate.

kklin: I was thinking that running it after spark gives us a chance to see how the network code deals with changing specs.

ejj: Fair enough, though I think additionally we should test that directly.  I.E. extend the network test to change the spec and re-check a couple of times.

PR #614, opened 2016-11-23
network: Use AddressSets for enforcing ACLs

kklin:
Before, we enforced ACLs by enumerating all the IPs that could
communicate. This led to very long, and redundant match strings. This
patch creates an AddressSet for each label, and then creates ACLs simply in
terms of the label name.

This significantly reduces the memory used for creating ACLs. Before, we
would run out of memory on the Master when booting 10,000 containers on
16 `m4.4xlarge`s.


minion/network/acl.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/NetSys/quilt/db"
+	"github.com/NetSys/quilt/join"
+	"github.com/NetSys/quilt/minion/ovsdb"
+	"github.com/NetSys/quilt/stitch"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+func updateACLs(connections []db.Connection, labels []db.Label) {
+	ovsdbClient, err := ovsdb.Open()
```
ejj: Why not take the ovsdbClient as an argument?  The caller has one laying around.

ejj:
Will merge once travis is happy.

PR #615, opened 2016-11-23
Pretty print `quilt containers`

ejj:
The output of `quilt containers` was pretty darn hard to parse before this patch.

### DIFF
--- a/quiltctl/command/container.go
+++ b/quiltctl/command/container.go
@@ -2,6 +2,11 @@ package command
 
 import (
 	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+	"text/tabwriter"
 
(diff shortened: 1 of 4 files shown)


quiltctl/command/container.go
```diff
@@ -51,17 +56,50 @@ func (cCmd *Container) Run() int {
 		return 1
 	}
 
-	str := containersStr(containers)
-	fmt.Print(str)
+	machines, err := localClient.QueryMachines()
+	if err != nil {
+		log.WithError(err).Error("Unable to query machines")
```
ejj: Nah, I don't think log messages should have periods.  We do it both ways, in the source, but most projects omit them.  Trying to standardize on that.

PR #620, opened 2016-11-26
Refactor the cluster package

ejj:
This patch is almost entirely aesthetic and interface changes.  Doesn't need to be reviewed all that carefully, should just be skimmed.


cluster/google/google.go
```diff
@@ -61,49 +67,46 @@ type gceCluster struct {
...
 	}
 
-	clst.projID = "declarative-infrastructure"
-	clst.ns = namespace
-	clst.imgURL = fmt.Sprintf(
-		"%s/%s",
-		computeBaseURL,
-		"ubuntu-os-cloud/global/images/ubuntu-1604-xenial-v20160921")
+	clst := Cluster{
+		projID:    "declarative-infrastructure",
```
mmussomele: quilt?

cluster/google/google.go
```diff
@@ -61,49 +67,46 @@ type gceCluster struct {
 	id int    // the id of the cluster, used externally
 }
 
-// Create a GCE cluster.
+// New creates a GCE cluster.
 //
 // Clusters are differentiated (namespace) by setting the description and
 // filtering off of that.
-//
-// XXX: A lot of the fields are hardcoded.
```
mmussomele: Does this block? If so, might want to note in the comment.

mmussomele:
Looks good, just two small comments

cluster/google/google.go
ejj: So both of these comments are sort of general improvements that need to be made to the gce provider.  I think we separately need to do a pass over it and clean things up, but I don't really thing this is the right patch for it.

PR #622, opened 2016-11-29
quilt-tester: Preserve machines that failed to initialize

kklin:
Before, the `--preserve-failed` flag only preserved machines when the
tests failed to run -- it still destroyed the machines when the
quilt-tester intialization failed.
Although the proper fix is to refactor how quilt-tester handles test
failures, this patch somewhat-hackily makes test initialization failures
look like test run failures.


ejj:
Eww

PR #626, opened 2016-11-30
minion: Sync `SSHKeys` machine attribute

kklin:
Before, the `authorized_keys` file which controls SSH access could only
be edited at boot time. Thus, if a user forgot to specify their machines
with the right public key before booting, they would have to reboot the
entire cluster in order to get access.

This patch makes the `SSHKeys` attribute of machines automatically sync.
We also still set the authorized keys at machine boot to allow logging
in before the minion container boots.


tejasmanohar:
Nice! Though implemented differently, it makes sense to have something like this for disk space and floating IPs since EBS, ElasticIP, and their counterparts are (a/de)ttachable _someday_ 🌈.

minion/keys_test.go
```diff
@@ -0,0 +1,87 @@
...
+		view.Commit(m)
+		return nil
+	})
+	err = runOnce(conn)
+	assert.EqualError(t, err, "open /home/quilt/.ssh/authorized_keys: "+
+		"file does not exist")
+
+	fs.Create(authorizedKeysFile)
+	err = runOnce(conn)
+	assert.EqualError(t, err, "operation not permitted")
```
secant: Just wondering why reading wouldn't be permitted since `Create` makes a file with mode `666`

kklin: It's failing to write because we make the `AppFs` a `ReadOnlyFs`

secant: Got it

ejj:
LGTM, just needs to be rebased.

kklin:
@ejj Rebased.

PR #627, opened 2016-12-01
Tests

ejj:
I had some time to kill where I wanted to write code but didn't feel like thinking that hard.  This pull requests converts several of our test packages to the testify/assert package.  The converted test packages were those that were bringing down our go report card score -- once this patch is in we should move up in the rankings significantly.  Not that it matters . . .

### DIFF
--- a/cluster/foreman_test.go
+++ b/cluster/foreman_test.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"testing"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/NetSys/quilt/db"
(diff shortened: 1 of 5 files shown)


cluster/foreman_test.go
```diff
@@ -263,13 +233,11 @@ func TestConfigConsistency(t *testing.T) {
 	newfm.conn.Transact(func(view db.Database) error {
 		machines := view.SelectFromMachine(nil)
 		for _, m := range machines {
-			if m.PublicIP == "1.1.1.1" && m.Role != db.Master {
-				t.Errorf("db Master had role %v, expected %v", m.Role,
-					db.Master)
+			if m.PublicIP == "1.1.1.1" {
+				assert.Equal(t, db.Role(db.Master), m.Role)
 			}
 			if m.PublicIP == "2.2.2.2" && m.Role != db.Worker {
```
ejj: good catch

PR #628, opened 2016-12-01
ovsdb: Break up ovsdb mock apply method

ejj:
Simplifies things a bit, main reason is the report card though.

### DIFF
--- a/minion/ovsdb/ovsdb_mock.go
+++ b/minion/ovsdb/ovsdb_mock.go
@@ -77,65 +77,79 @@ type fakeMutation struct {
 
 // Applies mutation on a row, and returns number of mutations applied.
 func (mutation fakeMutation) apply(row *fakeRow) int {
-	mutationCount := 0
-	column := (*row)[mutation.column].([]interface{})
-
 	switch mutationVal := mutation.value.(type) {
 	case *ovs.OvsSet:
-		data := column[1].([]interface{})
-
-		uuidGeneric := mutationVal.GoSet[0].(ovs.UUID).GoUUID
-		uuid, ok := uuidMap[uuidGeneric]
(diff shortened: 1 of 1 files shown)


minion/ovsdb/ovsdb_mock.go
```diff
@@ -82,60 +82,79 @@ func (mutation fakeMutation) apply(row *fakeRow) int {
...
+		uuid = uuidGeneric
+	}
+	switch mutation.mutator {
+	case "insert":
+		data = append(data, []interface{}{"uuid", uuid})
+		mutationCount++
+	case "delete":
+		for i, e := range data {
+			if e.([]interface{})[1] == uuid {
+				data = append(data[:i], data[i+1:]...)
```
LuiseV: Are the uuids in `data` guaranteed to be unique? If not, this loop isn't safe since we delete elements while looping. If they are unique, maybe `break` here for clarity?

minion/ovsdb/ovsdb_mock.go
```diff
@@ -82,60 +82,79 @@ func (mutation fakeMutation) apply(row *fakeRow) int {
...
-		data := column[1].([]interface{})
+		mutationCount = mutation.applySet(row, mutationVal)
+	case *ovs.OvsMap:
+		mutationCount = mutation.applyMap(row, mutationVal)
+	default:
+		panic("mutation type is not yet supported:" +
+			reflect.TypeOf(mutationVal).String())
+	}
+
+	(*row)[mutation.column] = column
```
kklin: This is redundant because we modify the row from within the `apply` functions right?

ejj:
Addressed comments, will merge once travis is happy

PR #633, opened 2016-12-06
Cluster refactoring

ejj:
The cluster code was pretty darn hard to reason about leading to a series of bugs, particularly when switching namespaces.  This series address two serious issues: First, an overuse of concurrency.  Second, poor handling of namespace changes.


engine/engine.go
```diff
@@ -13,26 +13,30 @@ import (
...
+		conn.Transact(updateTxn)
+	}
 }
 
-func clusterTxn(view db.Database, stitch stitch.Stitch) {
+func updateTxn(view db.Database) error {
 	cluster, err := view.GetCluster()
 	if err != nil {
-		cluster = view.InsertCluster()
+		return nil
```
mmussomele: why not return the error in case it's used by a later patch? Couldn't hurt

cluster/cluster.go
```diff
@@ -44,24 +44,19 @@ var sleep = time.Sleep
 func Run(conn db.Conn) {
```
ejj: Well, it's used in this patch, so it's not like we have this standalone unused function sitting around.  As a general philosophy, I think having many small patches is better than a few bigger ones because they're easier to review.  Glancing over this one I have a lot of confidence that it's correct without having to wade through a bunch of other junk

cluster/misc_test.go
```diff
@@ -59,7 +59,7 @@ func TestNewProviderSuccess(t *testing.T) {
 	// XXX: We should also try db.Google and db.Vagrant, but both of those access the
 	// network and/or the local filesystem.  We could fix it, but it's only two lines
 	// of coverage.
-	newProvider(db.Amazon, "namespace")
+	newProviderImpl(db.Amazon, "namespace")
```
ejj: So this test is actually designed to test newProviderImpl, the mock is for the other tests that use newProvider, but don't care about covering the Impl code.

That said, I don't have a good way to avoid hitting the network, and it's only one line of code, so I deleted the test.  I'll keep the failure case.

cluster/foreman/foreman.go
```diff
@@ -16,6 +16,8 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
+var minions map[string]*minion
```
ejj: Meh, I disagree.  I don't think stateful libraries a problem, adding more types just for the sake of having them complicates the code for no reason.

Btw, I think adding a foreman type would be appropriate if we needed multiple foremans for some reason, but that's not the situation we're in.

cluster/foreman/foreman_test.go
```diff
@@ -244,10 +243,11 @@ func TestConfigConsistency(t *testing.T) {
...
 }
 
-func startTest() (foreman, *clients) {
-	fm := createForeman(db.New())
+func startTest() (db.Conn, *clients) {
+	conn := db.New()
+	minions = map[string]*minion{}
 	clients := &clients{make(map[string]*fakeClient), 0}
-	fm.newClient = func(ip string) (client, error) {
+	newClient = func(ip string) (client, error) {
```
ejj: Why?

ejj: To expand, I know we do that in several places in the code, but I don't really see the point.  The unit tests are never going to use the default newClient implementation as they don't reach out to minions that don't exist.  For something like the random number generate it makes sense because some tests may want the default implementation, but in this case I don't think it's worth it.

cluster/cluster_test.go
```diff
@@ -108,9 +108,8 @@ func newTestCluster() cluster {
 func TestPanicBadProvider(t *testing.T) {
 	temp := allProviders
 	defer func() {
-		if r := recover(); r == nil {
-			t.Error("newCluster did not panic on bad provider")
-		}
+		r := recover()
+		assert.NotNil(t, r)
```
kklin: Found out recently there's an `assert.NoError`

cluster/cluster.go
```diff
@@ -98,39 +99,34 @@ func (clst *cluster) stop() {
...
-		aclRow, _ := view.GetACL()
-		adminACLs = aclRow.Admin
-		appACLs = aclRow.ApplicationPorts
-		return nil
-	})
-	clst.syncACLs(adminACLs, appACLs, machines)
+		if len(jr.boot) == 0 && len(jr.terminate) == 0 {
+			// ACLs must be processed after Quilt learns about what machines
+			// are in the cloud.  If we didn't, inter-machine ACLs could get
+			// removed when the Quilt controller jrtarts, even if there are
```
kklin: "restarts"

cluster/cluster.go
```diff
@@ -176,18 +172,45 @@ func (clst cluster) updateCloud(machines []machine.Machine, boot bool) {
 	}
 }
 
-func (clst cluster) syncMachines() (bootSet, terminateSet []machine.Machine) {
+type joinResult struct {
+	machines []db.Machine
```
kklin: Why do we need these?

ejj: syncACLs needs them unfortunately.  I really think it should be written such that it doesn't, but, it's probably best to do that for the patch where we pull the ACL join logic out of the cloud providers.

PR #637, opened 2016-12-10
cloudcfg: Restart the minion service on failure

mmussomele:
Sometimes the minion fails to initialize because of some mounting
issues with docker. This seems to be a transient error, so simply
trying again when this happens is the simplest fix.


mmussomele:
@ejj ready for you to look at

ejj:
@mmussomele bump ...  What's the plan on this?  We should either move forward with it or close it I think.

kklin:
IIRC @mmussomele found some open Docker issue on this. People seemed generally mystified, and couldn't reproduce it.
I'm fine with this approach. Perhaps we could use a `XXX` comment to remind ourselves to revisit the Docker issue later.

mmussomele:
Yea I did find an long-running issue on this.

PR #639, opened 2016-12-11
quilt login

sirspinach:


quiltctl/command/exec.go
```diff
@@ -28,13 +28,31 @@ type Exec struct {
 
 	SSHClient    ssh.Client
 	clientGetter client.Getter
+	helper       ExecHelper
+}
+
+// ExecHelper shares functionality between quilt commands similar to `quilt exec`, ie
+// quilt commands that execute a command on a remote docker container.
+type ExecHelper interface {
+	Init(ssh.Client, string, client.Getter, string, int, bool) int
```
kklin: I have a patch for refactoring the SSH package that might help, I'll think about it later.

quiltctl/command/login.go
```diff
@@ -0,0 +1,111 @@
...
+
+// InstallFlags sets up parsing for command line flags.
+func (lCmd *Login) InstallFlags(flags *flag.FlagSet) {
+	lCmd.common.InstallFlags(flags)
+
+	flags.StringVar(&lCmd.privKey, "i", "",
+		"the private key to use to connect to the host")
+	flags.Usage = func() {
+		fmt.Println("usage: quilt login [-H=<daemon_host>] [-i=<private_key>] " +
+			"<stitch_id> <command>")
```
kklin: There's no `<command>` right?

quiltctl/command/login.go
```diff
@@ -0,0 +1,111 @@
...
+		fmt.Println("`login` enters a shell in the specified container." +
+			"The container is identified by the stitch ID produced by " +
+			"`quilt containers`.")
+		fmt.Println("For example, to get a shell in container 5 with a " +
+			"specific private key: quilt login -t -i ~/.ssh/quilt 5")
```
kklin: Don't need the `-t`

quiltctl/command/login.go
```diff
@@ -0,0 +1,111 @@
...
+
+// Run logs into a shell on the remote container.
+func (lCmd *Login) Run() int {
+	if exit := lCmd.helper.Init(lCmd.SSHClient, lCmd.privKey, lCmd.clientGetter,
+		lCmd.common.host, lCmd.targetContainer, true); exit != 0 {
+		return exit
+	}
+
+	bestShell := findBestAvailableShell(shells, lCmd.helper)
+	cmd := fmt.Sprintf("%s", bestShell)
```
kklin: Why this `sprintf`?

quiltctl/command/login.go
```diff
@@ -0,0 +1,111 @@
...
+		checkCmd := fmt.Sprintf("/bin/sh -c 'hash %s 2>/dev/null; echo $?'",
+			shell)
+		out, exit := helper.Output(checkCmd)
+
+		if exit == 0 && strings.TrimSpace(out) == "0" {
+			log.Debug(shell, " is installed")
+			return shell
+		}
+	}
+	return "/bin/sh"
```
kklin: What if `/bin/sh` doesn't exist? I'd just add `/bin/sh` to the list of shells to try, and return an error if we reach this point.

quiltctl/command/login.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return 0
+}
+
+// findBestAvailableShell returns the first shell in `shells` that is installed
+// on the remote container associated with `helper`. If none of these shells are
+// installed, findBestAvailableShell returns "/bin/sh".
+func findBestAvailableShell(shells []string, helper ExecHelper) string {
+	for _, shell := range shells {
+		log.Debug("try shell ", shell)
```
kklin: "trying"

quiltctl/command/login.go
```diff
@@ -0,0 +1,111 @@
...
+		checkCmd := fmt.Sprintf("/bin/sh -c 'hash %s 2>/dev/null; echo $?'",
```
kklin: What if `sh` doesn't exist? Can't we just invoke `hash` directly? And why can't we just check the exit code directly?

sirspinach: `hash` is a shell built-in, so it needs to run within a shell.
```bash
shwangster@aetherboard ~/go/src/github.com/NetSys/quilt
 % docker run --rm ubuntu hash
docker: Error response from daemon: invalid header field value "oci runtime error: container_linux.go:247: starting container process caused \"exec: \\\"hash\\\": executable file not found in $PATH\"\n".

shwangster@aetherboard ~/go/src/github.com/NetSys/quilt
 % which hash
hash: shell built-in command
```

kklin: Meh, then I don't think `hash` is the right approach.

sirspinach: Don't you think it's a safe assumption to make that /bin/sh exists on a unix system?

kklin: I'm not sure. If it really is true that all containers have `sh` then a comment at least would be good.

quiltctl/command/login.go
ejj: Wouldn't the most direct approach, btw, be to just try to run /bin/bash, if it fails try the next one, if that fails try the next one?

quiltctl/command/exec.go
sirspinach: alright, i'll take out this interface

quiltctl/command/login.go
ejj: Meh good point, not sure what's best then.  I do think its reasonable to assume that all containers that have /bin/bash have /bin/sh as well, maybe we stick with something like the current approach.

mmussomele: why not do something like `echo $0` or `echo $SHELL`?

sirspinach: Not sure entirely what you mean, @mmussomele. If you're suggesting that we read `$SHELL` on the quilt client to use as a candidate shell, that sounds good to me.

mmussomele: Yea pretty much

kklin: Sounds complicated to me. At the very least, I think that'd be a feature that gets introduced in a later commit. The initial merge should just be trying from a list of shells.

sirspinach:
Removed ExecHelper interface and replaced it with a execHelper function. 
Check shells existence using exit codes rather than stdout.
Check existence of $SHELL, as suggested by @mmussomele.

sirspinach:
I'm waiting until @kklin's SSH refactor before I write tests for login.go, because login.go will probably become a lot shorter then.

kklin:
Ah, I didn't realize you were blocking on that @sirspinach. I'll prioritize it.

sirspinach:
Ready for further review.

quiltctl/ssh/ssh.go
```diff
@@ -2,12 +2,11 @@ package ssh
...
 	// Run runs a command over the SSH connection.
-	Run(string) error
-	// Disconnect closes the SSH connection.
-	Disconnect() error
+	Run(bool, string) error
+	// Close closes the SSH connection.
+	Close() error
 }
+
+// Getter is used to retrieve an Client
```
kklin: "a Client"

quiltctl/command/exec.go
```diff
@@ -118,6 +119,12 @@ func (eCmd *Exec) Run() int {
 	command := strings.Join(
 		[]string{"docker exec", flags, container.DockerID, eCmd.command}, " ")
 	if err = sshClient.Run(eCmd.allocatePTY, command); err != nil {
+		if exitErr, ok := err.(*golangSSH.ExitError); ok {
+			log.WithError(err).Debug(
+				"SSH command returned a nonzero exit code")
+			return exitErr.ExitStatus()
```
kklin: Ah, this is much better.

quiltctl/command/exec.go
```diff
@@ -97,7 +97,6 @@ func (eCmd *Exec) Run() int {
 		log.WithError(err).Error("Error getting container client")
 		return 1
 	}
-	defer containerClient.Close()
```
kklin: Did you mean to delete this line?

quiltctl/command/exec.go
```diff
@@ -112,19 +111,28 @@ func (eCmd *Exec) Run() int {
 	}
 	defer sshClient.Close()
 
+	if exitCode := execHelper(eCmd.command, sshClient, container.DockerID,
```
kklin: Just return the exit code

quiltctl/command/exec.go
```diff
@@ -112,19 +111,28 @@ func (eCmd *Exec) Run() int {
...
+		eCmd.allocatePTY); exitCode != 0 {
+		return exitCode
+	}
+
+	return 0
+}
+
+func execHelper(cmd string, sshClient ssh.Client, dockerID string, requestPTY bool) int {
```
kklin: I'd prefer the order of arguments to be: `(sshClient, dockerID, cmd, requestPTY)`

quiltctl/command/exec.go
```diff
@@ -112,19 +111,28 @@ func (eCmd *Exec) Run() int {
...
+
+	command := strings.Join([]string{"docker exec", flags, dockerID, cmd}, " ")
+
+	if err := sshClient.Run(requestPTY, command); err != nil {
 		if exitErr, ok := err.(*golangSSH.ExitError); ok {
 			log.WithError(err).Debug(
 				"SSH command returned a nonzero exit code")
 			return exitErr.ExitStatus()
 		}
-
```
kklin: You add this new line in a previous commit, then remove it here I think. To make the diff cleaner I think you should pick one of the two and just stick with it.

quiltctl/command/exec_test.go
```diff
@@ -155,3 +156,32 @@ func checkExecParsing(t *testing.T, args []string, expArgs Exec, expErr error) {
 	assert.Equal(t, expArgs.privateKey, execCmd.privateKey)
 	assert.Equal(t, expArgs.allocatePTY, execCmd.allocatePTY)
 }
+
+func TestExecHelperErrors(t *testing.T) {
+	//SSH ExitError
+	mockSSHClient := new(testutils.MockSSHClient)
+	mockSSHClient.On("Run", true, "docker exec -it id123 echo hello").
+		Return(&golangSSH.ExitError{}).Once()
```
kklin: Could you make the exit code non-zero? Makes it more obvious what it's testing.

quiltctl/command/exec_test.go
```diff
@@ -155,3 +156,32 @@ func checkExecParsing(t *testing.T, args []string, expArgs Exec, expErr error) {
 	assert.Equal(t, expArgs.privateKey, execCmd.privateKey)
 	assert.Equal(t, expArgs.allocatePTY, execCmd.allocatePTY)
 }
+
+func TestExecHelperErrors(t *testing.T) {
+	//SSH ExitError
```
kklin: Space after `//`.

quiltctl/command/exec_test.go
```diff
@@ -155,3 +156,32 @@ func checkExecParsing(t *testing.T, args []string, expArgs Exec, expErr error) {
...
+	mockSSHClient := new(testutils.MockSSHClient)
+	mockSSHClient.On("Run", true, "docker exec -it id123 echo hello").
+		Return(&golangSSH.ExitError{}).Once()
+
+	exitCode := execHelper("echo hello", mockSSHClient, "id123", true)
+	assert.Equal(t, 0, exitCode)
+	mockSSHClient.AssertExpectations(t)
+
+	//Other SSH error
```
kklin: Same here re space

quiltctl/command/login.go
```diff
@@ -0,0 +1,157 @@
...
+
+	sshClient, err := lCmd.sshGetter(containerClient.Host(), lCmd.privateKey)
+	if err != nil {
+		log.WithError(err).Info("Error opening SSH connection")
+		return 1
+	}
+	defer sshClient.Close()
+
+	bestShell := findBestAvailableShell(shells, sshClient, container.DockerID)
+	if bestShell == "" {
```
kklin: Eh, seems hacky. I'd prefer having `findBestAvailableShell` return an error.

quiltctl/command/login.go
```diff
@@ -0,0 +1,157 @@
...
+	log.Debug("executing ", bestShell)
+
+	return execHelper(bestShell, sshClient, container.DockerID, true)
+}
+
+// findBestAvailableShell returns the name of a shell that is executable on the remote
+// docker container, first trying shells listed in `shells`, then /bin/sh.
+//
+// If findBestAvailableShell fails to find a shell, then it returns the empty string.
+// The current implementation uses /bin/sh to test whether other shells exist, and will
```
kklin: I'm still iffy about depending on `/bin/sh`.. But this is definitely better than having nothing, and if it ends up being a bad assumption we can revisit this implementation.

quiltctl/command/login.go
```diff
@@ -0,0 +1,157 @@
...
+	for _, shell := range shells {
+		if tryShell(shell, sshClient, dockerID) {
+			return shell
+		}
+	}
+	return "/bin/sh"
+}
+
+var tryShell = func(shell string, sshClient ssh.Client, dockerID string) bool {
+	log.Debug("trying shell ", shell)
```
kklin: Nit - I prefer making the caller log as much as possible. So I'd move the `log`s in this func to inside the loop.

quiltctl/command/login.go
```diff
@@ -0,0 +1,157 @@
...
+
+	checkCmd := fmt.Sprintf("/bin/sh -c 'hash %s 2>/dev/null'", shell)
```
kklin: Does this redirect `stdout`?

quiltctl/command/login.go
```diff
@@ -0,0 +1,157 @@
...
+	errCode := execHelper(checkCmd, sshClient, dockerID, false)
```
kklin: Once you make the `log` changes, simply `return errCode == 0`.

quiltctl/command/login_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	shells := []string{"shellA", "shellB"}
+	mockSSHFail := new(testutils.MockSSHClient)
+	mockSSHFail.On("Run", mock.Anything, mock.Anything).Return(assert.AnError)
+
+	assert.Equal(t, "", findBestAvailableShell(shells, mockSSHFail, "id123"))
+}
+
+func TestLoginDefaultToSh(t *testing.T) {
+	origTryShell := tryShell
```
kklin: Eh, I don't think we need to set that func back

quiltctl/command/login.go
```diff
@@ -0,0 +1,157 @@
...
+
+	for _, shell := range shells {
+		if tryShell(shell, sshClient, dockerID) {
+			return shell
+		}
+	}
+	return "/bin/sh"
+}
+
+var tryShell = func(shell string, sshClient ssh.Client, dockerID string) bool {
```
kklin: I'd prefer the order to be `sshClient, dockerID, shell`

quiltctl/command/login.go
```diff
@@ -4,6 +4,8 @@ import (
 	"errors"
```
kklin: I'm not a big fan of this feature -- it seems to just complicate things without much benefit. I don't see any reason why you'd expect your local machine's shell to have any relation to the container's.

(cc @mmussomele @ejj @LuiseV)

sirspinach: In case this is relevant, docker run and exec both return the exit code of the commands they run.
 ```
% docker run --rm ubuntu bash -c false; echo $?
1
```

File: quiltctl/command/login_test.go (hunk 22 above)
Comment (sirspinach) [inline quiltctl/command/login_test.go]: `TestLoginTryShellSuccess` also uses the original `tryShell()`, and failed after I removed this `defer func() { tryShell = origTryShell }()`.

File: quiltctl/command/login.go (hunk 23 above)
Comment (sirspinach) [inline quiltctl/command/login.go]: For consistency, I'm also changing the order of findBestAvailableShell to `sshClient, dockerID, shells`

File: quiltctl/command/exec_test.go (hunk 14 above)
Comment (sirspinach) [inline quiltctl/command/exec_test.go]: Okay, but this will require editing an unexported variable in the `ssh.Waitmsg` struct.

Comment (sirspinach) [inline quiltctl/command/exec_test.go]: nvm, `reflect` doesn't allow me to change an unexported struct field.

Comment (sirspinach) [inline quiltctl/command/exec_test.go]: Gonna try making an interface.

Comment (kklin) [inline quiltctl/command/exec_test.go]: Meh... seems like a huge amount of effort.

Comment (kklin) [inline quiltctl/command/exec_test.go]: I'm fine leaving it as is

**Comment by sirspinach** [comment]
ready for further review.

File: quiltctl/command/exec_test.go [hunk 25]
```diff
@@ -14,6 +15,18 @@ import (
 	"github.com/NetSys/quilt/quiltctl/testutils"
 )
 
+type mockExitError struct {
+	exitCode int
+}
+
+func (err mockExitError) Error() string {
+	return fmt.Sprint("Exitted with error code: ", err.exitCode)
```
kklin: "Exited"

quiltctl/command/exec_test.go
```diff
@@ -20,7 +20,7 @@ type mockExitError struct {
 }
 
 func (err mockExitError) Error() string {
-	return fmt.Sprint("Exitted with error code: ", err.exitCode)
+	return fmt.Sprint("exitted with error code: ", err.exitCode)
```
kklin: Should be changed in the previous commit

quiltctl/command/login.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return fmt.Errorf("target container must be a number: %s", args[0])
+	}
+
+	lCmd.targetContainer = targetContainer
+	return nil
+}
+
+// Run logs into a shell on the remote container.
+func (lCmd *Login) Run() int {
```
kklin: Could you add some coverage for this method?

quiltctl/command/exec.go
```diff
@@ -31,6 +30,13 @@ type Exec struct {
 	clientGetter client.Getter
 }
 
+// exitError is an interface to "golang.org/x/crypto/ssh".ExitError that allows for
+// mocking in unit tests.
+type exitError interface {
```
sirspinach: exitError should have been added in the previous commit.

sirspinach:
ready for more review.

quiltctl/command/login.go
sirspinach: Don't feel strongly about whether or not to include this.

An argument for this patch: a user who is using `zsh` would most likely prefer `zsh` over others shells when he or she logs into a container. Though it's unlikely that anything other than bash, ash, and sh would be automatically installed in a container.

kklin: Yeah, and that user won't have their dotfiles in place.

quiltctl/ssh/native.go
```diff
@@ -86,10 +86,11 @@ func (c NativeClient) Run(requestPTY bool, command string) error {
 	}
 	defer session.Close()
 	session.Stdout = os.Stdout
-	session.Stdin = os.Stdin
 	session.Stderr = os.Stderr
 
 	if requestPTY {
+		session.Stdin = os.Stdin
```
ejj: I don't see the point of this patch, why not just do it in the previous patch when you introduce the bug?

sirspinach: This bug was around even before the SSH refactor: https://github.com/NetSys/quilt/pull/639/commits/bc9084794c5597be3e89fa6a11796391a572d612#diff-0a77957168c83c3e168d5673d6900f35L111

ejj:
@kklin, @sirspinach, what's the status on this one?  Did we come to a conclusion on the last commit?  I don't have enough bash knowledge to have an intelligent opinion on it.

kklin:
Let's ditch it. I think there was miscommunication on what feature @mmussomele wanted (it was an implementation suggestion, not a feature request), I don't like it, and @sirspinach isn't attached to it either.

Certainly shouldn't block the other stuff getting merged.

ejj:
kk @sirspinach, could you rebase and remove the last commit?  I'll review it then.

sirspinach:
rebased

quiltctl/command/login.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// InstallFlags sets up parsing for command line flags.
+func (lCmd *Login) InstallFlags(flags *flag.FlagSet) {
+	lCmd.common.InstallFlags(flags)
+
+	flags.StringVar(&lCmd.privateKey, "i", "",
+		"the private key to use to connect to the host")
+	flags.Usage = func() {
+		fmt.Println("usage: quilt login [-H=<daemon_host>] [-i=<private_key>] " +
```
ejj: I think you should pull this stuff into a global variable and use the triple backticks.  It would make the code a bit easier to read, and it wouldn't count against our code coverage.

quiltctl/ssh/ssh.go
```diff
@@ -2,12 +2,11 @@ package ssh
...
 type Client interface {
-	// Connect establishes an SSH connection.
-	Connect(string, string) error
-	// RequestPTY requests a pseudo-terminal.
-	RequestPTY() error
 	// Run runs a command over the SSH connection.
-	Run(string) error
-	// Disconnect closes the SSH connection.
-	Disconnect() error
+	Run(bool, string) error
```
ejj: I think I'd prefer a newline under Run

quiltctl/command/login.go
```diff
@@ -0,0 +1,153 @@
...
+	return execHelper(sshClient, container.DockerID, bestShell, true)
+}
+
+// findBestAvailableShell returns the name of a shell that is executable on the remote
+// docker container, first trying shells listed in `shells`, then /bin/sh.
+//
+// If findBestAvailableShell fails to find a shell, then it returns the empty string.
+// The current implementation uses /bin/sh to test whether other shells exist, and will
+// fail if /bin/sh is not found on the remote container.
+func findBestAvailableShell(sshClient ssh.Client, dockerID string, shells []string) (
```
ejj: This function seems a bit odd to me.  Why make the special case for `/bin/sh`?  Seems to me you could just add it to the end of the global `shells` variable and detect it in the tryShell loop.  If none of them works, it could just `log.Fatal`

ejj: Actually here's my biggest question.  Why can't we just loop through the `shells` variable calling `execHelper` on each one.  If it works, great, we've logged in.  If it doesn't work then we move on to the next one.

ejj:
K done with my review, couple of nits, but my biggest question is the very last comment.

quiltctl/command/login.go
sirspinach: Goes back to this discussion: https://github.com/NetSys/quilt/pull/639#discussion_r91862819
"docker logs scary messages in journalctl -u docker when you try to exec something that doesn't exit"

ejj: Ah I see.

Still, could we make this a bit simpler?  So for example you could imagine we could try `bash -c pwd`, `ash -c pwd`, `sh -c pwd` for each shell to check if it exists?  That way you could get rid of the special handling for sh?

sirspinach: All of those commands result in journalctl errors when the shell doesn't exist. :(

sirspinach: If we want to avoid journalctl logs, then we cannot `docker exec` any command that does not exist. So in this approach I assumed `sh`'s existence and then used that command to test the other shells.

kklin: FWIW docker assumes `sh` exists when using the "_shell_ form" of CMD https://docs.docker.com/engine/reference/builder/#/cmd.

kklin: Another option is to check the filesystem directly..

quiltctl/command/login.go
sirspinach: fixed

quiltctl/ssh/ssh.go
sirspinach: done

quiltctl/command/login.go
```diff
@@ -0,0 +1,109 @@
...
+
+// InstallFlags sets up parsing for command line flags.
+func (lCmd *Login) InstallFlags(flags *flag.FlagSet) {
+	lCmd.common.InstallFlags(flags)
+
+	flags.StringVar(&lCmd.privateKey, "i", "",
+		"the private key to use to connect to the host")
+
+	flags.Usage = func() {
+		fmt.Println(loginUsage)
```
sirspinach: I think that in a later patch, we could do away with this `flags.Usage = func() {` part (and therefore the lack of coverage in usage messages) by adding a `usageMsg` argument to `lCmd.common.InstallFlags(flags)`.

sirspinach:
Changed login to use `sh` only.

PR #640, opened 2016-12-11
Minor test tweaks

ejj:


minion/scheduler/worker_test.go
```diff
@@ -34,33 +35,22 @@ func TestRunWorker(t *testing.T) {
...
 	runWorker(conn, dk, "1.2.3.4", *subnet)
 	dkcs, err = dk.List(nil)
-	if err != nil {
-		t.Errorf("Unexpected err %v", err)
-	}
-	if len(dkcs) != 1 || dkcs[0].Image != "Image" {
-		t.Error(spew.Sprintf("Unexpected containers: %v", dkcs))
-	}
+	assert.Nil(t, err)
+	assert.Equal(t, 1, len(dkcs))
```
sirspinach: Not a big deal, but there's an `assert.Len` method if you want to use that.

PR #641, opened 2016-12-11
ipdef: Simplify and rename the ip package

ejj:
The `ip` package had a lot of unnecessary library code that would have
been more clearly implemented by its callers directly.  Also, it's
name, `ip`, was rather annoying as it prevented the programmer from
using `ip` as a variable name.  This patch solves the problem by
creating a new package, `ipdef`, which contains our network constants
without all of the helper functions in the old `ip` package.


minion/etcd/minion.go
```diff
@@ -252,21 +245,25 @@ func syncSubnet(conn db.Conn, store Store, minion db.Minion) {
...
+	return ipNet
 }
 
-func subnetKey(subnet string) string {
-	return path.Join(subnetStore, subnet)
+func subnetKey(subnet net.IPNet) string {
+	return path.Join(subnetStore, subnet.IP.String())
 }
+
+var rand32 = rand.Uint32
```
mmussomele: would this be better at the top?

minion/etcd/minion_test.go
```diff
@@ -130,23 +131,18 @@ func TestFilter(t *testing.T) {
...
 	store := newTestMock()
-	subnetAttempts = 5000 // big so we don't error unless something is wrong
-	defer func() {
-		subnetAttempts = 1000
-	}()
-	minionMaskBits, _ := ip.SubMask.Size()
+	minionMaskBits, _ := ipdef.SubMask.Size()
 	emptyBits := uint32(0xffffffff >> uint(minionMaskBits))
 	for i := 0; i < 3500; i++ {
 		sub, err := generateSubnet(store, db.Minion{PrivateIP: "10.1.0.1"})
```
mmussomele: No reason not to call this `subnet` now, might help with clarity slightly

minion/network/plugin/plugin.go
```diff
@@ -199,3 +200,5 @@ func endpointExists(eid string) (bool, error) {
 	_, name := network.VethPairNames(eid)
 	return network.LinkExists("", name)
 }
+
+var rand32 = rand.Uint32
```
mmussomele: same here with moving to the top

minion/etcd/minion.go
ejj: So here's my reasoning on this, i generally like functions to be ordered from most important to least important, with variables/constants at the top.  So if you consider this a function, it's the least important and thus belongs at the bottom.  It's also similar to some other packages where we have the sorting interface near the bottom.  Basically, in short, readers shouldn't care about it hence its location.

That said, I don't feel that strongly about it...

mmussomele: Yea sure then. I don't particularly care either, seems like you've thought about it though so that's fine with me

PR #645, opened 2016-12-12
minion: Wait for self to exist before syncing keys

kklin:
Should fix the CI errors.


minion/keys_test.go
```diff
@@ -85,3 +86,22 @@ func TestSyncKeysError(t *testing.T) {
...
 	assert.EqualError(t, err, "operation not permitted")
 }
+
+func TestWaitForMinion(t *testing.T) {
+	t.Parallel()
+
+	start := time.Now()
+	conn := db.New()
+	go func() {
+		time.Sleep(1 * time.Second)
```
ejj: Meh, I don't like this much.  I'd prefer we either not unit test this function, or mock out the time.Sleep.  I generally don't like it if unit tests have to sleep

secant: Just wondering, why's that? (I'm assuming the tests are run in parallel, 1 second seems pretty short)

mmussomele: One thing you could do is run both functions in goroutines, and have them close channels after committing the minion and waiting for the minion, respectively. Then just have this at the end
```
select {
    case <-commitChan:
        break
    case <-waitChan:
        t.Fatal(...)
    case <- timeout:
        ....
}

mmussomele: actually what might be cleaner is to have each process send a known value into a channel, and assert that the values come out in the right order

ejj: Personally I think we should either (1) mock out time.Sleep (as done in some other tests) or (2) just delete the test.  It's only three lines of code that are "obviously correct", not that critical that its covered.

PR #646, opened 2016-12-15
minion: Use the database to signal when the OVS is ready

ejj:
Before this patch, the network code attempted to detect whether or not
OVS was ready by inspecting the system.  This turned out to be rather
fragile -- just because OVS is up, and may have partially configured
the system, doesn't mean it's finished.  This patch, instead, has the
supervisor signal directly when it's finished configuring OVS.  Since
OVS commands themselves block until OVS has fully implemented their
instruction, this approach blocks the network until OVS is fully
ready.


minion/network/network.go
```diff
@@ -154,3 +144,8 @@ func (dbs dbslice) Len() int {
 func (dbs dbslice) Get(i int) interface{} {
 	return dbs[i]
 }
+
+func supervisorInit(view db.Database) bool {
```
mmussomele: maybe checkSupervisorInit? As is sounds like it's init-ing the supervisor

PR #647, opened 2016-12-15
network: Rely on libnetwork to set MAC addresses

ejj:
This patch allows libnetwork to set container MAC addresses instead of
doing it in minion/network/worker.go.  This simplifies the code
somewhat.

Note that a previous patch, commit 646149f89, attempted to
do something similar.  It failed and had to be reverted (commit
646149f89) because libnetwork was not actually setting the
container MAC addresses, instead failing silently.  This was because
the IP address passed to our network plugin was actually a prefix,
that our plugin failed to parse before converting to a MAC.  This
patch resolves that issue.


minion/docker/docker.go
```diff
@@ -286,6 +287,7 @@ func (dk Client) Get(id string) (Container, error) {
 	if len(networks) == 1 {
 		config := dkc.NetworkSettings.Networks[networks[0]]
 		c.IP = config.IPAddress
+		c.Mac = config.MacAddress
```
mmussomele: should we default to `dkc.NetworkSettings.MacAddress` like we do for the other fields above?

ejj: Nice catch

PR #650, opened 2016-12-16
network: Bring up quilt-int bridge port in the supervisor

ejj:
The quilt-int bridge port needs two simple modifications after its
created: it must be brought up, and it must have the IP prefix
10.0.0.1/8 assigned to it.  Attempting to achieve these things in the
minion/network/worker run loop is overly complex.  This patch instead,
simply does them once immediately after the port is created.


minion/supervisor/supervisor.go
```diff
@@ -162,14 +163,27 @@ func (sv *supervisor) updateWorker(IP string, leaderIP string, etcdIPs []string)
...
+	}
+
+	err = execRun("ip", "link", "set", "dev", "quilt-int", "up")
+	if err != nil {
+		log.WithError(err).Warnf("Failed to bring up quilt-int")
+		return
+	}
+
+	ip := net.IPNet{IP: ipdef.GatewayIP, Mask: ipdef.QuiltSubnet.Mask}
+	err = execRun("ip", "addr", "add", ip.String(), "dev", "quilt-int")
```
kklin: I saw we were referencing a variable in the old code, should we reference the same one here?

PR #652, opened 2016-12-16
Haproxy Simplifications

ejj:
For reasons I don't fully understand, this seems to make haproxy a lot more stable for me ...

In general the mean API is so unstable, we really need an integration test for it (I'll start on that next), but hopefully simplifying the haproxy config file will make it easier to debug when we run into problems.


specs/haproxy/config/haproxy.cfg
```diff
@@ -1,31 +1,14 @@
...
-    timeout client          1m
-    timeout server          1m
-    timeout http-keep-alive 10s
-    timeout check           10s
-    maxconn                 3000
+    log global
+    mode    http
+    timeout connect 5000
+    timeout client 5000
+    timeout server 5000
```
mmussomele: there seemed to be some effort to align these before, does that matter?

ejj: well certainly, they should either be all aligned or all unaligned ...

PR #653, opened 2016-12-17
ovsdb: Remove unused SetBridgeMac() function

ejj:


ejj:
I don't think this needs to be reviewed, just double checking it passes CI.

PR #654, opened 2016-12-18
link: Remove dead listVeths() function

ejj:


ejj:
No need to review, just pushing so it gets CI

PR #655, opened 2016-12-20
network: Partially migrate to the netlink library

ejj:
This pull request partially migrates us to the netlink library instead of relying on the `ip` command line utility.  More to come, but this is a good starting point.


minion/ipdef/ipdef.go
```diff
@@ -45,3 +46,18 @@ func IPToMac(ip net.IP) string {
 	ip = ip.To4()
 	return fmt.Sprintf("02:00:%02x:%02x:%02x:%02x", ip[0], ip[1], ip[2], ip[3])
 }
+
+// Allow mocking out for unit tests.
+var ifNameSize = syscall.IFNAMSIZ
```
mmussomele: Perhaps we could subtract the 1 here instead? Might simplify things, and make the tests more obvious

minion/network/plugin/plugin_test.go
```diff
@@ -112,16 +112,17 @@ func TestDeleteEndpoint(t *testing.T) {
...
-	err := d.DeleteEndpoint(delReq)
-	assert.Nil(t, err)
+	err = d.DeleteEndpoint(delReq)
+	assert.NoError(t, err)
 
 	d.Leave(&dnet.LeaveRequest{EndpointID: zero})
 	delReq = &dnet.DeleteEndpointRequest{EndpointID: zero}
 	err = d.DeleteEndpoint(delReq)
-	assert.NotNil(t, err)
+	assert.EqualError(t, err, "endpoint 0000000000000 doesn't exists")
```
mmussomele: It might be worth storing all these error strings in format strings, in case we change them at some point

minion/network/plugin/plugin.go
```diff
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/NetSys/quilt/minion/ipdef"
-	"github.com/NetSys/quilt/minion/network"
+	"github.com/vishvananda/netlink"
```
mmussomele: Just a nit, but I think the imports look best when native imports, quilt imports and third party imports are grouped by separating newlines

minion/network/plugin/plugin.go
```diff
@@ -92,61 +94,58 @@ func (d driver) EndpointInfo(req *dnet.InfoRequest) (*dnet.InfoResponse, error)
...
-	}
-
 	// We just need to create the Veth and tell Docker where it should go; Docker
 	// will take care of moving it into the container and renaming it.
-	tempPeer, err := network.AddVeth(req.EndpointID)
+	outer := ipdef.IFName(req.EndpointID)
+	inner := ipdef.IFName("tmp_" + req.EndpointID)
+	err := linkAdd(&netlink.Veth{
+		LinkAttrs: netlink.LinkAttrs{Name: outer, MTU: mtu},
+		PeerName:  inner})
```
mmussomele: Should these closing braces/parentheses be on their own line?

minion/network/plugin/plugin_test.go
```diff
@@ -2,66 +2,66 @@ package plugin
...
-	one  = "1111111111111"
+	zero  = "000000000000000000000000000000000000000000000000"
+	one   = "111111111111111111111111111111111111111111111111"
+	links = map[string]netlink.Link{}
 )
 
-type mock map[string]struct{}
+func mockLinkAdd(l netlink.Link) error {
+	name := l.Attrs().Name
+	if len(name) > syscall.IFNAMSIZ {
```
mmussomele: Shouldn't this be `>=`, because of the null terminator?

minion/network/worker.go
```diff
@@ -21,6 +22,8 @@ import (
 	"github.com/NetSys/quilt/minion/ovsdb"
 	"github.com/NetSys/quilt/stitch"
 	"github.com/NetSys/quilt/util"
+	"github.com/vishvananda/netlink"
```
mmussomele: same with the import grouping

mmussomele:
One thing I was wondering, when migrating to netlink, could we also get rid of getPublicInterface? If we could do that, I think we could ditch ipExecVerbose and be done with shelling out to IP completely

minion/ipdef/ipdef.go
ejj: I disagree -- I think its much more clear for the constant to be consistent with what the system defines (and what linux network programmers already know and are used to), and then let the unit tests verify that the code accounts for the -1 properly.

minion/network/plugin/plugin.go
ejj: Meh, I don't care, but I'll change it.

ejj:
Updated comments

minion/ipdef/ipdef.go
tejasmanohar: I agree w/ @ejj here.

minion/network/plugin/plugin.go
tejasmanohar: sounds like a python thing :P ... but i thought goimports or gofmt does this automatically

minion/network/worker.go
```diff
@@ -964,22 +890,8 @@ func generateEtcHosts(dbc db.Container, labels map[string]db.Label,
...
-}
-
 func patchPorts(id string) (br, quilt string) {
-	return fmt.Sprintf("%s_br", id[0:12]), fmt.Sprintf("%s_q", id[0:12])
-}
-
-func ipExec(namespace, format string, args ...interface{}) error {
-	_, _, err := ipExecVerbose(namespace, format, args...)
-	return err
+	return ipdef.IFName("br_" + id), ipdef.IFName("q_" + id)
```
tejasmanohar: nit: sprintf :)

ejj: Meh I disagree

PR #657, opened 2016-12-21
Table restrict

mmussomele:


cluster/foreman/foreman.go
```diff
@@ -91,6 +93,7 @@ func RunOnce(conn db.Conn) {
 	updateMinionMap(machines)
 
 	/* Request the current configuration from each minion. */
+	tr := conn.WithTables(db.MachineTable)
```
mmussomele: I think it's safe now, but that's a good point. I'll change that for future proofing

db/db_test.go
```diff
@@ -125,6 +126,244 @@ func TestMachineString(t *testing.T) {
...
+// table sets do not overlap.
+func TestWithTablesConcurrent(t *testing.T) {
+	// Run the deadlock test multiple times to increase the odds of detecting a race
+	// condition
+	for i := 0; i < 10; i++ {
+		checkIndependentTransacts(t)
+	}
+}
+
+// returns false when the transactions deadlock
```
LuiseV: Capitalize

db/db_test.go
```diff
@@ -125,6 +126,244 @@ func TestMachineString(t *testing.T) {
...
+				return nil
+			}
+		})
+	}
+
+	check := make(chan bool)
+	defer close(check)
+	go func() {
+		first := <-results
+		second := <-results
```
LuiseV: How do we enforce the order between `first` and `second`? I assume it has something to do with `<-oneStarted` in line 326, but I'm not super familiar with go routines, and can't really convince myself how this works.

mmussomele: Yea, oneStarted isn't closed until one has entered its transaction, so if these transactions block like they should, two should not enter it's transaction until after one has finished

mmussomele:
Comments fixed. Can't seem to get the mean app to work, but nothing is noticeably wrong in the logs, haproxy seems to be acting up. SparkPI works fine if that's any consolation

db/db_test.go
LuiseV: Right, thanks, that makes sense!

mmussomele:
Yea it's not the prettiest construct, but I think it's worth it, since it means we have to do ugly line breaks less often.

db/db.go
```diff
@@ -28,38 +29,63 @@ type row interface {
...
 }
 
-// Transact executes database transactions.  It takes a closure, 'do', which is operates
-// on its 'db' argument.  Transactions are not concurrent, instead each runs
-// sequentially on it's database without conflicting with other transactions.
-func (cn Conn) Transact(do func(db Database) error) error {
-	cn.lock.Lock()
-	err := do(cn.db)
+// Txn creates a new Transaction object connected to the same database, but with
+// restricted access to only the given tables. Transaction objects can only execute
```
ejj: I'd ditch this last sentence "Transaction objects can only ..."  Easy for that sort of thing to get stale.

db/db_test.go
```diff
@@ -15,7 +16,7 @@ func TestMachine(t *testing.T) {
 	conn := New()
 
 	var m Machine
-	err := conn.Transact(func(db Database) error {
+	err := conn.Txn(AllTables...).Run(func(db Database) error {
```
ejj: Btw, why not make it so `Txn()` implies all tables?  I don't feel all that strongly about it, just curious if there was a reason not to.

mmussomele: I felt that it was too easy for someone to just forget the tables and have a transaction lock them all in the future (and no-body notice). This way, you actually have to think about what tables you use, or explicitly use all of them.

db/db_test.go
```diff
@@ -125,6 +126,242 @@ func TestMachineString(t *testing.T) {
...
+	tr.Run(func(view Database) error {
+		view.InsertEtcd()
+
+		return nil
+	})
+}
+
+// Transactions should be able to run concurrently if their table sets do not overlap.
+func TestTxnConcurrent(t *testing.T) {
+	// Run the deadlock test multiple times to increase the odds of detecting a race
```
ejj: Meh, this is unlikely to work ...  that said, not a bad test.

We should run with the race detector at some point btw.  That would help diagnose a lot of potential problems.

ejj: Yeah actually, more I look at this -- I'm not sure it's a good test.  I think it's very hard to reliable detect a deadlock, not sure this test will do it.  Also, it gives us somewhat of a false sense of confidence.  A much better approach would be to run with the race detector.

ejj:
LGTM, one minor nit in the comment, and I think the deadlock test is kind of overkill.  Let me know when updated, and I'll merge.

PR #658, opened 2016-12-22
tests: Update check_logs to use the quilt client

ejj:
The check_logs test relied on shelling out to `quilt machines` and
parsing the output.  When the output changed, the test began to fail.
This patch switches over to using the quiltctl client directly,
hopefully solving this problem in the future.


sirspinach:
Looks good to me.

PR #662, opened 2016-12-24
[WIP] haproxy: Delay haproxy DNS resolution

nlsun:


nlsun:
This doesn't fix the issue, it instead just initializes with no available backends. Even if it worked this doesn't fix the underlying issue, which is that haproxy can't re-resolve an ip address unless it has a nameserver to talk to.

PR #663, opened 2016-12-24
haproxy: Change from resolving DNS to ping

nlsun:
For whatever reason, DNS resolution was not reliable. Instead of trying
to solve it, we're switching the wait method.

### DIFF
--- a/specs/haproxy/scripts/start.go
+++ b/specs/haproxy/scripts/start.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"log"
-	"net"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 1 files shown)


specs/haproxy/scripts/start.go
```diff
@@ -63,15 +61,17 @@ func configureHAProxy(addrs []string) error {
...
 }
 
-func resolveHostname(hostname string) []string {
-	for {
-		addrs, err := net.LookupHost(hostname)
-		if err == nil {
-			log.Printf("Resolved %s to %+v", hostname, addrs)
-			return addrs
-		}
+var pingComm = "until ping -q -c1 '%s' > /dev/null 2>&1; do sleep 1; done"
```
ejj: Meh, I don't like this much.  Why not just exec ping directly in a loop and check the exit code?

nlsun: I was just lazy about the extra error handling effort in golang world since the error is overloaded with both the exit code and failing to run the actual exec. I can change it.

PR #664, opened 2016-12-24
haproxy: Support running a custom haproxy image

ejj:
Many users may want to change their default haproxy to run custom
configuration of some sort. This patch allows them to by simply
calling a `SetImage` function.


specs/mean/example.js
```diff
@@ -25,7 +25,7 @@ var app = new Node({
     MONGO_URI: mongo.uri("mean-example")
   }
 });
-var haproxy = new HaProxy(3, app.services());
+var haproxy = new HaProxy.New(3, app.services());
```
kklin: This looks funky.. I think we should either rename the `HaProxy.New` function, or make it so we don't have to instantiate it with the `new` keyword.

ejj:
meh
broken, closing

PR #666, opened 2016-12-25
mean: Tweak boot waiting

kklin:
Should fix CI


ejj:
Sigh, the ping wait thing is fine for now. I think we need to look into ways to ditch it though.  I think I'll work on that next.

PR #668, opened 2016-12-27
mean: Refactor the run script

ejj:
This patch makes several simplifications to the run script which
should make the mean app much more robust.  First, it pings the mongo
containers before attempting resolve their hostnames (thus avoiding
go's aggressive /etc/hosts cache).  Second, it increases timeouts.


specs/mean/scripts/run.go
```diff
@@ -12,48 +12,33 @@ import (
...
+	dialInfo, err := mgo.ParseURL(os.Getenv("MONGO_URI"))
+	if err != nil {
+		log.Fatalf("Failed to parse MONGO_URI: %s\n", err)
 	}
 
-	if err := specs.PingWait([]string{host}); err != nil {
+	// Due to a cache in the go name resolution code, we shell out to ping first
+	// before relying on Dial.  Otherwise, dial would cache /etc/hosts before it was
+	// ready causing irrecoverable problems.
+	if err := specs.PingWait(append(dialInfo.Addrs)); err != nil {
```
kklin: Why the `append`?

PR #669, opened 2016-12-27
Default to 'latest' tag if no image tag is given

sirspinach:


minion/docker/docker.go
```diff
@@ -217,8 +220,18 @@ func (dk Client) Pull(image string) error {
 		return nil
 	}
 
+	if numColons := strings.Count(image, ":"); numColons > 1 {
+		return fmt.Errorf("Expected at most one colon in image %v", image)
+	} else if numColons == 0 {
+		image = image + ":latest"
+	}
+
+	parts := strings.Split(image, ":")
```
kklin: The fsouza dockerclient has a `ParseRepositoryTag`: https://godoc.org/github.com/fsouza/go-dockerclient#ParseRepositoryTag

sirspinach: oh, nice

sirspinach: In my current implementation, stitch refuses to compile a container whose image name contains more than one colon. 

But `ParseRepositoryTag` allows two colons so that an image name can contain the SHA-256 digest. Ex:
```busybox:latest@sha256:4a731fb46adc5cefe3ae374a8b6020fc1b6ad667a279647766e9a3cd89f6fa92 -> busybox, latest```

Should I allow and consider digests (will make the dk.Pull logic a little more complicated), or leave the code as is?

Comment (kklin) [inline minion/docker/docker.go]: How would it affect the `Pull` logic?

Comment (sirspinach) [inline minion/docker/docker.go]: hmm actually it doesn't. I was thinking about the imageCache in the docker tests.

File: stitch/bindings.js [hunk 2]
```diff
@@ -340,6 +340,14 @@ Machine.prototype.replicate = function(n) {
 };
 
 function Container(image, command) {
+    var parts = image.split(":");
+    var numColons = parts.length - 1
+    if (numColons > 2) {
```
ejj: You know I think I would simplify this a bit and just say `if len(parts) > 2 || (len(parts) > 1 && parts[1] == "")`

And then just make the error message a simple "Could not parse container image" or something.

Though on the other end, I'm not sure this patch should exist though, I really want @kklin's opinion.  I could imagine alternate container formats that support a colon for example -- seems like an implementation detail of the deployment engine more than something should be in the language.

kklin: I do think that the feedback loop for finding that you mistyped the image name is very big, and that it's nice to close the gap. It's slightly better in the deployment engine I think. It'd return an error if it found any errors while vetting.
Or I could imagine this being a feature of a linter.. But we're nowhere close to ready for that.

sirspinach:
Moved image check to `deployment.prototype.vet` and simplified error messages.

kklin:
@sirspinach I was thinking of moving the check to https://github.com/NetSys/quilt/blob/master/api/server/server.go#L93

sirspinach:
Moved image check to api/server.

PR #670, opened 2016-12-28
Tweak quilt-tester order

ejj:


ejj:
Meh, i think it's fine, I don't really want to mess with it, when we move to jenkins it will have it's own way of doing things, and we'll switch to that.

PR #673, opened 2016-12-29
quiltctl: Parallelize `quilt ps`

ejj:
The `quilt ps` utility felt quite slow, also it printed machines some
time before printing the containers which was a bit jarring.  This
patch parallelizes it making it quite a bit zippier.


quiltctl/command/ps.go
```diff
@@ -44,42 +43,68 @@ func (pCmd *Ps) Parse(args []string) error {
...
+			connectionErr <- err
+		}()
+
+		go func() {
+			containers, err = leaderClient.QueryContainers()
+			containerErr <- err
+		}()
+	}
+
+	if err := <-machineErr; err != nil {
```
sirspinach: Ah, this is a neat use of channels

quiltctl/command/ps.go
```diff
@@ -44,42 +43,68 @@ func (pCmd *Ps) Parse(args []string) error {
 
 // Run retrieves and prints all machines and containers.
 func (pCmd *Ps) Run() int {
+	if err := pCmd.run(); err != nil {
+		fmt.Println(err)
```
kklin: Should go to `stderr`

PR #675, opened 2016-12-30
Another attempt to fix the broken mean test.

ejj:


quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -64,18 +67,23 @@ func httpGetTest(machines []db.Machine, containers []db.Container) {
...
+			} else {
+				log.Error(resp)
+				failed = true
 			}
-
-			log.Info(resp)
 		}
 	}
+
+	return !failed
```
secant: I think this should just be `failed`

specs/mongo/scripts/run.go
```diff
@@ -7,6 +7,8 @@ import (
 	"os/exec"
```
sirspinach: In the commit message: "dieing" -> "dying"

quilt-tester/tests/30-mean/check_mean.go
ejj: No, I think it returns true if passed, false if failed.  So !failed => true if it passed.

secant: yeah my bad, got confused

PR #676, opened 2016-12-31
tests: Log containers in the mean test

ejj:
Hopefully this will help us debug failures a little bit more easily.


ejj:
I agree btw.  Something to do with Jenkins I think.

PR #677, opened 2017-01-01
network: Rely on the default docker nameserver

ejj:
Before this patch, we were manually figuring out what the default
nameserver for the host machine was and writing that into the
container after it was up.  Not only did this require us to include a
large amount of rather ugly code, it also meant we were writing files
into a running container which isn't particularly safe.

Ideally, we would instead be able to configure each container's
/etc/resolv.conf file from docker plugin.  However docker hates plugin
writers and wants us instead to use the DNS server they bundled.
Hopefully, they'll give us a way to disable it in future, but for now
it works so we'll stick with it.


tejasmanohar:
Ah, I see. I was wondering why the defaults weren't kicking in after reading docs. Makes sense now.

PR #679, opened 2017-01-03
quiltctl: Colored diff output

Baisang:
Uses github.com/fatih/color to colorize diff output printed to the console.

### DIFF
--- a/quiltctl/command/run.go
+++ b/quiltctl/command/run.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/fatih/color"
 	"github.com/pmezard/go-difflib/difflib"
 	"github.com/robertkrimen/otto"
 
@@ -118,7 +119,7 @@ func (rCmd *Run) Run() int {
 		if diff == "" {
 			fmt.Println("No change.")
 		} else {
-			fmt.Println(diff)
+			fmt.Println(colorizeDiff(diff))
 		}
 		shouldDeploy, err := confirm(os.Stdin, "Continue with deployment?")
 		if err != nil {
@@ -184,6 +185,21 @@ func prettifyJSON(toPrettify string) (string, error) {
 		return "", err
 	}
 	return prettified.String(), nil
+}
+
+func colorizeDiff(toColorize string) string {
+	var colorized bytes.Buffer
+	for _, line := range strings.SplitAfter(toColorize, "\n") {
+		switch {
+		case strings.HasPrefix(line, "+"):
+			colorized.WriteString(color.GreenString("%s", line))
+		case strings.HasPrefix(line, "-"):
+			colorized.WriteString(color.RedString("%s", line))
+		default:
+			colorized.WriteString(line)
+		}
+	}
+	return colorized.String()
 }
 
 // Saved in a variable so that we can mock it for unit testing.
--- a/quiltctl/command/run_test.go
+++ b/quiltctl/command/run_test.go
@@ -10,6 +10,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	logrusTestHook "github.com/Sirupsen/logrus/hooks/test"
+	"github.com/fatih/color"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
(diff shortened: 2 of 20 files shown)


quiltctl/command/run.go
```diff
@@ -118,7 +119,18 @@ func (rCmd *Run) Run() int {
 		if diff == "" {
 			fmt.Println("No change.")
 		} else {
-			fmt.Println(diff)
+			for _, line := range strings.Split(diff, "\n") {
```
ejj: Should we check if the terminal supports color?  Not sure what's standard.  What happens to this if you pipe it into `less` for example?

Baisang: I was thinking about that, but when I redirected output into a file things appeared to display OK (at least in bash), and I just tried with `less` now and at least mine was able to skip displaying all of the ANSI color escapes.

We could possibly make a call to `tput colors` or something and see what that returns, probably other ways to check as well.

kklin: Looks like the library handles it for us: https://github.com/fatih/color/blob/master/color.go#L18

Baisang: Ah, I must have missed that. Hooray!

quiltctl/command/run.go
```diff
@@ -118,7 +119,18 @@ func (rCmd *Run) Run() int {
...
+				if len(line) > 0 {
+					switch line[0] {
+					case '+':
+						color.Red("%v\n", line)
```
kklin: It's much more standard for additions to be green, I think.

quiltctl/command/run.go
```diff
@@ -118,7 +119,18 @@ func (rCmd *Run) Run() int {
 		if diff == "" {
 			fmt.Println("No change.")
 		} else {
-			fmt.Println(diff)
+			for _, line := range strings.Split(diff, "\n") {
+				if len(line) > 0 {
```
kklin: What if the line is empty? It won't get printed right?
There's a `strings.HasPrefix` that would make this slightly cleaner btw.

quiltctl/command/run.go
```diff
@@ -118,7 +119,18 @@ func (rCmd *Run) Run() int {
 		if diff == "" {
```
kklin: I'd pull the diff output into a function so that it can be unit tested. (it's my fault that it's not already unit tested, I wrote the code originally :))

quiltctl/command/run.go
Baisang: You're right, I meant it to be the other way around. I guess sometime when I was linting/squashing it got reversed, will fix

quiltctl/command/run.go
Baisang: Yeah, I think empty lines shouldn't be printed. I'll switch to strings.HasPrefix, had that originally but wasn't sure which seemed cleaner.

kklin: Why not? If it's part of the diff output..

Baisang: Since I'm splitting by "\n", "abc\n" would split into ["abc", ""], and since we are using `Println` we would get an extra "\n" if we printed empty lines. So the output would be "abc\n\n" instead of "abc\n". If that's OK, then I think printing empty lines is alright.

quiltctl/command/run.go
Baisang: So like writing a function that returns a ANSI-colorized diff string? (or "No change." if it's empty)

quiltctl/command/run.go
kklin: Oh.. meh. If there's a `"\n\n"` in the diff, it's wrong though right? I think if you pull the colorization code into a function it'll be easier to deal with those newlines.

quiltctl/command/run.go
kklin: I was thinking the entire `if` statement originally, but just the colorized diff makes more sense.

quiltctl/command/run.go
Baisang: This also means that of course if we are splitting `"abc\n\ndef"` the printed output would just be `"abc\ndef"`, but I'm not sure which is a bigger issue since we're working with JSON strings so there shouldn't be a pattern like `"abc\n\ndef"`? I guess just printing empty lines would make it easier since any extra newlines would just be at the end.

kklin: I guess my larger point is that it's a bad sign that we have to think about all these cases. I suspect if we tweak the code a bit the code will be more obviously correct.

Baisang:
OK, so I changed the colorizing to happen in a separate function and use `strings.SplitAfter` instead. This way we should preserve more information regarding newlines, e.g. `"abc\ndef\n"` should split into `["abc\n", "def\n", ""]` and we can just print all 3.

I just noticed some checks failed, might be because I originally used a different function (`difflib.SplitLines`) and decided that `strings.SplitAfter` made more sense after I pushed and just forced-pushed an update?

quiltctl/command/run.go
```diff
@@ -198,6 +187,21 @@ func prettifyJSON(toPrettify string) (string, error) {
 	return prettified.String(), nil
 }
 
+func colorizeDiff(toColorize string) string {
+	var colorized bytes.Buffer
+	for _, line := range strings.SplitAfter(toColorize, "\n") {
+		switch {
+		case strings.HasPrefix(line, "+"):
+			colorized.WriteString(color.GreenString("%v", line))
```
kklin: Nit - the format verb should be `%s` (also in `RedString`)

quiltctl/command/run.go
```diff
@@ -198,6 +187,21 @@ func prettifyJSON(toPrettify string) (string, error) {
...
+		case strings.HasPrefix(line, "-"):
+			colorized.WriteString(color.RedString("%v", line))
+		default:
+			colorized.WriteString(fmt.Sprint(line))
```
kklin: No need for the `Sprint`

PR #680, opened 2017-01-03
tests: Add bandwidth test

sirspinach:
This patch adds a test that sets up 5 iperf containers and confirms
that the bandwidth between the containers is at least 100 Mb/s.

Bandwidth is currently about 300 Mb/s between containers on different machines, and about 8000Mb/s between containers on the same host.

Test output looks like this (with a higher bandwidth requirement to show errors):
https://gist.github.com/sirspinach/a7a62524ffe3aa97ed5c40af45cc9ede


quilt-tester/tests/75-bandwidth/bandwidth.js
```diff
@@ -0,0 +1,11 @@
+var infrastructure = require("github.com/NetSys/quilt/quilt-tester/config/infrastructure.js")
```
kklin: We can't have two checks for the same test suite right now, but what do you think about combining this and the ping test once it's in?

quilt-tester/tests/75-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,141 @@
...
+			res.bandwidthMbPerSec = -1
+		} else {
+			res.bandwidthMbPerSec = bandwidth
+		}
+	}
+
+	results <- res
+}
+
+func parseBandwidth(output string) (int, error) {
```
kklin: IIRC iperf can output JSON, might be easier to parse.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,141 @@
...
+type testResult struct {
+	client, server    db.Container
+	output            string
+	bandwidthMbPerSec int
+	err               error
+}
+
+func test(client, server db.Container, results chan<- testResult) {
+	cmd := exec.Command("quilt", "exec", strconv.Itoa(client.StitchID),
+		"iperf3", "-c", server.IP, "-f", "m", "-t", "30")
```
kklin: What do you think about using the server's hostname rather than IP? Would be an easy way to get coverage for `/etc/hosts`.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
sirspinach: Thought about that, but decided to parse normal output because it's more readable in logs, and I couldn't figure out how to output both normal output and JSON for the same test.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
sirspinach: hmmm... I think it's probably best to have another test for hostnames. Partly because modularity, partly because we need to check all the other hostnames to properly test `/etc/hosts`.

quilt-tester/tests/75-bandwidth/bandwidth.js
sirspinach: It would be nice to have related network tests in one suite. We'd have to later be careful about adding other tests that might affect the iperf bandwidth though.

quilt-tester/tests/75-bandwidth/bandwidth.js
```diff
@@ -0,0 +1,11 @@
...
+
+var deployment = new createDeployment({namespace: "shwang"});
```
secant: i'd opt for the default namespace

quilt-tester/tests/75-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,141 @@
...
+	wg := new(sync.WaitGroup)
+	resultsChan := make(chan testResult, len(containers))
+
+	// Simultaneously run iperf tests from each container i to container i+1.
+	for i, client := range containers {
+		if client.IP == "" {
+			log.WithError(err).Fatalf("FAILED, %s has no private IP", client)
+		}
+
+		server := containers[(i+1)%len(containers)]
```
secant: why are you adding one and modding here?

quilt-tester/tests/75-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,141 @@
...
+
+		wg.Add(1)
+		go func(client, server db.Container) {
+			test(client, server, resultsChan)
+			wg.Done()
```
secant: (just style) but I would `defer wg.Done()`.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,141 @@
...
+		}(client, server)
+	}
+
+	wg.Wait()
+	close(resultsChan)
```
secant: I don't think you need to close the channel, since you should be sending `len(containers)` worth of `testResults`.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,141 @@
...
+
+	var failed bool
+	for res := range resultsChan {
+		fmt.Println("client:\t", res.client)
```
secant: Use `log` (this comment applies up to line 83)

quilt-tester/tests/75-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,141 @@
...
+
+func parseBandwidth(output string) (int, error) {
+	lines := strings.Split(output, "\n")
+	if len(lines) < 3 {
+		return -1, errors.New(
+			"parsing error: expected at least 3 lines in output")
+	}
+
+	// Use receiver's reported bandwidth. We could instead choose the sender's
+	// result, which is outputted one line before the receiver's result.
```
secant: might be worth putting a comment of what an expected `output` looks like.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,141 @@
...
+	receiverFields := strings.Fields(lines[len(lines)-3])
+	if len(receiverFields) < 4 {
+		return -1, errors.New(
+			"parsing error: expected at least 4 fields in receiver result")
+	}
+
+	bandwidthStr := receiverFields[len(receiverFields)-4]
+	bandwidth, err := strconv.Atoi(bandwidthStr)
+	if err != nil {
+		return -1, errors.New("parsing error: " + err.Error())
```
secant: consider `fmt.Errorf`

quilt-tester/tests/75-bandwidth/check_bandwidth.go
sirspinach: The `len(containers)-1`th container should connect to the `0`th container. I think I'll add a comment.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
sirspinach: hm... I added this in because I thought `for res := range resultChan` might block if I didn't `close` the chan. I'll try omitting this guy later.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
sirspinach: IIRC, the other tests use `fmt.Println` for their test output (as opposed to their log output). I'll check this.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
sirspinach: good idea

quilt-tester/tests/75-bandwidth/check_bandwidth.go
sirspinach: I don't feel strongly about this, but I think ejj told matt to make the opposite change once (because appending is simpler than Printf in this case), so I'm leaving this error message as is.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
secant: I looked at the [spark](https://github.com/NetSys/quilt/blob/master/quilt-tester/tests/20-spark/check_spark.go) test

quilt-tester/tests/75-bandwidth/check_bandwidth.go
secant: i don't feel too strongly about it either, but with `Errorf` you don't need to call `Error()`

sirspinach:
directory needs to be renamed to `11-bandwidth` or `15-bandwidth`

quilt-tester/tests/75-bandwidth/check_bandwidth.go
kklin: I like `fmt.Println` because the output is a little easier to read.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
kklin: Good point re checking all the hostnames (btw, I personally don't care about modularity as much for the integration tests.. I'd much rather have coverage).

quilt-tester/tests/75-bandwidth/check_bandwidth.go
kklin: Ah good point

quilt-tester/tests/75-bandwidth/bandwidth.js
kklin: I'll make a card

quilt-tester/tests/75-bandwidth/bandwidth.js
```diff
@@ -0,0 +1,11 @@
...
+deployment.deploy(infrastructure);
+
+var c = new Container("networkstatic/iperf3", ["-s"]);
+var iperf = new Service("iperf", c.replicate(5));
```
kklin: You know, it'd be a good chance to flex the language a bit and create exactly `nWorkers` that are exclusively placed on all machines. Maybe then create another service to guarantee that there's one machine that has two iperfs so that we can measure intra-machine bandwidth.

I get `nWorkers` like this in another spec:
```
var nWorker = deployment.machines.filter(function(m) {
    return m.role == "Worker"
}).length;
```

quilt-tester/tests/75-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,141 @@
...
+	"sync"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/NetSys/quilt/api"
+	"github.com/NetSys/quilt/api/client/getter"
+	"github.com/NetSys/quilt/db"
+)
+
+const requiredBandwidthMbPerSec = 100
```
kklin: Can we also assert intra-machine bandwidth?

quilt-tester/tests/75-bandwidth/bandwidth.js
sirspinach: good idea

quilt-tester/tests/75-bandwidth/check_bandwidth.go
sirspinach: Needed `close` here.

quilt-tester/tests/75-bandwidth/check_bandwidth.go
sirspinach: fixed

sirspinach:
Added check for intra-machine bandwidth. When two iperf containers on the same machine were sending traffic to each other bidirectionally, the average bandwidth was about `4100 MB/s`.

I set the required intramachine bandwidth to be 2000 MB/s.

sirspinach:
ready for more review

quilt-tester/tests/15-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+
+	for _, grouped := range minionIPToContainers {
+		if len(grouped) == 1 {
+			interContainers = append(interContainers, grouped[0])
+		} else if len(grouped) > 1 {
+			intraContainers = append(intraContainers, grouped)
+		}
+	}
```
secant: Are you writing this test such that more than one machine will have >1 container? You could add more containers to the test spec then, since the current spec only gives 1 extra container.

secant: Also, instead of splitting them up into separate lists, you could just run the specific tests while you iterate through the map.
```
for _, grouped := range minionToCont {
    run intercontainer test
    if len(grouped) > 1 {
        run intracontainer test
    }
}
```

In fact, since you are closing the channel, you don't even need a buffered channel since the waitgroup does all the syncing.

quilt-tester/tests/15-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,207 @@
...
+			failed = true
+			fmt.Printf("FAILED, expected at least %d Mb/s, got %d Mb/s\n",
+				required, res.bandwidthMbPerSec)
+			fmt.Println("iperf output follows:")
+			fmt.Println(res.iperfOutput)
+			continue
+		}
+
+		fmt.Printf("Average bandwidth: %d Mb/s\n", res.bandwidthMbPerSec)
+		fmt.Println()
```
secant: Just add another `\n` to the previous line?

quilt-tester/tests/15-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,207 @@
...
+	for i, client := range containers {
+
+		// The (len(containers) - 1)th container should connect to the 0th
+		// container rather than the nonexistent (len(container))th container.
+		server := containers[(i+1)%len(containers)]
+
+		wg.Add(1)
+		go func(client, server db.Container) {
+			defer wg.Done()
+			test(client, server, results)
```
secant: here you can just get result back from `test` and pass it into the results chan instead of making `test` do it.

I guess isolating the parallelism doesn't apply as much here since you don't test the test code, but it's useful for unit testing.

secant: It'll also make the last parts of your `test` function a little less indented

quilt-tester/tests/15-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,207 @@
...
+	for _, grouped := range minionIPToContainers {
+		if len(grouped) == 1 {
+			interContainers = append(interContainers, grouped[0])
+		} else if len(grouped) > 1 {
+			intraContainers = append(intraContainers, grouped)
+		}
+	}
+
+	wg := new(sync.WaitGroup)
+	resultsChan := make(chan testResult, len(containers))
```
secant: If you don't implement the second comment on line 76, I would make the results channel in `runTests` and return the results as a list.

secant: My reasoning is because channels are generally for syncing data pertaining to some task; but here the intercontainer and intracontainer tests are split up so it feels weird /shrug.

sirspinach: Good point here. After reading your comment, I think this part looks a bit strange to me too.

quilt-tester/tests/15-bandwidth/check_bandwidth.go
sirspinach: good point, fixed.

quilt-tester/tests/15-bandwidth/check_bandwidth.go
sirspinach: first comment: 
I decided to have only one machine measure intra-machine traffic. I'm trying to leave the test's assumptions about the container placement flexible so that the spec can easily change in the future.

second comment:
I think my explanation/variable names are not clear. Renaming. 

It's possible to run the tests on the intra-machine containers within the loop because they are already together in the appropriate slice. However, the inter-machine containers are not in the same slice until after the loop. I'm leaving this loop as is since inter-machine traffic can't be tested until after the loop, and I think it is slightly cleaner to have all the tests launch at the same time.

quilt-tester/tests/15-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,213 @@
...
+
+	for i, client := range containers {
+		// The (len(containers) - 1)th container should connect to the 0th
+		// container rather than the nonexistent (len(container))th container.
+		server := containers[(i+1)%len(containers)]
+
+		wg.Add(1)
+		go func(client, server db.Container) {
+			defer wg.Done()
+			resultsChan <- test(client, server)
```
sirspinach: I think I'll keep both the buffered channel and the `WaitGroup`. `resultsChan <- test(client, server)` would block if `resultChan` were not buffered. And without the `WaitGroup` the program would not know when's the right time to `close(resultsChan)`. I am pretty sure it is necessary to `close(resultsChan)` because `for res := range resultsChan {` does not terminate until resultsChan is empty and resultsChan is closed.

secant: Yeah you're right, I made a wrong assumption about an unbuffered channel.

sirspinach:
@secant, is there anything else I can do to improve this patch?

quilt-tester/tests/15-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	// Use receiver's reported bandwidth. We could instead choose the sender's
+	// result, which is outputted one line before the receiver's result.
+	receiverFields := strings.Fields(lines[len(lines)-4])
+
+	var bandwidthStr string
+	for i, field := range receiverFields {
+		if field == "Mbits/sec" && i > 0 {
+			bandwidthStr = receiverFields[i-1]
```
sirspinach: good idea

quilt-tester/tests/15-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,213 @@
...
+	minionIPToContainers := make(map[string][]db.Container)
+	lonelyContainers := []db.Container{}
+	groupedContainers := [][]db.Container{}
+
+	for _, dbc := range containers {
+		if dbc.IP == "" {
+			log.WithError(err).Fatalf("FAILED, %s has no private IP", dbc)
+		}
+
+		if containers, ok := minionIPToContainers[dbc.Minion]; !ok {
```
sirspinach: Oh, cool! I thought accessing minionContainers[dbc.Minion] would error. https://play.golang.org/p/7G3lwiuozO

quilt-tester/tests/15-bandwidth/check_bandwidth.go
ejj: Iperf has json output.  Why not just do that?  Trivial to parse then.  If that doesn't work for some reason, you should at least use a regexp

quilt-tester/tests/15-bandwidth/check_bandwidth.go
```diff
@@ -0,0 +1,213 @@
...
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/NetSys/quilt/api"
+	"github.com/NetSys/quilt/api/client/getter"
+	"github.com/NetSys/quilt/db"
+)
+
+// The required bandwidth in Mb/s between two containers on different machines.
+const requiredBandwidthIntermachine = 100
```
ejj: Even this feels a bit high to me for such small VMs on shared infrastructure.  I'd say 50

quilt-tester/tests/15-bandwidth/check_bandwidth.go
sirspinach: non-json output because: https://github.com/NetSys/quilt/pull/680#discussion_r94362549

I'll try replacing this for loop with regex.

sirspinach:
changed `requiredBandwidthIntermachine` from 100 Mb/s to 50 Mb/s.

changed parsing function to use regexp.

PR #681, opened 2017-01-05
Use DNS instead of /etc/hosts

ejj:


minion/network/dns.go
```diff
@@ -0,0 +1,192 @@
+package network
```
kklin: In the commit message, "_we_ were handling"

minion/network/dns.go
```diff
@@ -0,0 +1,192 @@
...
+		table.recordLock.Lock()
+		table.records = records
+		table.recordLock.Unlock()
+		return table
+	}
+	table = makeTable(records)
+
+	// There could me multiple messages depending on how listenAndServe is
+	// implemented.  We don't want anyone to block, so we make a bit of a buffer.
+	errChan := make(chan error, 8)
```
kklin: I don't understand how this channel is used. From what I understand, `errChan` could potentially have one message from `NotifyStartedFunc`, and one message from `listenAndServe` returning. But if `listenAndServe` exits with an error after the server starts, nothing happens. Is this to handle when the server doesn't start, and `listenAndServe` errors?

minion/network/network.go
```diff
@@ -32,13 +31,15 @@ type dbport struct {
 type dbslice []dbport
 
 // Run blocks implementing the network services.
-func Run(conn db.Conn, dk docker.Client) {
+func Run(conn db.Conn) {
+
```
kklin: Is this necessary?

minion/network/network.go
```diff
@@ -32,13 +31,15 @@ type dbport struct {
 type dbslice []dbport
 
 // Run blocks implementing the network services.
-func Run(conn db.Conn, dk docker.Client) {
```
kklin: Nice 😄

minion/network/worker.go
```diff
@@ -958,12 +952,20 @@ func getPublicInterface() (string, error) {
...
+
+	logAction := "additions"
 	args := []string{"add-flows", quiltBridge, "-"}
 	if !add {
 		args = []string{"del-flows", "--strict", quiltBridge, "-"}
+		logAction = "deletions"
 	}
 	cmd := exec.Command("ovs-ofctl", args...)
 
+	log.Infof("%d OpenFlow %s", len(flows), logAction)
```
ejj: So one could make an argument either way.  The reason I chose to do it on Info, was we could easily have a bug where we constantly add and remove all of the OpenFlow rules on each run execution.  And the nasty thing about that bug is we wouldn't notice it because usually the traffic would run fine except for the occasional blip.   Doing it at Info means t that at least we'd have a large number of suspicious looking log messages.

All of that said, I really don't feel strongly about it.  Your call, I could switch it back to debug.

vendor/github.com/miekg/dns/AUTHORS
```diff
@@ -0,0 +1 @@
+Miek Gieben <miek@miek.nl>
```
ejj: I dunno, some people do it one way, some the other.  I suppose we need to just pick a style and stick with it.

The advantage of doing it in a separate commit, is that sometimes, when the commit is really huge, github can get upset.  I've had the code review window basically fail on me.  Doubt that would happen with this commit, but I like the idea of being consistent.  Anyways, warrants further discussion ...

minion/network/dns.go
```diff
@@ -0,0 +1,192 @@
...
+	records := labelsToDNS(labels)
+	if table != nil {
+		table.recordLock.Lock()
+		table.records = records
+		table.recordLock.Unlock()
+		return table
+	}
+	table = makeTable(records)
+
+	// There could me multiple messages depending on how listenAndServe is
```
ejj: Jeez, gotta work on that  . . .

minion/network/dns.go
ejj: Ah yes, so this is a tad fragile, but I read the dns code and they never call NotifyStartedFunc on a server that's going to fail.  It's a bit fragile, but I don't think the library gives us a good way to deal with the error and this is better than nothing.

minion/network/dns.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	tbl.server.Handler = tbl
+	return tbl
+}
+
+func labelsToDNS(labels []db.Label) map[string]net.IP {
+	records := map[string]net.IP{}
+	for _, label := range labels {
+		if ip := net.ParseIP(label.IP); ip != nil {
+			records[label.Label+".q."] = ip
```
ejj: Meh, feels like overkill in this case.

minion/network/network.go
ejj: You mean the newline?

kklin: Yeah

minion/network/dns.go
kklin: Fair enough

minion/network/worker.go
kklin: I'm fine leaving it at `Info` for now.

I do think the right thing long-term is to run the integration tests at the debug level, and do some sort of line-length test.

PR #683, opened 2017-01-05
quiltctl: Show whether container is up

sirspinach:


quiltctl/command/command_test.go
```diff
@@ -102,15 +102,16 @@ func TestContainerOutput(t *testing.T) {
...
-4_____Machine-5____image3_cmd________label1____________7.7.7.7:80
-_______________________________________________________
-7_____Machine-6____image1_cmd_3_4____label1____________
-_______________________________________________________
-8_____Machine-7____image1______________________________
+	expected := `ID____MACHINE______CONTAINER_________LABELS____________` +
+		`UP_______PUBLIC_IP
+3__________________image1_cmd_1________________________true_____
+________________________________________________________________
+1_____Machine-5____image2____________label1,_label2____false____7.7.7.7:80
```
ejj: I actually think we should change this a bit.  We should change the column name to be "Status".  And we should have three statuses, the empty string, "Scheduled" for those containers that have been assigned a minion, and "Running" for those containers that have an IP.

sirspinach: fixed

PR #684, opened 2017-01-06
amazon: Change wait to require machines to be instantiated

mmussomele:
When listing machines on Amazon, it was possible to get a response
between the assignment of cloud IDs but before fields such as the
machine's IP and size were filled in on Amazon's end. This caused
the cluster's join to fail, booting twice as many machines as needed.
The Amazon wait() function now requires machines to have both a
cloud ID and a size before returning.


cluster/amazon/amazon.go
```diff
@@ -362,20 +362,32 @@ OuterLoop:
 		}
 
 		exists := make(map[awsID]struct{})
+		ready := make(map[awsID]struct{})
```
kklin: I think we could just have one map for keeping track of existence and ready

cluster/amazon/amazon.go
```diff
@@ -363,11 +363,17 @@ OuterLoop:
...
+			// twice.
+			if inst.Size == "" {
+				continue
+			}
+
 			id := awsID{
 				spotID: inst.ID,
 				region: inst.Region,
 			}
-
```
sirspinach: Not sure if you meant to remove this whitespace :)

cluster/amazon/amazon.go
```diff
@@ -363,11 +363,17 @@ OuterLoop:
```
sirspinach: The first sentence of the commit message, "When listing machines on Amazon, it was possible to get a response between the assignment of cloud IDs but before fields such as the machine's IP and size were filled in on Amazon's end", took me a while to parse. Do you think it would be worthwhile to reword it more simply or break it up into two sentences?

cluster/amazon/amazon.go
```diff
@@ -363,11 +363,17 @@ OuterLoop:
 
 		exists := make(map[awsID]struct{})
 		for _, inst := range machines {
+			// If the machine wasn't configured completely when the List()
+			// call was made, the cluster will fail to join and boot them
+			// twice.
+			if inst.Size == "" {
+				continue
```
sirspinach: Do you think it would be worth the effort to add coverage for this case?

sirspinach: In case you want to add even more coverage to this function, I've uploaded the cov.html file for reference: http://tilde.town/~shwang/amazon.cov.html#file0

cluster/amazon/amazon.go
mmussomele: Yea I'll reword it

cluster/amazon/amazon.go
mmussomele: I'm not sure it's worth the effort to cover this statement, as it's pretty simple. Plus, once the cluster series I'm working on is in, the cluster will do the `wait`ing, so I think it makes more sense to wait and cover that well since this code will be gone soon enough.

kklin: Could you write a test that would've failed before this fix?

cluster/amazon/amazon.go
mmussomele: Yea, it looked awkward to me, as I like to group things by related logic.

cluster/amazon/amazon.go
mmussomele: Ah true, perhaps that's what @sirspinach meant. I'll do that tonight.

mmussomele:
Updated with a test, confirmed to fail before patch.

sirspinach:
LGTM!

PR #688, opened 2017-01-08
Migrate to codecov.io

ejj:


tejasmanohar:
789e9dcee7d320a174c9405139d6b219f47c50b9 is separate from codecov migration, right?

Either way, LGTM!

ejj:
@tejas, something about the coverage changes was causing the unit tests to fail due to that issue.

PR #689, opened 2017-01-08
Codecov Tweaks

ejj:


.codecov.yaml
```diff
@@ -24,3 +24,6 @@ parsers:
       method: false
   javascript:
     enable_partials: false
+ignore:
```
kklin: What about the stuff here? https://github.com/NetSys/quilt/blob/master/Makefile#L31

BTW, can we remove that variable now that we're using codecov?

ejj: yeah so I tried to remove that variable, but if we do the actual `make coverage` target fails.  So I was thinking that we leave it in and cover everything we can with the command line utility, and then filter out when importing to codecov.

The tooling around all this sucks =(

PR #690, opened 2017-01-09
Floating IPs

tejasmanohar:
@ejj


ejj:
Btw, I think there's a lot of low hanging code coverage fruit, notably in google.go

ejj:
So one fairly high level comment.  Something feels slightly wonky about the API, but I can't quite put my finger on it.  I think it might be better if you combined the Assign and Remove functions into a single `UpdateFloatingIPs` function.  For each machine, if the floating IP was "" the function would remove the floating IP, and if it was not "" it would assign it.  The awkward thing about this is you're moving the join into the cloud provider which doesn't seem ideal.

Another option -- Have a function `AssignFloatingIP(m machine.Machine, ip string)` and it's remove analog that does the assignment for a single machine and explicitly takes the IP that you intend to assign.  At least that way it's not buried implicitly in the machine.Machine.

What do you think?  Something about the API feels off to me, but I'm not quite sure what it is.

util/util_test.go
```diff
@@ -80,6 +81,33 @@ func TestEditDistance(t *testing.T) {
 	}
 }
 
+func TestWaitFor(t *testing.T) {
```
ejj: and I assume this will be pulled out of the PR?

cluster/amazon/amazon.go
```diff
@@ -308,14 +325,122 @@ func (clst Cluster) List() ([]machine.Machine, error) {
 	return machines, nil
 }
 
-// AssignFloatingIPs is not supported.
+// AssignFloatingIPs assigns Elastic IPs to machines.
 func (clst *Cluster) AssignFloatingIPs(machines []machine.Machine) error {
```
ejj: Btw, one sign we don't have the API quite right ... there's a lot of duplicate code between the Assign and Remove functions.

cluster/amazon/amazon.go
```diff
@@ -308,14 +325,122 @@ func (clst Cluster) List() ([]machine.Machine, error) {
...
+		for _, machine := range machines {
+			spotIDs = append(spotIDs, machine.ID)
+		}
+		instances, err := clst.getReadyInstances(region, spotIDs)
+		if err != nil {
+			return err
+		}
+
+		for _, machine := range machines {
+			if alloc := availableAddresses[machine.FloatingIP]; alloc != nil {
```
ejj: Do we really need to do this check?  If the address is available will the ec2 client fail on its own?

cluster/amazon/amazon.go
```diff
@@ -308,14 +325,122 @@ func (clst Cluster) List() ([]machine.Machine, error) {
...
+		associations := map[string]*string{}
+		for _, address := range addressDesc.Addresses {
+			if address.InstanceId != nil {
+				associations[*address.InstanceId] = address.AssociationId
+			}
+		}
+
+		for _, machine := range machines {
+			instanceID := *instances[machine.ID].InstanceId
+			if assoc := associations[instanceID]; assoc != nil {
```
ejj: Similar question here:  Do we really need to check the association?  It will just fail if its not associated right?

cluster/amazon/amazon.go
```diff
@@ -308,14 +325,122 @@ func (clst Cluster) List() ([]machine.Machine, error) {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// getReadyInstances() polls getInstances() until all instances corresponding to
+// a set of spot IDs are ready.
+func (clst Cluster) getReadyInstances(region string, spotIDs []string) (
```
ejj: This should be in a different commit definitel

ejj: definitely

cluster/amazon/amazon.go
```diff
@@ -326,6 +451,45 @@ func (clst Cluster) getClient(region string) client {
 	return clst.clients[region]
 }
 
+func (clst Cluster) getInstances(region string, spotIDs []string) (
```
ejj: Same with this, should be in its own commit before the floating IP addition I think

tejasmanohar:
@ejj I like a separate add/remove. Does the following sound good to you?
- AssignFloatingIP(machine.Machine, ip string)
- RemoveFloatingIP(machine.Machine)

I'm also open to changing `machine.Machine` -> `id string` for both functions since the provider should be able to work off of that. The reason I don't really like combining add/remove is I think it's awkward since there are lots of states of `machine.Machine` (e.g. the DB has no floating IP for machine X which we pass to the "update" function-- intent is not clear).

cluster/amazon/amazon.go
tejasmanohar: Probably. I'm for handling it via `else { ... log stuff so user doesn't see wonky AWS error ... }` rather than letting AWS fail... Something like "%s is not a valid AWS Elastic IP address in Region %s". Thoughts?

cluster/amazon/amazon.go
tejasmanohar: Hm, why shouldn't it be in the AWS Floating IPs commit? I find it a bit strange to add an unused function in a preceding commit... but don't feel strongly about this TBH.

ejj: Oh is floating IPs the only user?  Seems like something that could be useful more widely in the amazon cluster

tejasmanohar: Ah yeah, this is the only caller rn. getInstances may be able to be used in List(), whereas getReadyInstances won't since the condition doesn't match... will check

util/util_test.go
tejasmanohar: fixed

cluster/amazon/amazon.go
tejasmanohar: I don't think Quilt should fail the disassociation (or even log an error) if the machine has no Elastic IP associated with it for whatever reason. It's possible it's been re-assigned by this point by Quilt or even manually.

tejasmanohar:
@ejj I think splitting `machine` and `ip` into separate arguments in `AssignFloatingIP(m machine.Machine, ip string)` is weird because the cluster would have already updated the `FloatingIP` of the machine before the method is called. That's why I chose to combine them. Do you still think it's nicer that way? What are your concerns with the current methods?

tejasmanohar:
After some refactoring, I decided to go with `UpdateFloatingIPs([]machine.Machine)`. I started with the other approach but found the code less intuitive, mainly because
- separating assign & remove in cluster introduces more ordering concerns that the cluster alone doesn't have enough context about... for example, user requests IP to be updated from X -> Y, do we assign or remove first? ...or do we just assign? only the provider knows
- `UpdateFloatingIPs([]machine.Machine)` signature is more similar to other provider methods

tejasmanohar:
@ejj I have finished incorporating your feedback and will rebase and add tests for (GCE, Floating IPs) once https://github.com/NetSys/quilt/pull/706 is reviewed, revised, and merged to `master`.

tejasmanohar:
@ejj Rebased.

cluster/cluster.go
```diff
@@ -116,23 +129,19 @@ func (clst cluster) runOnce() {
...
-		clst.updateCloud(jr.boot, true)
-		clst.updateCloud(jr.terminate, false)
+		clst.updateCloud(jr.boot, boot)
+		clst.updateCloud(jr.terminate, stop)
+		clst.updateCloud(jr.updateIPs, updateIPs)
 	}
 }
 
-func (clst cluster) updateCloud(machines []machine.Machine, boot bool) {
+func (clst cluster) updateCloud(machines []machine.Machine, todo action) {
```
ejj: I think I'd rather this variable be named `act` or something.  Whenever I read `todo` my brain thinks "unfinished code".

cluster/cluster.go
```diff
@@ -320,7 +350,7 @@ func syncDB(cloudMachines []machine.Machine, dbMachines []db.Machine) syncDBResu
 		case dbm.PrivateIP == m.PrivateIP:
 			return 2
 		default:
-			return 3
+			return 4
```
ejj: Why?  This change should be a noop?

cluster/amazon/amazon.go
```diff
@@ -308,9 +326,98 @@ func (clst Cluster) List() ([]machine.Machine, error) {
...
+		}
+
+		for _, inst := range instances {
+			if inst == nil ||
+				*inst.State.Name != ec2.InstanceStateNameRunning {
+				return false
+			}
+		}
+		return true
+	}, 10*time.Second, timeout)
```
ejj: So this could take up to 15 minutes??? That seems way too long.  I'm not even sure we need the WatiFor at all here, why not just try once, if we fail return an error, and then the cluster run loop will try again later?  Seems odd to block in the provider for this long.

ejj:
I'm happy with this overall.  Just the one big question about the amazon spec

cluster/cluster.go
tejasmanohar: changed. no longer noop!

cluster/cluster.go
tejasmanohar: lol alright changed

cluster/amazon/amazon.go
tejasmanohar: actually ya, let's leave it to the cluster.

tejasmanohar: Tejass-MacBook-Pro:quilt tejas$ git diff
diff --git a/cluster/amazon/amazon.go b/cluster/amazon/amazon.go
index 629da3b8..43414310 100644
--- a/cluster/amazon/amazon.go
+++ b/cluster/amazon/amazon.go
@@ -351,7 +351,7 @@ func (clst *Cluster) UpdateFloatingIPs(machines []machine.Machine) error {
                for _, machine := range machines {
                        spotIDs = append(spotIDs, machine.ID)
                }
-               instances, err := clst.getReadyInstances(region, spotIDs)
+               instances, err := clst.getInstances(region, spotIDs)
                if err != nil {
                        return err
                }
@@ -387,39 +387,6 @@ func (clst *Cluster) UpdateFloatingIPs(machines []machine.Machine) error {
        return nil
 }

-// getReadyInstances() polls getInstances() until all instances corresponding to
-// a set of spot IDs are ready.
-func (clst Cluster) getReadyInstances(region string, spotIDs []string) (
-       map[string]*ec2.Instance, error) {
-       var instances map[string]*ec2.Instance
-
-       var innerError error
-       err := util.WaitFor(func() bool {
-               instances, innerError = clst.getInstances(region, spotIDs)
-               if innerError != nil {
-                       return true
-               }
-
-               for _, inst := range instances {
-                       if inst == nil ||
-                               *inst.State.Name != ec2.InstanceStateNameRunning {
-                               return false
-                       }
-               }
-               return true
-       }, 10*time.Second, timeout)
-
-       if innerError != nil {
-               return nil, innerError
-       }
-
-       if err != nil {
-               return nil, err
-       }
-
-       return instances, nil
-}
-
 func (clst Cluster) getClient(region string) client {
        if _, ok := clst.clients[region]; !ok {
                clst.clients[region] = clst.newClient(region)

tejasmanohar:
I think it should be ready to go now :D

tejasmanohar:
@ejj Have you reviewed this since https://github.com/NetSys/quilt/pull/690#issuecomment-273687896? is it ready to merge?

ejj:
@tejas, yes it's ready.  Just needs to be rebased (the unit tests fail)

tejasmanohar:
pushed a small fix for GCE after rebasing
```diff
$ git diff
diff --git a/cluster/google/google.go b/cluster/google/google.go
index c9848def..27daf782 100644
--- a/cluster/google/google.go
+++ b/cluster/google/google.go
@@ -40,6 +40,12 @@ import (
 // region preference.
 const DefaultRegion = "us-east1-b"

+// ephemeralIPName is a constant for what we label NATs with ephemeral IPs in GCE.
+const ephemeralIPName = "External NAT"
+
+// floatingIPName is a constant for what we label NATs with floating IPs in GCE.
+const floatingIPName = "Floating IP"
+
 const computeBaseURL string = "https://www.googleapis.com/compute/v1/projects"
 const (
        // These are the various types of Operations that the GCE API returns
@@ -114,16 +120,21 @@ func (clst *Cluster) List() ([]machine.Machine, error) {
                        // XXX: This make some iffy assumptions about NetworkInterfaces
                        machineSplitURL := strings.Split(item.MachineType, "/")
                        mtype := machineSplitURL[len(machineSplitURL)-1]
+
+                       accessConfig := item.NetworkInterfaces[0].AccessConfigs[0]
+                       floatingIP := ""
+                       if accessConfig.Name == floatingIPName {
+                               floatingIP = accessConfig.NatIP
pick 6a687587 cluster: Add floating IP machine rule
+                       }
+
                        mList = append(mList, machine.Machine{
-                               ID: item.Name,
-                               PublicIP: item.NetworkInterfaces[0].
-                                       AccessConfigs[0].NatIP,
-                               FloatingIP: item.NetworkInterfaces[0].
-                                       AccessConfigs[0].NatIP,
-                               PrivateIP: item.NetworkInterfaces[0].NetworkIP,
-                               Size:      mtype,
-                               Region:    zone,
-                               Provider:  db.Google,
+                               ID:         item.Name,
+                               PublicIP:   accessConfig.NatIP,
+                               FloatingIP: floatingIP,
+                               PrivateIP:  item.NetworkInterfaces[0].NetworkIP,
+                               Size:       mtype,
+                               Region:     zone,
+                               Provider:   db.Google,
                        })
                }
        }
@@ -293,7 +304,7 @@ func (clst *Cluster) instanceNew(name string, size string, zone string,
                                AccessConfigs: []*compute.AccessConfig{
                                        {
                                                Type: "ONE_TO_ONE_NAT",
-                                               Name: "External NAT",
+                                               Name: ephemeralIPName,
                                        },
                                },
                                Network: fmt.Sprintf("%s/global/networks/%s",
@@ -441,7 +452,7 @@ func (clst *Cluster) UpdateFloatingIPs(machines []machine.Machine) error {
                _, err = service.Instances.AddAccessConfig(clst.projID, m.Region, m.ID,
                        networkInterface.Name, &compute.AccessConfig{
                                Type: "ONE_TO_ONE_NAT",
-                               Name: "External NAT",
+                               Name: floatingIPName,
                                // Google will automatically assign a dynamic IP
                                // if none is provided (i.e. m.FloatingIP == "").
                                NatIP: m.FloatingIP,
```

tejasmanohar:
alright, i think it's ready :D https://github.com/NetSys/quilt/pull/690#issuecomment-275720398

cluster/cluster.go
```diff
@@ -116,23 +129,19 @@ func (clst cluster) runOnce() {
...
 	}
 
-	actionString := "halt"
-	if boot {
-		actionString = "boot"
-	}
-
 	log.WithField("count", len(machines)).
-		Infof("Attempt to %s machines.", actionString)
+		Infof("Attempt to %s machines.", act)
```
kklin: This prints out funny stuff because `act` isn't a string.

cluster/cluster.go
```diff
@@ -144,20 +153,38 @@ func (clst cluster) updateCloud(machines []machine.Machine, boot bool) {
...
+
 		if err != nil {
 			noFailures = false
-			log.WithError(err).
-				Warnf("Unable to %s machines on %s.", actionString, p)
+			switch act {
+			case boot:
+				log.Infof("Unable to boot machines on %s.", p)
+			case stop:
+				log.Infof("Successfully stopped machines on %s", p)
```
kklin: This shouldn't be successful right?

cluster/cluster.go
```diff
@@ -144,20 +153,38 @@ func (clst cluster) updateCloud(machines []machine.Machine, boot bool) {
...
+			case updateIPs:
+				log.Infof("Successfully updated floating IPs on %s", p)
```
kklin: Same here.

tejasmanohar:
my bad @kklin pr'd #734

PR #693, opened 2017-01-09
util: Refactor quilt-tester to use util.WaitFor

secant:
This generalizes the `waitFor()` function from quilt-tester/exec
to accept an arbitrary timeout and migrates it to the `util` package
so that other packages can make use of it.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -3,14 +3,19 @@ package util
 import (
 	"archive/tar"
 	"bytes"
+	"errors"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/spf13/afero"
 )
+
+// Sleep stores time.Sleep so we can mock it out for unit tests.
+var Sleep = time.Sleep
 
 func httpRequest(url string) (string, error) {
 	resp, err := http.Get(url)
@@ -143,4 +148,19 @@ func StrStrMapEqual(x, y map[string]string) bool {
 		}
 	}
 	return true
+}
+
+// WaitFor waits until `pred` is satisfied, or `timeout` Duration has passed, checking
+// at every `interval`.
+func WaitFor(pred func() bool, interval time.Duration, timeout time.Duration) error {
+	deadline := time.Now().Add(timeout)
+	for {
+		if pred() {
+			return nil
+		}
+		if time.Now().After(deadline) {
+			return errors.New("timed out")
+		}
+		Sleep(interval)
+	}
 }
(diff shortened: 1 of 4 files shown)


util/util.go
```diff
@@ -144,3 +149,20 @@ func StrStrMapEqual(x, y map[string]string) bool {
 	}
 	return true
 }
+
+// WaitFor waits until `pred` is satisfied, or `timeout` Duration has passed, checking
+// at every `interval`.
+func WaitFor(pred func() bool, interval time.Duration, timeout time.Duration) error {
+	timeoutChan := time.After(timeout)
```
ejj: This feels like overkill to me.  Why not just do the check directly?  I record `deadline := time.Now() + timeout` and then check in the loop if time.Now() is after the deadline?

util/util.go
```diff
@@ -144,3 +149,18 @@ func StrStrMapEqual(x, y map[string]string) bool {
...
+
+// WaitFor waits until `pred` is satisfied, or `timeout` Duration has passed, checking
+// at every `interval`.
+func WaitFor(pred func() bool, interval time.Duration, timeout time.Duration) error {
+	deadline := time.Now().Add(timeout)
+	for {
+		if time.Now().After(deadline) {
+			return errors.New("timed out")
+		}
+		if pred() {
```
ejj: One last little nit.   I think you should check pred before you check the deadline.  That way if we sleep too long we don't end up failing the unit test for no reason.

util/util.go
```diff
@@ -144,3 +149,18 @@ func StrStrMapEqual(x, y map[string]string) bool {
...
+			return nil
+		}
+		Sleep(interval)
```
ejj: Btw, I wonder if you should sleep for Min(interval, time.Now() - deadline)?

tejasmanohar: IMO, the channel + select we had before was simpler / more idiomatic Go.

ejj: That old code had the same problem though . . . it was still possible to sleep way past the deadline.  I suppose one could have fixed with two channels, and After and a Ticker.

That said, I don't really agree.  Concurrency is hard to reason about, using a simple variable is really clear.  I think go makes it easy to use channels when necessary, but that doesn't mean it demands we use them when more straight forward approaches work.

ejj: I should say, @secant, I don't care that much about this second point.  I think the only thing that really matters is that the predicate check comes before the deadline check.  If you change that I'll merge.

secant:
@ejj changed.

ejj:
LGTM will merge once CI passes

PR #694, opened 2017-01-09
foreman: Properly log connection errors

kklin:
The previous code threw away all connection errors, not just connection
errors caused by the machine not being setup yet. Furthermore, it did so
in a very fragile way, relying on matching the text in the error string.
Thus, when the returned error changed after the grpc update, the previous
code broke.

We now only log connection errors at `Error` if we had previously
connected to the minion. Otherwise, we log at `Debug`.


ejj:
LGTM

ejj:
Waiting for CI, but if travis is happy I'll merge.

PR #702, opened 2017-01-11
quiltctl: Properly size PTYs

kklin:
We had the height and width arguments swapped in the PTY request. This
often resulted in text being prematurely wrapped when using `quilt
exec`.


ejj:
Ah this had really annoyed me

PR #704, opened 2017-01-16
quiltctl: Add flag to `stop` to destroy only containers

kklin:


quiltctl/command/run.go
```diff
@@ -143,16 +143,18 @@ func (rCmd *Run) Run() int {
 	return 0
 }
 
-func getCurrentDeployment(c client.Client) (string, error) {
+var errNoCluster = errors.New("no cluster")
```
sirspinach: seems a bit strange to me that `errNoCluster` is defined here when it is referenced earlier in the file.

quiltctl/command/stop_test.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func TestStopNamespaceDefault(t *testing.T) {
+	t.Parallel()
+
+	mockGetter := new(testutils.Getter)
+	c := &clientMock.Client{}
+	mockGetter.On("Client", mock.Anything).Return(c, nil)
+
+	c.ClusterReturn = []db.Cluster{
```
sirspinach: ah, I realize now that this is a pretty obvious thing to do, but I like how you assigned `c.ClusterReturn` after you made `mockGetter.Client()` return `c`. It's cleaner this way than it is in my unit tests.

quiltctl/command/run.go
kklin: Moved the variable, and removed a constant made dead by the commit.

PR #705, opened 2017-01-17
Various network Tweaks

ejj:
Most of these are small fixes in preparation for moving IP allocation to the master.

### DIFF
--- a/db/container.go
+++ b/db/container.go
@@ -12,17 +12,17 @@ import (
 // eventually be instantiated within its corresponding cluster.
 // Used only by the minion.
 type Container struct {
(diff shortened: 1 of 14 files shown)


db/container.go
```diff
@@ -12,17 +12,17 @@ import (
 // eventually be instantiated within its corresponding cluster.
 // Used only by the minion.
 type Container struct {
-	ID int // The database ID of this container.
```
tejasmanohar: why remove the comments?

ejj: They're kind of redundant, and they don't fit that well with the annotations.

PR #706, opened 2017-01-17
GCE Tests

tejasmanohar:
This is a prerequisite for testing GCE changes on https://github.com/NetSys/quilt/pull/690.


tejasmanohar:
Btw, the `vendor: ...` commit is before the `google: Test List()` commit in `git`, but the GitHub UI is displaying it differently (and incorrectly IMO) because it was modified in a rebase (I guess it records push time). https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/

tejasmanohar:
Haha wow, this PR failed the coverage check even though it increased overall coverage!

Should we exclude [cluster/google/client.go](https://github.com/tejasmanohar/quilt/blob/gce_tests/cluster/google/client.go) from coverage checks... or just ignore the patch coverage check for this PR... or {{ something else }}? If we're going to go to the extent of testing that those functions call GCE code correctly, I think it's worth rethinking how we wrap the GCE client (and probably make some complicated interfaces instead).

kklin:
The entire google cluster should be covered eventually, but I don't think it has to be done in this PR (unless you particularly want to 😄).

tejasmanohar:
@kklin Yeah, for sure... but not in this PR :P -- I just wanted to scaffold here so that we can add some Floating IP tests.

ejj:
Btw I'm fine with this failing codecov.  As long as it passes code review and travis, we can merge.

tejasmanohar:
Will test GCE manually tonight on this branch since there are no integration tests yet and it's a lot of changes

cluster/google/client.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (c *clientImpl) InsertInstance(in insertInstanceInput) (*compute.Operation, error) {
+	return c.gce.Instances.Insert(in.project, in.zone, in.instance).Do()
+}
+
+type deleteInstanceInput struct {
+	project  string
+	zone     string
+	instance string
```
Baisang: Shouldn't this be type `*compute.Instance`?

Baisang: just kidding, saw the relevant docs

cluster/google/google.go
```diff
@@ -72,12 +68,15 @@ type Cluster struct {
...
 func New(namespace string) (*Cluster, error) {
-	if err := gceInit(); err != nil {
-		log.WithError(err).Debug("failed to start up gce")
+	gce, err := newClient()
+	if err != nil {
+		log.WithError(err).Error("Failed to initialize GCE client")
 		return nil, err
 	}
 
+	// TODO: initialize homemade client
```
ejj: No TODOs in the code please

tejasmanohar: oops, meant to remove that in a later commit, will do

tejasmanohar:
> Why does the GCE client method chaining preclude the need for all inputs to be structs? I don't quite understand

Each wrapper function is incomplete. Later, we may want to add more optional fields like "filter", "maxResults", etc. to the input structs (as seen in `ListInstances`), and it'd be nice to be able to do this without breaking the API / changing usage of function in other places.

Baisang:
Ahh I see, that makes sense and is a good compromise!

cluster/google/google.go
tejasmanohar: removed

tejasmanohar:
Confirmed that I can still boot machines @ejj @Baisang

ejj:
I don't really agree about the struct thing.  Using straight function arguments is a lot more straight forward IMO.  I also don't really buy the argument that it's easier to change in future.  In fact, I kind of prefer function arguments in this respect, if you add or remove an argument, you *have* to update all the functions or else the compiler complains.

Anyways, IMO, it feels like premature generalization, I'd rather do the straight forward thing.

ejj:
Rest of the series looks fine, just want that one change.

Baisang:
I think that's a valid viewpoint as well, really just a question of more incremental work (fixing function calls when arguments are added/removed) vs. a larger amount of work later (ideally, converting from structs to normal function calls and arguments at some point in time)

tejasmanohar:
>a larger amount of work later (ideally, converting from structs to normal function calls and arguments at some point in time)

I don't think this ever should happen. I think our client should be a minimal wrapper over Google's, and I think having structs is way more readable than making everything ordered arguments (e.g. project, zone, instance, filter, maxresults)

tejasmanohar:
How about this? I'll put all the optional arguments in structs and all the required arguments in comma-separated... This is most similar to  google's builder pattern api

tejasmanohar:
Updated. Btw, I renamed the commit so the conversation / activity tab of this PR looks weird, but the commits tab doesn't have any duplicates so ignore that weirdness.

cc @ejj

ejj:
@tejas, I'm happy with this.  I just think you should squash the API wrapper to commit into the first commit that introduces the wrapper.  Otherwise its ready

tejasmanohar:
@ejj awesome, done...

tejasmanohar:
(noticed make lint is failing, i can fix it tonight)

ejj:
Ping

tejasmanohar:
Fixed linting @ejj -- sorry about delay

PR #711, opened 2017-01-19
specs: Implement a Django specification

Baisang:
A basic Django specification running with Gunicorn and MongoDB.

### DIFF
--- a/specs/django/Dockerfile
+++ b/specs/django/Dockerfile
@@ -1,0 +1,10 @@
+FROM python:2.7
+Maintainer Brian Sang <sang.bri@gmail.com>
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+RUN git clone https://github.com/baisang/django-polls-example.git .
+RUN pip install -r requirements.txt
+
+ENTRYPOINT ["gunicorn", "-w", "4", "-b", "0.0.0.0:80", "mysite.wsgi"]
--- a/specs/django/django-example.js
+++ b/specs/django/django-example.js
@@ -1,0 +1,29 @@
+var Django = require("github.com/NetSys/quilt/specs/django/django")
+var HaProxy = require("github.com/NetSys/quilt/specs/haproxy/haproxy").Haproxy;
+var Mongo = require("github.com/NetSys/quilt/specs/mongo/mongo");
+
+// Infrastructure
+var deployment = createDeployment({});
+
+var baseMachine = new Machine({
+  provider: "Amazon",
+  // sshKeys: githubKeys("CHANGE_ME"), // Replace with your GitHub username
+});
+
+// Applications
+var mongo = new Mongo(3);
+
+var django = new Django({
+  nWorker: 3,
+  image: "quilt/django-polls",
+}, mongo);
+
+var haproxy = new HaProxy(1, django.services());
+
+// Connections
+haproxy.public();
+
+// Deployment
+deployment.deploy(baseMachine.asMaster())
+deployment.deploy(baseMachine.asWorker().replicate(3))
+deployment.deploy([django, mongo, haproxy]);
--- a/specs/django/django.js
+++ b/specs/django/django.js
@@ -1,0 +1,39 @@
+// Specs for Django web service
+function Django(cfg, mongo) {
+  if (typeof cfg.nWorker !== 'number') {
+    throw new Error('`nWorker` is required');
+  }
+  if (typeof cfg.image !== 'string') {
+    throw new Error('`image` is required');
+  }
+  this._port = cfg.port || 80;
+  var env = cfg.env || {};
+  env.MONGO_URI = mongo.uri("django-example")
+
+  var containers = new Container(cfg.image).withEnv(env).replicate(cfg.nWorker);
+  this._app = new Service("app", containers);
+
+  mongo.connect(mongo.port(), this);
+  this.connect(mongo.port(), mongo);
+};
+
+Django.prototype.deploy = function(deployment) {
+  deployment.deploy(this.services());
+};
+
+Django.prototype.services = function() {
+  return [this._app];
+};
+
+Django.prototype.port = function() {
+  return this._port;
+};
+
+Django.prototype.connect = function(port, to) {
+  var self = this;
+  to.services().forEach(function(service) {
+    self._app.connect(port, service);
+  });
+};
+
+module.exports = Django;
--- a/specs/spec_test.go
+++ b/specs/spec_test.go
@@ -57,4 +57,5 @@ func TestConfigs(t *testing.T) {
 	testConfig("./mean/example.js", quiltPath)
 	testConfig("./elasticsearch/main.js", quiltPath)
 	testConfig("./kibana/main.js", quiltPath)
+	testConfig("./django/django-example.js", quiltPath)
 }


specs/django/docker-entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -e
+
+timestamp() {
```
ejj: In *theory* this shouldn't be necessary.  Try without it?

And if that's the case, I'm not sure we need the entrypoint at all, should be able to just call gunicorn directly?

specs/django/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:2.7
+Maintainer Brian Sang <sang.bri@gmail.com>
+
+RUN pip install Django==1.10.2
```
ejj: Because docker hates us, it's best to combine all of this until a single RUN statement.  It's a limitation of their weird layered file system

specs/django/django-example.js
```diff
@@ -0,0 +1,41 @@
+var Django = require("github.com/NetSys/quilt/specs/django/django")
```
kklin: We have a unit test somewhere that checks that specs compile, you should add this file to it

ejj:
Didn't read this carefully, just skimmed and commented on one or two things I noticed

specs/django/Dockerfile
Baisang: For this example I can probably just have it install things in the later statement `pip install -r requirements.txt` later, so I'll just get rid of these. Just remember reading somewhere that usually good practice to have those run statements separate from `pip install -r requirements.txt` so you can have intermediate docker build images

ejj: Yeah so there's somewhat of a tradeoff here. Having a bunch of intermediate docker images is useful for development because you can change things with less effort in the rebuild.  But it slows down boot significantly because it makes the compete container much larger.  I think in this case I prefer faster boot times.

specs/django/docker-entrypoint.sh
ejj: Oh, also please test it when you remove it

Baisang: Works without, will update PR

specs/django/django-example.js
```diff
@@ -0,0 +1,41 @@
...
+var HaProxy = require("github.com/NetSys/quilt/specs/haproxy/haproxy").Haproxy;
+var Mongo = require("github.com/NetSys/quilt/specs/mongo/mongo");
+
+// Infrastructure
+var deployment = createDeployment({
+    namespace: "CHANGE_ME",
```
ejj: I'd remove this line, most people should run the default namespace.

specs/django/django-example.js
```diff
@@ -0,0 +1,41 @@
...
+    adminACL: ["local"],
+});
+
+var baseMachine = new Machine({
+    provider: "Amazon",
+    region: "us-west-1",
+    size: "m3.medium",
+    sshKeys: githubKeys("ejj"), // Replace with your GitHub username.
```
ejj: Here, I think I'd change this to "CHANGE_ME"

ejj: Ah, or even better, leave it commented out.  The system will work if they can't ssh in, and they can always change it later.

specs/django/django-example.js
```diff
@@ -0,0 +1,41 @@
...
+// Infrastructure
+var deployment = createDeployment({
+    namespace: "CHANGE_ME",
+    adminACL: ["local"],
+});
+
+var baseMachine = new Machine({
+    provider: "Amazon",
+    region: "us-west-1",
+    size: "m3.medium",
```
ejj: I'd express the size in terms of RAM/CPU, that way it's easier to change provider if people want to.

specs/django/django-example.js
```diff
@@ -0,0 +1,40 @@
...
+
+// Applications
+var mongo = new Mongo(3);
+
+var django = new Django({
+  nWorker: 3,
+  image: "baisang/django-polls", // Replace with your own image, for example
+  env: {
+    PORT: "80",
+    MONGO_URI: mongo.uri("django-example")
```
kklin: I think we should pass in the mongo object. That way, `Django` can setup the connections to `mongo`.

specs/django/django.js
```diff
@@ -0,0 +1,42 @@
...
+
+  this._port = cfg.port || 80;
+
+  var env = cfg.env || {};
+  var containers = new Container(cfg.image).withEnv(env).replicate(cfg.nWorker);
+  this._app = new Service("app", containers);
+
+  var hosts = this._app.children();
+  for (var i = 0; i < containers.length; i++) {
+    containers[i].setEnv("HOST", hosts[i]);
```
kklin: Does your container actually use the `HOST` environment variable?

specs/django/django.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  this._app.connect(this._port, this._app);
```
kklin: Is this necessary?

specs/django/django-example.js
```diff
@@ -0,0 +1,40 @@
...
+
+// Infrastructure
+var deployment = createDeployment({
+    adminACL: ["local"],
+});
+
+var baseMachine = new Machine({
+    provider: "Amazon",
+    cpu: new Range(2),
+    ram: new Range(2),
```
kklin: Indentation doesn't match with below

specs/django/django-example.js
Baisang: Not quite sure what you mean, like we pass in the `mongo` object into the `django` constructor which then sets the environment variable for the django containers?

specs/django/django.js
Baisang: No, I'll remove this

specs/django/django.js
Baisang: So I wasn't too sure what this line did from the other examples, are you saying that because in this spec the `gunicorn` instance is listening on port 80 already that this isn't necessary? Or something else?

kklin: What this does is make it so each of the django workers can talk to each other on port 80. I don't think your app needs that.

The only necessary `connect` I think is `HaProxy` connecting to the django worker https://github.com/NetSys/quilt/blob/master/specs/haproxy/haproxy.js#L20.

specs/django/django-example.js
kklin: Ah, I meant passing in the `mongo` object so that you can setup these connections (https://github.com/NetSys/quilt/pull/711/files/3125a553ededc31654289d034924fc599701c7b8#diff-9c96ae336b6ef154ff611bc24615638dR33) inside the `Django` constructor.

Baisang: Oh I see. I still need to pass down the `MONGO_URI` env variable into the `django` container, but I can do that in the constructor as well.

Do you think that's coupling together the Django and MongoDB instances too much though? That was the main reason I didn't do that. I guess as an outsider to much of `stitch`'s development I kind of view all of the `django.js` and `node.js` etc. specs as like separate and distinct entities, like someone else trying to run a Django application could copypaste my `django.js` example and just pass in the relevant environment variables they needed for whatever database configuration and in a file like `example.js` glue all of the services together. But of course I don't really know the specifics behind `stitch` so I'm not sure what is more idiomatic.

kklin: We don't know what's idiomatic yet 😄 

I do have the general opinion that we should write our specs to be as simple as possible for our ecosystem right now. So if we introduce another django app and find that passing in the environment variable is a better approach, we could switch.

BTW, I think of the specific environment variable as an implementation detail of the `django` container, and is thus better abstracted in the constructor. Maybe we could pass in some sort of object that wraps the database and defines a `djangoConnector` method that the constructor calls to setup the environment. Don't know though, it's really hard to say without writing more specs.

Baisang: Okay...so weird things happened. If I pass in the environment variable `MONGO_URI` inside the django constructor, it doesn't get set on the containers. e.g. if i do:

```  js
var env = cfg.env || {};
env["MONGO_URI"] = mongo.uri("django-example")
```

in the constructor for `Django`, it's not getting set on the running containers. Also, if I try to connect the app with `mongo` inside the `Django` constructor like so:

``` js
mongo.connect(mongo.port(), this);
this.connect(mongo.port(), mongo);
```

I end up getting 504s when everything finishes loading, but it's fine when I connect the apps like I did before in `django-example.js`

kklin: Interesting, I'd try running the deployment that you know works, then running your modified one and see what the diff says.

Baisang: Here's the diff I'm getting:

https://gist.github.com/Baisang/3f0900c7bc43a73fdd0703c3ff85d407

So seems like connections need to be made outside of constructor? It's weird how adding env variables in constructor doesn't seem to work...

Baisang: https://gist.github.com/Baisang/99a42dc3c53694a5d07b2206d0660c55

specs/django/django.js
```diff
@@ -0,0 +1,38 @@
...
+
+Django.prototype.port = function() {
+  return this._port;
+};
+
+Django.prototype.connect = function(port, to) {
+  to.services().forEach(function(service) {
+    this._app.connect(port, service);
+  }.bind(this));
+};
```
tejasmanohar: bind is ugly, just use a loop or write `var self = this` above haha (i probably did this elsewhere, sorry if so)

tejasmanohar: this is what => fn / lambdas are for but we use otto so dont have that :\

ejj:
Assigning to @kklin as he already looked at it. 

@Baisang Could you please rebase?

specs/django/django-example.js
```diff
@@ -0,0 +1,34 @@
+var Django = require("github.com/NetSys/quilt/specs/django/django")
+var HaProxy = require("github.com/NetSys/quilt/specs/haproxy/haproxy").Haproxy;
+var Mongo = require("github.com/NetSys/quilt/specs/mongo/mongo");
+
+// Infrastructure
+var deployment = createDeployment({
+  adminACL: ["local"],
```
kklin: The `local` is implicit, I'd just delete this line entirely.

specs/django/django-example.js
```diff
@@ -0,0 +1,34 @@
...
+});
+
+var baseMachine = new Machine({
+  provider: "Amazon",
+  cpu: new Range(2),
```
kklin: I think this will result in big machines being used, I'd just omit the CPU and RAM constraints, and make it use the smallest instance.

specs/django/django-example.js
```diff
@@ -0,0 +1,34 @@
...
+  ram: new Range(2),
+  // sshKeys: githubKeys("CHANGE_ME"), // Replace with your GitHub username
+});
+
+// Applications
+var mongo = new Mongo(3);
+
+var django = new Django({
+  nWorker: 3,
+  image: "baisang/django-polls", // Replace with your own image, for example
```
kklin: I think we can delete the comment.

specs/django/django-example.js
```diff
@@ -0,0 +1,34 @@
...
+}, mongo);
+
+var haproxy = new HaProxy(3, django.services());
```
kklin: Let's just do 1 haproxy instance.. it makes things simpler. If there's only 1, we'd also get rid of the placement constraint below, which was put in place for the MEAN README guide.

specs/django/django-example.js
kklin: And we should make the image `quilt/django-polls`.

Baisang:
fixed, I don't have access to the quilt dockerhub yet so someone should just `docker build` the Dockerfile I have contained and tag it appropriately

specs/django/django-example.js
```diff
@@ -0,0 +1,30 @@
+var Django = require("github.com/NetSys/quilt/specs/django/django")
+var HaProxy = require("github.com/NetSys/quilt/specs/haproxy/haproxy").Haproxy;
+var Mongo = require("github.com/NetSys/quilt/specs/mongo/mongo");
+
+// Infrastructure
+var deployment = createDeployment({
```
kklin: Nit - we could just put this all on one line.

PR #713, opened 2017-01-20
Public internet test

kklin:


specs/nginx/Makefile
```diff
@@ -0,0 +1,9 @@
+REPO = quilt/nginx
```
kklin: This image needs to be pushed for this spec to work.

specs/nginx/default.tmpl
```diff
@@ -0,0 +1,16 @@
...
+
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+    }
+
+    # redirect server error pages to the static page /50x.html
+    #
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
```
LuiseV: Is it necessary to set this location explicitly? The root is the same as we set above, and the path prefix matches.

quilt-tester/tests/pub-facing/check_public_connections.go
```diff
@@ -0,0 +1,126 @@
...
+		log.Info("PASSED")
+	} else {
+		log.Info("FAILED")
+	}
+}
+
+func test(machines []db.Machine, containers []db.Container,
+	connections []db.Connection) bool {
+
+	// Map of label to publicly-connected ports to the label.
```
LuiseV: Maybe "Map of label to its publicly exposed ports" or something like that. I had to read this comment a few times.

specs/nginx/default.tmpl
kklin: The config is copied from the official nginx container, so I'm inclined to just keep it the same. Does that sound alright?

quilt-tester/tests/pub-facing/check_public_connections.go
kklin: Fixed, that sounds much better 😄

specs/nginx/default.tmpl
LuiseV: Sure, then just keep it. But I wonder why they wrote it that way. Exact location matches are apparently faster, but I would think that has practically no effect here.... But sounds good, I'm good with this then.

PR #716, opened 2017-01-22
Quilt ps

Baisang:


quiltctl/command/ps.go
```diff
@@ -103,8 +107,93 @@ func (pCmd *Ps) run() error {
 		return fmt.Errorf("unable to query containers: %s", err)
 	}
 
+	workerContainers, err := pCmd.queryWorkers(machines)
+	// no error means we can update our list of containers with information from
+	// the local worker, otherwise we just display what the leader sees
+	if err == nil && len(workerContainers) > 0 {
+		var updatedContainers []db.Container
+		pairs, onlyLeader, onlyWorkers :=
+			join.Join(containers, workerContainers, containerJoinScore)
```
ejj: Probably cleaner to just use a hash join and have the key be the stitchid

quiltctl/command/ps.go
```diff
@@ -103,8 +107,93 @@ func (pCmd *Ps) run() error {
...
+		for _, pair := range pairs {
+			lc := pair.L.(db.Container)
+			wc := pair.R.(db.Container)
+			if wc.Status != "" && lc.Status != wc.Status {
```
ejj: I don't see the point of the `lc.Status != wc.Status` check (or the similar check below

ejj: Yeah why not just always display the `wc` container?  Seems a bit simpler

quiltctl/command/ps.go
```diff
@@ -103,8 +107,93 @@ func (pCmd *Ps) run() error {
 		return fmt.Errorf("unable to query containers: %s", err)
 	}
 
+	workerContainers, err := pCmd.queryWorkers(machines)
+	// no error means we can update our list of containers with information from
+	// the local worker, otherwise we just display what the leader sees
+	if err == nil && len(workerContainers) > 0 {
```
ejj: This whole if block is a bit ugly here.  I may pull it into its own function.  Would make it a tad easier to read.

minion/etcd/network_test.go
```diff
@@ -62,6 +63,10 @@ func TestUpdateEtcdContainers(t *testing.T) {
 
 	resultSlice := []db.Container{}
 	json.Unmarshal([]byte(resultContainers), &resultSlice)
+	// hack: zero out time values so assert doesn't get mad
```
ejj: Comments should start with capitals and end with periods.

Also, I'd put a newline before the comment.

ejj:
Also there's still some low hanging fruit w.r.t. testing:

https://codecov.io/gh/NetSys/quilt/compare/8e392599513f98e23b1bbc13931744f3b8aba11c...5618cde896a8688338cb49ed8dee14c9bb8d9ea0#64622F636F6E7461696E65722E676F-97

https://codecov.io/gh/NetSys/quilt/compare/8e392599513f98e23b1bbc13931744f3b8aba11c...5618cde896a8688338cb49ed8dee14c9bb8d9ea0#7175696C7463746C2F636F6D6D616E642F70732E676F-187

https://codecov.io/gh/NetSys/quilt/compare/8e392599513f98e23b1bbc13931744f3b8aba11c...5618cde896a8688338cb49ed8dee14c9bb8d9ea0#7175696C7463746C2F636F6D6D616E642F636F6E7461696E65722E676F-148

Baisang:
ahh thanks for the pointer, I'll add/modify tests to cover those statements +1

minion/scheduler/worker.go
```diff
@@ -90,10 +90,13 @@ func syncWorker(dbcs []db.Container, dkcs []docker.Container, subnet net.IPNet)
 		dbc := pair.L.(db.Container)
 		dkc := pair.R.(docker.Container)
 
-		if dbc.DockerID != dkc.ID {
+		if dbc.DockerID != dkc.ID || dbc.Status != dkc.Status ||
```
ejj: Eh, this feels pretty fragile.  The if statement is somewhat of a premature optimization anyways.  I would just  ditch the if statement and put every pair in changed.

quiltctl/command/ps.go
```diff
@@ -62,8 +63,8 @@ func (pCmd *Ps) run() error {
 	var machines []db.Machine
 
 	connectionErr := make(chan error)
-	containerErr := make(chan error)
```
ejj: Useless change

quiltctl/command/ps.go
```diff
@@ -83,6 +84,7 @@ func (pCmd *Ps) run() error {
 			containers, err = leaderClient.QueryContainers()
 			containerErr <- err
 		}()
+
```
ejj: Useless change

quiltctl/command/ps.go
```diff
@@ -103,8 +105,89 @@ func (pCmd *Ps) run() error {
...
+		}
+	}
+
+	var workerContainers []db.Container
+	numMachines := len(workerMachines)
+	workerChannel := make(chan []db.Container, numMachines)
+	workerErr := make(chan error, numMachines)
+	for _, m := range workerMachines {
+		client, err := pCmd.clientGetter.Client(api.RemoteAddress(m.PublicIP))
+		if err == nil {
```
ejj: In general, I think it's better to do `if err != nil` and the return.  That allows you to have less indenting for the main code block, making the code a bit easier to read

quiltctl/command/ps.go
```diff
@@ -103,8 +105,89 @@ func (pCmd *Ps) run() error {
...
+				workerChannel <- qContainers
+				workerErr <- qErr
+			}()
+		} else {
+			return nil, err
+		}
+	}
+
+	for i := 0; i < numMachines; i++ {
+		if err := <-workerErr; err != nil {
```
ejj: I'm not convinced this is the right behavior.  If one worker has an error, that means we can't display containers for the rest of them?  I think the better thing to do would be log  in the go routine (not sure if Warn or Debug is better).  That way you could ditch `workerErr` and just return `nil` on `workerChannel` if there's an error.

kklin:
I forget, weren't we not going to commit `container.Created` into etcd? (@ejj)

BTW, I think `quilt containers` needs to be updated to use the same logic for querying container information. It'd be kind of odd for `quilt ps` to display different information than `quilt containers`, although I think that doesn't happen now because we're copying the `Created` information into the database on the leader via etcd.

PR #717, opened 2017-01-24
Add Elasticsearch and Kibana specs

kklin:


kklin:
I even wrote a readme for it! I'd appreciate feedback from you, @LuiseV 😄

specs/kibana/kibana.js
```diff
@@ -0,0 +1,24 @@
...
+                "--elasticsearch", es.uri(),
+            ]
+        )
+    ]);
+    es.addClient(this.service);
+}
+
+Kibana.prototype.public = function() {
+    publicInternet.connect(this.port, this.service);
+    return this;
```
Baisang: so is this style,
``` js
var kb = new Kibana(es).public();
```
preferable to
``` js
var kb = new Kibana(es);
kb.public()
```

specs/elasticsearch/elasticsearch.js
```diff
@@ -0,0 +1,44 @@
...
+                "--network.host", hosts[i]
+            );
+        });
+    }
+
+    this.service.connect(this._transportPorts, this.service);
+}
+
+Elasticsearch.prototype.uri = function() {
+    return "http://" + this.service.hostname() + ":" + this.port;
```
Baisang: I assume we're not going to worry about `https` for this example?

specs/kibana/main.js
```diff
@@ -0,0 +1,13 @@
+var Elasticsearch = require("github.com/NetSys/quilt/specs/elasticsearch").Elasticsearch;
+var Kibana = require("github.com/NetSys/quilt/specs/kibana").Kibana;
+
+var clusterSize = 2;
+
+var deployment = createDeployment({});
+var baseMachine = new Machine({provider: "Amazon"});
```
Baisang: What happens when we don't specify ram/cpu here? What's the default?

specs/kibana/kibana.js
kklin: I think you can argue either way. The first kind of implies that `public()` returns a new instance of `Kibana` when it really just mutates one instance, which could be confusing. But right now my main concern is making the top level specs as pretty-looking as possible.

specs/elasticsearch/elasticsearch.js
kklin: I don't think we need to figure that out in this first commit, although we definitely need to eventually.

specs/kibana/main.js
kklin: We end up picking the smallest size in https://github.com/NetSys/quilt/blob/master/cluster/machine/amazon.go because the `cpu` and `ram` constraints are zero.

kklin:
re `quilt get`, I could see how that'd be confusing, but I think that belongs in the getting started guide. I'll add a short sentence about the specific things you need configured (e.g. AWS credentials, QUILT_PATH) to this Readme though.

specs/kibana/README.md
```diff
@@ -0,0 +1,134 @@
+# Elasticsearch and Kibana
+
+[main.js](main.js) contains an example deployment that boots an Elasticsearch
+cluster connected to Kibana, and makes the Kibana web interface publically
+accessible.
+
+[Elasticsearch](https://www.elastic.co/products/elasticsearch) is a
+"distributed, RESTful search and analytics engine"
```
LuiseV: Missing period.

LuiseV:
@kklin The readme looks great! Just a single missing period.

specs/kibana/README.md
kklin: Good catch, fixed!

kklin:
@ejj Do you want to review this?

ejj:
Sure

Ethan (Phone)

specs/elasticsearch/elasticsearch.js
```diff
@@ -0,0 +1,44 @@
+function Elasticsearch(n) {
+    var ref = new Container("elasticsearch:2.4",
```
ejj: It's not the patch for it, but I do wish we could make `5` work ...  I think it's at least looking into making the sysctl change.  I don't know what the implications are though

specs/elasticsearch/elasticsearch.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    this.service.connect(this._transportPorts, this.service);
+}
+
+Elasticsearch.prototype.uri = function() {
+    return "http://" + this.service.hostname() + ":" + this.port;
+}
+
+Elasticsearch.prototype.public = function() {
```
ejj: I wonder if a lot of this stuff could be handled using inheritance from some default service object.  May help us cut down on the boilerplate, plus you could still override the default object.

Not for this patch of course, just a thought.

kklin: Yeah, I think that could work. It's a huge pain right now.

specs/elasticsearch/elasticsearch.js
kklin: I'll make a trello card.

PR #718, opened 2017-01-24
stitch: Make StitchID order-independent

kklin:


stitch/bindings.js
```diff
@@ -26,10 +23,48 @@ function Deployment(deploymentOpts) {
...
+    containers.forEach(function(c) {
+        var k = containerKey(c);
+        if (!containerCount[k]) {
+            containerCount[k] = [];
+        }
+        containerCount[k].push(c._refID);
+    });
+
+    Object.keys(containerCount).forEach(function(k) {
+        containerCount[k] = _.uniq(containerCount[k]).sort();
```
LuiseV: Why would the refIDs not be unique? If they aren't unique would it be a sign of something bad, so we should warn or error instead of ignore?

kklin: It could happen if the same container is in multiple services:
```
var foo = new Container("foo");
var bar = new Service("bar", [foo]);
var baz = new Service("baz", [foo]);
```

I'll add a comment.

LuiseV: Ah, I see! I misunderstood the comment about refID. Would be great with a comment here!

kklin:
I'll merge this after @ejj figures out the current problem with the network test

PR #722, opened 2017-01-24
gitignore: Ignore *.cov metadata

Baisang:


.gitignore
```diff
@@ -25,6 +25,7 @@ _testmain.go
 
 quilt
 
+*.cov
 *.cov.html
 *.cov.out
```
ejj: Do these still happen?

Baisang: Yeah, I just made a test directory with `abc.cov`, `abc.cov.html`, and `abc.cov.out`. Adding a `.gitignore` with just `*.cov` only gets rid of `abc.cov`

ejj: Oh what I'm saying is, do we still have .cov.out files when we run `make coverage`?

ejj: Also, what about *.cov*

Baisang: `*.cov*` would get rid of all of them. I'm not opposed to that.

I get `*.cov` and `*.cov.html` files, I'm not sure if others are still getting `*.cov.out`?

PR #724, opened 2017-01-25
minion: Reorganize IP management

ejj:
Before this patch, IP addresses were allocated by the workers, who
    then told the master about them so logical ports could be created.
    This round trip needlessly complicated the code, and increased the
    amount of time a container needed to wait for its network to be ready.

    This patch, instead, allocates container IP address in the master
    before communicating them to the minions. This has a number of
    significant benefits:

        * DNS entries of every container in the cluster can be populated
          even before containers boot.

        * The master can create logical ports in parallel as the workers
          boot containers.

        * Container IPs need not be written to Etcd by the workers.

        * Minions no longer need allocate subnets. Both simplifying the
          minion etcd synchronization code and ipdef.

    Also, in the process, this does a general refactoring of the code to
    make it easier to reason about.

Note that this series depends on @Baisang's quilt ps patch.


minion/docker/docker_test.go
```diff
@@ -222,6 +224,23 @@ func TestRunEnv(t *testing.T) {
 	assert.Equal(t, env, container.Env)
 }
 
+func TestConfigureNetwork(t *testing.T) {
+	md, dk := NewMock()
+
+	err := dk.ConfigureNetwork("quilt")
+	assert.NoError(t, err)
+	assert.Equal(t, &dkc.Network{
```
Baisang: Can you pull the `&dkc.Network{...` definition out in a `exp` variable? I think it would be more readable that way.

minion/etcd/util.go
```diff
@@ -0,0 +1,54 @@
+package etcd
+
+import (
+	"encoding/json"
+	"sort"
+
+	"github.com/coreos/etcd/client"
+)
+
+func joinNotifiers(a, b chan struct{}) chan struct{} {
```
Baisang: I'm not too sure what this does, is it just ensuring that whenever `c` has a "new" value that means that `a` and `b` have both been equally drained of 1 value?

It wasn't immediately clear to me (of course I am relatively new to `golang`) so I think a brief comment would help others reading it understand.

ejj:
Rebased on master

minion/docker/docker.go
```diff
@@ -117,7 +113,20 @@ func (dk Client) Run(opts RunOptions) (string, error) {
 		DNSSearch:   opts.DNSSearch,
 	}
 
-	id, err := dk.create(opts.Name, opts.Image, opts.Args, opts.Labels, env, hc, nil)
+	var nc *dkc.NetworkingConfig
+	if opts.IP != "" {
+		nc = &dkc.NetworkingConfig{
```
Baisang: I'm not sure if there is a cleaner way to define these structs and the map without it getting really nested. Defining separately might just make it more confusing. Just makes me feel funny with all the tabs.

ejj: Yeah, this is just kinda how Go is unfortunately.  I don't think there's really a good way to do it.

minion/etcd/connection.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+const connectionPath = "/connections"
+
+func runConnection(conn db.Conn, store Store) {
+	etcdWatch := store.Watch(connectionPath, 1*time.Second)
+	trigg := conn.TriggerTick(60, db.ConnectionTable)
+	for range joinNotifiers(trigg.C, etcdWatch) {
+		if err := runConnectionOnce(conn, store); err != nil {
+			log.WithError(err).Warn("Failed to sync connections with Etcd.")
```
kklin: No period in logs, right?

minion/etcd/container.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	conn.Txn(db.ContainerTable).Run(func(view db.Database) error {
+		joinContainers(view, etcdDBCs)
+		return nil
+	})
+}
+
+func joinContainers(view db.Database, etcdDBCs []db.Container) {
+	// The join contains only those fields that require restart of the container.
```
kklin: Why "only those fields that require restart of the container"? Could we just join on StitchID?

minion/etcd/util.go
ejj: So it's actually pretty simple, every time `a` or `b` has a notification, it notifies on `c`.  With on caveat.  If the receiver of `c` is too slow and has queued up notifications, then it will drop new notifications from `a` and `b` until it can catch up.

minion/etcd/container.go
ejj: So on master right now, there's no guarantee that changes to StitchID imply changes to things like the Image.  Of course, with your StitchID patch that will be the case, but I also like the idea of not relying on that in the QRI.

PR #726, opened 2017-01-26
Docker: Add network inactivity timeout to docker operations

stevenlin598:
Before, pulling docker images could stall the thread indefinitely if the
pull was somehow interrupted. Pull operations now error in the case of
extended network inactivity.

### DIFF
--- a/minion/docker/docker.go
+++ b/minion/docker/docker.go
@@ -18,6 +18,7 @@ import (
 )
 
 var pullCacheTimeout = time.Minute
+var networkTimeout = time.Minute
 
 // ErrNoSuchContainer is the error returned when an operation is requested on a
 // non-existent container.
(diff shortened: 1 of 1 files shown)


kklin:
@stevenlin598 I found this in the go-dockerclient API: https://godoc.org/github.com/fsouza/go-dockerclient#Client.SetTimeout

Would setting this be equivalent? If so, I'd prefer setting the timeout once and having it apply to all Docker operations.

stevenlin598:
@kklin Tried Client.SetTimeout and seems to be for something different; the docker pull didn't return with a fail after the timeout duration passed.

kklin:
Thanks @stevenlin598. Looking through the source it looks liek `Client.SetTimeout` sets the initial timeout, not the inactivity timeout.

kklin:
@LuiseV Yeah, this PR was already merged.

PR #727, opened 2017-01-26
Logs test improvements

kklin:


quilt-tester/tests/100-logs/check_logs.go
```diff
@@ -42,7 +42,8 @@ func main() {
 		outputStr := string(logsOutput)
 		fmt.Println(outputStr)
 
-		if strings.Contains(outputStr, "ERROR") ||
+		if strings.Contains(outputStr, "goroutine 0") ||
```
LuiseV: We should have a comment explaining why we look for `goroutine 0`.

kklin: Done.

PR #728, opened 2017-01-28
provider: Support for DigitalOcean

vivi:
~Still need to add tests that cause the DigitalOcean API to throw an error.~


cluster/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,293 @@
...
+		{
+			ID:        123,
+			Name:      testNamespace,
+			Networks:  network,
+			SizeSlug:  "size",
+			VolumeIDs: []string{"foo"},
+			Region:    nyc,
+		},
+
+		// This droplet should not be listed.
```
sirspinach: "should not be listed because it has the wrong namespace"

cluster/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,365 @@
...
+	opt := &godo.ListOptions{}
+	_, _, err = clst.client.ListDroplets(opt)
+
+	return clst, err
+}
+
+// Creation is broken out for unit testing.
+var newDigitalOcean = func(namespace string) (*DoCluster, error) {
+	// Underscores are illegal. Uppercase is illegal in volume names.
+	namespace = strings.ToLower(strings.Replace(namespace, "_", "-", -1))
```
secant: ah, good idea

secant:
ready for more review (all tests done)

cluster/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,386 @@
...
+	_, _, err = clst.client.ListDroplets(opt)
+
+	return clst, err
+}
+
+// Creation is broken out for unit testing.
+var newDigitalOcean = func(namespace string) (*DoCluster, error) {
+	// Underscores are illegal. Uppercase is illegal in volume names.
+	if strings.Contains(namespace, "-") || strings.Contains(namespace, "_") {
+		log.Warning("DigitalOcean namespaces cannot have underscores, hypens, " +
```
sirspinach: "hyphens"

cluster/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,386 @@
...
+	opt := &godo.ListOptions{}
+	_, _, err = clst.client.ListDroplets(opt)
+
+	return clst, err
+}
+
+// Creation is broken out for unit testing.
+var newDigitalOcean = func(namespace string) (*DoCluster, error) {
+	// Underscores are illegal. Uppercase is illegal in volume names.
+	if strings.Contains(namespace, "-") || strings.Contains(namespace, "_") {
```
sirspinach: "-" is allowed

sirspinach: did you mean to check uppercase here too?

cluster/digitalocean/digitalocean.go
sirspinach: Might want to note that namespace is automatically being changed to X.

cluster/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+}
+
+// Creates a new machine, and waits for the machine to become active.
+func (clst DoCluster) createAndAttach(m machine.Machine) error {
+	/*
+		if len(machine.FloatingIP) > 0 {
+			panic("DigitalOcean floating IPs are unimplemented")
+		}
+	*/
```
sirspinach: Do you mean to leave this comment in here?

cluster/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,386 @@
...
+func (clst DoCluster) Stop(machines []machine.Machine) error {
+	if len(machines) <= 0 {
+		return nil
+	}
+
+	var wg sync.WaitGroup
+	errChan := make(chan error, 1)
+	for _, m := range machines {
+		wg.Add(1)
+		go func(m machine.Machine) {
```
sirspinach: I think it would be good to add some concurrency limit here.

cluster/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,386 @@
...
+	if len(bootSet) <= 0 {
+		return nil
+	}
+
+	var wg sync.WaitGroup
+	// If any calls to createAndAttach fail, one of the errors will be fed to this
+	// channel.
+	errChan := make(chan error, 1)
+	for _, m := range bootSet {
+		wg.Add(1)
```
sirspinach: same re concurrency limit

cluster/digitalocean/digitalocean.go
sirspinach: Hmmm... I wonder how automatically changing the namespace will fit in with quilt stop. If I start a DO cluster with namespace "UPPERCASE", will running `quilt stop UPPERCASE` stop that cluster later or leave it running?

cluster/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+func TestNew(t *testing.T) {
+	mc := new(mockClient)
+	clust := &DoCluster{
+		namespace: testNamespace,
+		client:    mc,
+	}
+
+	// Log a bad namespace.
+	newDigitalOcean("___ILLEGAL---")
```
sirspinach: might want to check check the expected namespace here.

cluster/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+	// ListDroplets throws an error.
+	mc.On("ListDroplets", mock.Anything).Return(nil, nil, errMock)
+	outClust, err = New(testNamespace)
+	assert.Equal(t, clust, outClust)
+	assert.EqualError(t, err, errMsg)
+}
+
+func TestToken(t *testing.T) {
+	tokenSrc := &tokenSource{
```
sirspinach: Can this test (and other tests) be parallelized with `t.Parallel()`?

cluster/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+func TestStop(t *testing.T) {
+	mc := new(mockClient)
+	doClust, err := newDigitalOcean(testNamespace)
+	assert.Nil(t, err)
+	doClust.client = mc
+
+	util.Sleep = func(t time.Duration) {}
+
+	// Test empty stop set
```
sirspinach: end comment with period

cluster/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+	mc.On("DeleteDroplet", 123).Return(nil, nil).Once()
+
+	mc.On("DeleteVolume", "abc").Return(nil, nil).Once()
+
+	err = doClust.Stop(stopSet)
+
+	// Make sure machines are stopped.
+	mc.AssertNumberOfCalls(t, "GetDroplet", 2)
+	mc.AssertNumberOfCalls(t, "DeleteVolume", 1)
```
sirspinach: These assertions are unnecessary if `mc.AssertExpectations()` is called.

cluster/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	mc.On("GetDroplet", 123).Return(&godo.Droplet{
+		Status:    "active",
+		VolumeIDs: []string{"abc"},
+	}, nil, nil).Once()
+
+	mc.On("GetDroplet", 123).Return(nil, nil, nil).Once()
+
+	mc.On("DeleteDroplet", 123).Return(nil, nil).Once()
```
sirspinach: Can't guarantee that DeleteDroplet is called exactly one time (rather than 0 or 1 times) without calling `mc.AssertExpectatations()` or `mc.AssertNumberOfCalls`.

cluster/digitalocean/digitalocean_test.go
sirspinach: However, using `mc.AssertExpectations()` might (i'm not sure) require you to reinitialized `mc` when it's used later in the same test.

cluster/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+	mc.On("DeleteVolume", "abc").Return(nil, errMock).Once()
```
sirspinach: Same here, re `mc.AssertExpectations()`. The `Once()` only guarantees that this method is called 0 or 1 times.

cluster/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,505 @@
...
+	mc.On("CreateVolume", mock.Anything).Return(&godo.Volume{
+		ID: "abc",
+	}, nil, nil).Once()
+
+	mc.On("AttachVolume", mock.Anything, mock.Anything).Return(nil, nil, nil).Once()
+
+	err = doClust.Boot(bootSet)
+	// Make sure machines are booted.
+	mc.AssertNumberOfCalls(t, "CreateDroplet", 1)
+	mc.AssertNumberOfCalls(t, "CreateVolume", 1)
```
sirspinach: better to use `t.AssertExpectations()` here too.

cluster/digitalocean/digitalocean.go
ejj: Meh, I'm not sure it matters in this case.  It's not like containers where you've got thousands.

If you're going to do a concurrency limit, you can use a channel as a semaphore.  Better than the way we've been doing it elsewhere in the code base.

cluster/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,386 @@
...
+	var wg sync.WaitGroup
+	errChan := make(chan error, 1)
+	for _, m := range machines {
+		wg.Add(1)
+		go func(m machine.Machine) {
+			defer wg.Done()
+			if err := clst.deleteAndWait(m.ID); err != nil {
+				log.WithError(err).Warning("Failed to stop droplet")
+				select {
+				case errChan <- err:
```
ejj: It would be much cleaner if you wrote the err to the channel whether or not it it was nil.  That way you could loop over the errors outside of the go routine and log, while ditching the wait group.

PR #730, opened 2017-01-28
google: Fix error caused by subsequent inits.

tejasmanohar:
If a user who doesn't have GCE setup locally init'd the cluster
twice, the error wouldn't be returned by the client init function
the second time since we run the block that causes an error in
once.Do. This removes the clever code in favor of the simpler,
correct solution.


cluster/google/client.go
```diff
@@ -4,12 +4,10 @@ import (
 	"context"
```
kklin: Commit message shouldn't have trailing period

cluster/google/client.go
```diff
@@ -37,16 +35,13 @@ type clientImpl struct {
...
-
-		service, err = newComputeService(context.Background())
-	})
+	keyfile := filepath.Join(os.Getenv("HOME"), ".gce", "quilt.json")
+	err := os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", keyfile)
+	if err != nil {
+		panic(err)
+	}
+
+	service, err = newComputeService(context.Background())
```
kklin: Doesn't have to be this PR, but does `service` have to be a global variable?

tejasmanohar: ya fixed

cluster/google/client.go
tejasmanohar: fixed

tejasmanohar:
ready :)

kklin:
Thanks @tejasmanohar, will merge after travis is happy

PR #732, opened 2017-01-28
network: Don't try to recreate Quilt network

kklin:
Before, we tried to create the Quilt Docker network whenever the minion
started. However, if it already existed (e.g. if the minion were
restarted and the previous instance already created the network), the
call to `CreateNetwork` failed, and the minion crashed.

Note that restarting the minion still doesn't always work. There's some race conditions in the supervisor code that @Baisang will probably improve with his current work.


ejj:
Why were the minions restarting?

kklin:
I was manually restarting the container. I used to be able to push a new version of my dev build, and use it just by restarting the minion containers. You could imagine this happening if the minion crashed, though, and systemd restarted it, like in the integration test failure a couple days ago.

ejj:
My other question is ... could we just check the error that CreateNetwork returned, and swallow it silently if it was "already exists" or some equivalent?  Would be a lot easier than listing the networks if possible.

ejj:
Ping, what do we need to do to get this merged?

kklin:
The exact error is `failed to allocate gateway (10.0.0.1): Address already in use`. I feel like checking for that error message is less direct than checking that the Quilt network hasn't already been created.

I don't feel that strongly about it, though, so if you'd prefer to just check the error I'm fine with that.

ejj:
Ah no that error isn't as clear as I'd like. I'd say merge it as is.

PR #734, opened 2017-01-28
assorted cluster fixes

tejasmanohar:
@kklin this is in direct response to https://github.com/NetSys/quilt/pull/690#pullrequestreview-18982827


cluster/cluster.go
```diff
@@ -140,8 +140,18 @@ func (clst cluster) updateCloud(machines []machine.Machine, act action) {
 		return
 	}
 
+	actionString := ""
```
mmussomele: 'fix' in the commit message should be capitalized

cluster/cluster.go
```diff
@@ -179,9 +179,9 @@ func (clst cluster) updateCloud(machines []machine.Machine, act action) {
 			case boot:
 				log.Infof("Unable to boot machines on %s.", p)
 			case stop:
-				log.Infof("Successfully stopped machines on %s", p)
```
mmussomele: Same with the commit message

cluster/cluster.go
mmussomele: Also unrelated, the build seems to be failing due to some kind of non-deterministic bug. The TestSync in cluster_test.go fails every once in a while on the master branch, perhaps also related to #690?

kklin: We have an ugly amount code of in this function related to picking the appropriate `actionString`. Could we pull it out into a method of `action` perhaps?

kklin: Ethan and I discussed this a bit in slack. It's more like a deficiency in the join library was revealed by the floating IP tests.

mmussomele: oh gotcha, I didn't realize they were related 👍

tejasmanohar: i could make some past/present tense functions though to me, that feels overengineered (but i don't mind much)... is that what you were thinking? @kklin

tejasmanohar: interesting... i'll take a look

kklin: What if we did something like `log.WithField("action", act).Info("Updating cloud")`?
BTW, what if we made `action` type `string`?

tejasmanohar: I changed action from type string to just an enum (via int) because we still had to switch on the action type to change tense of string or make it fit right in the sentence anyways

mmussomele: Why does switching necessitate it be an enum? It seems like there's less work if it's a string.

tejasmanohar: it doesn't, but it's only slightly less work-- you'll save
```
actionString := ""
switch act {
case boot:
  actionString = "boot"
case stop:
  actionString = "stop"
case updateIPs:
  actionString = "update floating IPs of"
}
```
but it's also awkward IMO since the updateIPs action would be == "update floating IPs of"

i think it's better to just switch since the string is entirely dependent on usage

tejasmanohar: but if you both think the other method is better, i can do that- it's minor

cluster/cluster.go
tejasmanohar: changed

cluster/cluster.go
kklin: If we did `log.WithField("action", act).Info("Updating cloud")` we wouldn't have to worry about tense at all.

PR #739, opened 2017-01-30
make: Fix misspell invocation

kklin:
Some versions of `find` require the `-type` argument be only one letter.


kklin:
@ejj can you try it?

kayousterhout:
LGTM

ejj:
@kklin, I actually noticed the error in the travis logs -- doesn't happen for me locally.  Looking through the logs this times, it appears to be fixed.

PR #740, opened 2017-01-30
Container ID UX Improvements

ejj:
The new longer stitch IDs required me to copy and paste them from my terminal which I found intolerable.  This series change is it so you need only type the prefix of an ID (much like is possible with git and docker).  It also prints shortened IDs instead of the full length version to make `quilt ps` more readable.

### DIFF
--- a/api/util/util.go
+++ b/api/util/util.go
@@ -15,10 +15,25 @@ func GetContainer(c client.Client, stitchID string) (db.Container, error) {
 		return db.Container{}, err
 	}
 
+	var choice *db.Container
 	for _, c := range containers {
-		if c.StitchID == stitchID {
-			return c, nil
+		if len(stitchID) > len(c.StitchID) ||
+			c.StitchID[0:len(stitchID)] != stitchID {
+			continue
 		}
+
+		if choice != nil {
+			err := fmt.Errorf("ambiguous stitchIDs %s and %s",
+				choice.StitchID, c.StitchID)
+			return db.Container{}, err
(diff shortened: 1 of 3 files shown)


api/util/util.go
```diff
@@ -15,10 +15,25 @@ func GetContainer(c client.Client, stitchID string) (db.Container, error) {
 		return db.Container{}, err
 	}
 
+	var choice *db.Container
```
kklin: Why all this pointer stuff?

ejj: I'm using `nil` as an easy way to tell that no choice was made. Not sure of a cleaner way to do it other than having a bool (which seems just as ugly)

kklin: Do we have to do the copy thing? Could we just use `&c`?

ejj:
Nope doesn't work.  This is something rather ugly about go, but c is actually scoped outside of the for loop, not inside.  So as a result, on each iteration of the loop, c is updated to point at the next element.  If you do &c you're going to end up with the final element of the list every time.  (I found this in the unit tests, it was quite surprising).

ejj:
Nope doesn't work.  This is something rather ugly about go, but c is actually scoped outside of the for loop, not inside.  So as a result, on each iteration of the loop, c is updated to point at the next element.  If you do &c you're going to end up with the final element of the list every time.  (I found this in the unit tests, it was quite surprising).

kklin:
Ahh I see

PR #744, opened 2017-01-31
minion: Join containers based exclusively on the StitchID

ejj:
Now that the StitchID is more carefully crafted by Quilt.js, it seems
logical (and makes the system easier to reason about) if a given
container always maintains the same StitchID even as the spec changes.

This fixes a bug where booting a second copy of a particular system
on the same infrastructure could cause new nodes to connect to the
old system.

### DIFF
--- a/minion/engine.go
+++ b/minion/engine.go
@@ -6,7 +6,6 @@ import (
 	"github.com/NetSys/quilt/db"
 	"github.com/NetSys/quilt/join"
 	"github.com/NetSys/quilt/stitch"
-	"github.com/NetSys/quilt/util"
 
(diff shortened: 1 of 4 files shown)


kayousterhout:
LGTM. I fixed a small typo in your PR description (not sure if that gets used as the commit message?)

ejj:
What was the typo?  I think I'll have to edit it in the commit message directly, since we're using rebase and merge, it doesn't automatically move over.

Btw, could you formally approve the PR by making a review?  Doesn't matter in this case, but that's how quilt-bot knows to send it to the next person.

PR #745, opened 2017-01-31
ps: Log "no leader found" warning at debug level

ejj:
This warning message is needlessly alarmist as its normal not to find
the leader when the cluster is in the process of booting.


quiltctl/command/ps.go
```diff
@@ -97,7 +97,8 @@ func (pCmd *Ps) run() error {
 	fmt.Println()
```
kklin: In the commit message, "as _it's_ normal".

PR #746, opened 2017-01-31
scheduler: Try to spread out similar containers

ejj:
Even when a placement constraint isn't explicitly specified, it would
still be nice to spread out containers with the same image.  While
solving this problem fully doesn't make sense until we've implemented
a more robust scheduler, this patch implements a simple hack that in
practices spreads out similar containers in most cases.

### DIFF
--- a/minion/scheduler/master.go
+++ b/minion/scheduler/master.go
@@ -2,8 +2,11 @@ package scheduler
 
 import (
 	"container/heap"
+	"fmt"
+	"sort"
 
 	"github.com/NetSys/quilt/db"
+	"github.com/NetSys/quilt/util"
 	log "github.com/Sirupsen/logrus"
 )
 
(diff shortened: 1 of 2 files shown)


minion/scheduler/master.go
```diff
@@ -220,6 +223,8 @@ func makeContext(minions []db.Minion, constraints []db.Placement,
 		minion.containers = append(minion.containers, dbc)
 	}
 
+	sort.Sort(dbcSlice(ctx.unassigned))
```
kklin: I think the comment below is more appropriate here.

minion/scheduler/master.go
```diff
@@ -237,3 +242,27 @@ func (mh *minionHeap) Pop() interface{}   { panic("Not Reached") }
 func (mh minionHeap) Less(i, j int) bool {
 	return len(mh[i].containers) < len(mh[j].containers)
 }
+
+// We sort containers based on their image and command in an effort to encourage the
+// scheduler to spread them out. XXX: This is somewhat of a hack -- we need a more clever
```
kklin: New line before `XXX`?

kklin:
Oh, a unit test for the sorting would be nice too.

PR #749, opened 2017-02-01
specs: Remove PingWait()

ejj:
PingWait() was a holdover from when containers could come up before
their network was ready.  Now that the master allocates IP addresses,
the first time a container attempts to resolve one of its
dependencies, it's guaranteed to work.  Thus PingWait() is no longer
necessary.


specs/mongo/scripts/run.go
```diff
@@ -42,14 +40,10 @@ func main() {
...
 
-	if err := specs.PingWait(peers); err != nil {
-		log.Fatalf("Error ping wait: %s", err)
-	}
-
 	// At this point, mongod may or may not have gotten to the point where it can
 	// receive instructions to configure the replica set.  Thus, we have to wait
 	// until its ready by sleeping.
-	time.Sleep(2 * time.Second)
+	time.Sleep(5 * time.Second)
```
tejasmanohar: though this is still yucky :\

tejasmanohar: maybe events would help here

ejj: Hopefully

PR #753, opened 2017-02-02
Machine ID UX improvements

kklin:


quiltctl/command/ssh.go
```diff
@@ -35,6 +38,19 @@ func NewSSHCommand() *SSH {
 	}
 }
 
+var sshUsage = `usage: quilt ssh <id> [command]
+
+Create a SSH session with the specified id.
+Either a container or machine ID can be supplied.
+If no command is supplied, a login shell is created.
+
+To login to machine 5 with a specific private key:
```
ejj: I think this should be a ShortUUID to make it more consistent with what the user will actually be doing

PR #755, opened 2017-02-02
aws: Decrease boot and stop timeout

kklin:
Sometimes Amazon fails to create machines due to an "internal error on
launch". When this happens, 15 minutes is much too long to wait before
retrying.

The new timeout was somewhat arbitrarily picked based on the fact that
in the past 4 months on quilt-tester, the average boot time was 33
seconds, and the max was 90 seconds. The average and max times to stop
machines were 9 seconds and 40 seconds.


kklin:
@ejj What do you think about the new timeout? I wasn't sure how low to make it. There's some stats in the commit message.

PR #756, opened 2017-02-03
haproxy: Allow DNS changes

kklin:
Before, HAproxy resolved hostnames once at startup, and kept the same
association for the lifetime of the service, even if the IP changed in
DNS. This is obviously incorrect when load balanced containers get
restarted with new IPs. HAproxy now tries to updates IPs when health
checks to the old IP fail.


specs/haproxy/config/haproxy.cfg
```diff
@@ -7,6 +7,9 @@ defaults
     timeout client 5000
     timeout server 5000
 
+resolvers dns
+    nameserver gateway 10.0.0.1:53
```
ejj: I was under the impression that this wasn't necessary and it could use /etc/resolv.conf instead?

kklin: Oh, that'd be nice. I can't find any documentation on that online though, could you point me to it?

kklin: ping @ejj

ejj: Yeah I think I just misinterpreted the documentation.  I think I'd give it a go once without that line, but if it doesn't work we'll just hardcode it for now.

ejj: Yeah that was my thinking.

kklin:
Going to submit this PR on the new quilt/haproxy repo.

PR #759, opened 2017-02-03
cluster: Remove multi-region logic from providers

mmussomele:
Previously, each provider had to implement lots of boilerplate
logic to handle multi-region deployments. The desired behavior
could be implemented more generally by the cluster, allowing each
provider to assume that a given instance only deals with a single
region/cluster. This patch moves this code to the cluster.


cluster/amazon/amazon.go
```diff
@@ -133,277 +140,271 @@ func (clst Cluster) Boot(bootSet []machine.Machine) error {
...
 		}
 	}
 
 	return nil
 }
 
-func (clst Cluster) getClient(region string) client {
-	if _, ok := clst.clients[region]; !ok {
-		clst.clients[region] = clst.newClient(region)
+func (clst *Cluster) connectClient() {
```
ejj: I don't quite see the point of this function anymore.  Why not just connect to the client when the Cluster is created?  Is the issue that we're creating a cluster for each region even if we don't use it?  In that case, a future patch should simplify this.

mmussomele: Don't remember the exact reason, but it makes the unit tests annoying if we do it that way.

PR #761, opened 2017-02-05
Fix minion role at boot time through cloud cfg

Baisang:
This patch aims to fix minion role at boot time through the cloud cfg, in order to more easily simplify the minion modules (e.g. supervisor, network, etc.)

Some high level highlights:

- The `foreman` is no longer able to change a minion's role
- role has been added to the join in `cluster`. `cluster` gets the role for the `machine.Machine`s it gets from the cloud providers by asking the `foreman`
- role is passed down through cloud cfg
- minions accept role from cloud cfg (via modifying the `quilt minion` command) and use it instead of getting role from protobuf

### DIFF
--- a/cluster/cloudcfg/cloudcfg.go
+++ b/cluster/cloudcfg/cloudcfg.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"strings"
 	"text/template"
+
+	"github.com/quilt/quilt/db"
 )
 
 const (
@@ -12,7 +14,7 @@ const (
 
 // Ubuntu generates a cloud config file for the Ubuntu operating system with the
 // corresponding `version`.
-func Ubuntu(keys []string, version string) string {
+func Ubuntu(keys []string, version string, role db.Role) string {
 	t := template.Must(template.New("cloudConfig").Parse(cfgTemplate))
 
 	var cloudConfigBytes bytes.Buffer
@@ -20,10 +22,12 @@ func Ubuntu(keys []string, version string) string {
 		QuiltImage    string
 		UbuntuVersion string
 		SSHKeys       string
+		Role          string
 	}{
 		QuiltImage:    quiltImage,
 		UbuntuVersion: version,
 		SSHKeys:       strings.Join(keys, "\n"),
+		Role:          string(role),
 	})
 	if err != nil {
 		panic(err)
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -262,7 +262,10 @@ func (clst cluster) join() (joinResult, error) {
 			dbm := pair.L.(db.Machine)
 			m := pair.R.(machine.Machine)
 
-			dbm.CloudID = m.ID
+			if m.Role != db.None && m.Role == dbm.Role {
+				dbm.CloudID = m.ID
+			}
+
 			dbm.PublicIP = m.PublicIP
 			dbm.PrivateIP = m.PrivateIP
 
@@ -352,14 +355,18 @@ type syncDBResult struct {
 
 func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
 	ret := syncDBResult{}
+	cms = foreman.GetMachineRoles(cms)
 
 	pair1, dbmis, cmis := join.Join(dbms, cms, func(l, r interface{}) int {
 		dbm := l.(db.Machine)
 		m := r.(machine.Machine)
 
-		if dbm.CloudID == m.ID && dbm.Provider == m.Provider &&
+		// Don't join if m.ID and dbm.CloudID are empty.
+		if m.ID != "" && dbm.CloudID == m.ID &&
+			dbm.Provider == m.Provider &&
 			dbm.Region == m.Region && dbm.Size == m.Size &&
-			(m.DiskSize == 0 || dbm.DiskSize == m.DiskSize) {
+			(m.DiskSize == 0 || dbm.DiskSize == m.DiskSize) &&
+			(m.Role == db.None || dbm.Role == m.Role) {
 			return 0
 		}
 
@@ -374,16 +381,19 @@ func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
 		case dbm.Provider != m.Provider ||
 			dbm.Region != m.Region ||
 			dbm.Size != m.Size ||
-			(m.DiskSize != 0 && dbm.DiskSize != m.DiskSize):
+			(m.DiskSize != 0 && dbm.DiskSize != m.DiskSize) ||
+			(m.Role != db.None && dbm.Role != m.Role):
 			return -1
-		case dbm.CloudID == m.ID:
+		case m.ID != "" && dbm.CloudID == m.ID:
 			panic("Not Reached") // Should have been hit by the first join.
-		case dbm.FloatingIP == m.FloatingIP:
+		case m.Role != db.None && dbm.Role == m.Role:
 			return 1
-		case dbm.PublicIP == m.PublicIP || dbm.PrivateIP == m.PrivateIP:
+		case dbm.FloatingIP == m.FloatingIP:
 			return 2
-		default:
+		case dbm.PublicIP == m.PublicIP || dbm.PrivateIP == m.PrivateIP:
 			return 3
+		default:
+			return 4
 		}
 	})
 
@@ -398,7 +408,9 @@ func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
 			Provider: m.Provider,
 			Region:   m.Region,
 			DiskSize: m.DiskSize,
-			SSHKeys:  m.SSHKeys})
+			SSHKeys:  m.SSHKeys,
+			Role:     m.Role,
+		})
 	}
 
 	for _, pair := range append(pair1, pair2...) {
--- a/db/minion.go
+++ b/db/minion.go
@@ -1,7 +1,5 @@
 package db
 
-import "errors"
-
 // The Minion table is instantiated on the minions with one row.  That row contains the
 // configuration that minion needs to operate, including its ID, Role, and IP address
 type Minion struct {
@@ -39,9 +37,9 @@ func (db Database) SelectFromMinion(check func(Minion) bool) []Minion {
 	return result
 }
 
-// MinionSelf returns the Minion Row corresponding to the currently running minion, or an
-// error if no such row exists.
-func (db Database) MinionSelf() (Minion, error) {
+// MinionSelf returns the Minion Row corresponding to the currently running minion.
+// If there is no Minion Row, it panics
+func (db Database) MinionSelf() Minion {
 	minions := db.SelectFromMinion(func(m Minion) bool {
 		return m.Self
 	})
@@ -51,24 +49,23 @@ func (db Database) MinionSelf() (Minion, error) {
 	}
 
 	if len(minions) == 0 {
-		return Minion{}, errors.New("no self minion")
+		panic("no minion labeled Self")
 	}
 
-	return minions[0], nil
+	return minions[0]
 }
 
-// MinionSelf returns the Minion Row corresponding to the currently running minion, or an
-// error if no such row exists.
-func (conn Conn) MinionSelf() (Minion, error) {
+// MinionSelf returns the Minion Row corresponding to the currently running minion.
+// If there is no Minion Row, it panics
+func (conn Conn) MinionSelf() Minion {
 	var m Minion
-	var err error
 
 	conn.Txn(MinionTable).Run(func(view Database) error {
-		m, err = view.MinionSelf()
+		m = view.MinionSelf()
 		return nil
 	})
 
-	return m, err
+	return m
 }
 
 // SelectFromMinion gets all minions in the database that satisfy the 'check'.
--- a/db/minion.go
+++ b/db/minion.go
@@ -10,7 +10,6 @@ type Minion struct {
 	Self           bool   `json:"-"`
 	Spec           string `json:"-" rowStringer:"omit"`
 	AuthorizedKeys string `json:"-" rowStringer:"omit"`
-	SupervisorInit bool   `json:"-"`
 
 	// Below fields are included in the JSON encoding.
 	Role       Role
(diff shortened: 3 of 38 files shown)


ejj:
Does the previous patch work without `cluster: Add role to machine join`?  If not they should be squashed.

Baisang:
No, I'll squash them later. Just had them separate to make it easier to review.

The quiltctl and minion commits would also be squashed. I guess should squash the cluster: cloudcfg commit with those two as well.

ejj:
Ah just go ahead and squash them,  I like to see what the final commits are going to be during the review process

ejj:
Ping @tejasmanohar

cluster/cluster.go
```diff
@@ -339,14 +339,16 @@ type syncDBResult struct {
 
 func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
 	ret := syncDBResult{}
+	cms = foreman.GetMachineRoles(cms)
 
 	pair1, dbmis, cmis := join.Join(dbms, cms, func(l, r interface{}) int {
 		dbm := l.(db.Machine)
 		m := r.(machine.Machine)
 
 		if dbm.CloudID == m.ID && dbm.Provider == m.Provider &&
```
tejasmanohar: this code looks overwhelming, maybe it can be broken up into functions or similar? what do you think @ejj

Baisang: we could move the score function to be defined outside `syncDB`, that might help

minion/run.go
```diff
@@ -32,8 +32,24 @@ func Run() {
 	// Not in a goroutine, want the plugin to start before the scheduler
 	plugin.Run()
 
+	// XXX: As we are developing minion modules to use this passed down role
+	// instead of querying their db independently, we need to do this.
+	// Possibly in the future just pass down role into all of the modules,
+	// but may be simpler to just have it use this entry.
+	conn.Txn(db.MinionTable).Run(func(view db.Database) error {
```
ejj: Couple of things occur to me here.

1) This should happen even before we start the plugin

2) We should go ahead and change the plugin not to create the minion self in this patch 

3) I really want to see a patch in this series (after this one) that goes and and makes it so MinionSelf() can't return an error.

4) A minion writes itself to etcd -- what happens with this code if we don't know its region/provider/etc yet?  Does the code handle that properly?

minion/supervisor/supervisor.go
```diff
@@ -100,8 +100,7 @@ func (sv *supervisor) runSystemOnce() {
 		etcdRow = etcdRows[0]
 	}
 
-	if sv.role == minion.Role &&
-		reflect.DeepEqual(sv.etcdIPs, etcdRow.EtcdIPs) &&
+	if reflect.DeepEqual(sv.etcdIPs, etcdRow.EtcdIPs) &&
```
ejj: While you're in here, go ahead and change this to `util.StrSliceEqual`

minion/supervisor/supervisor.go
```diff
@@ -249,23 +241,17 @@ func (sv *supervisor) Remove(name string) {
 	}
 }
 
-func (sv *supervisor) SetInit(init bool) {
+func (sv *supervisor) SetInit() {
```
ejj: This change seems wrong ... why?

Baisang: There didn't seem to be a case where `SetInit(false)` would be called once minion role is fixed, so that's why I changed it.

ejj: Ah good point.

I think it would be much better if we just refactor the supervisor so we no longer have the init flag (it's one of the ugliest warts in the minion right now).  You could imagine that the minion could block starting up until the relevant containers have booted for example.  Could be a future patch (but I want that soon).

minion/run.go
Baisang: For 2, I'm not quite sure what you mean by plugin creating minion self? I don't see where that occurs.

About 4, for `supervisor.Run`, inside the [`updateWorker`](https://github.com/Baisang/quilt/blob/minion_role/minion/supervisor/supervisor.go#L146) and [`updateMaster`](https://github.com/Baisang/quilt/blob/minion_role/minion/supervisor/supervisor.go#L182) functions, the function returns early before the passed in minion private IP is passed in, so I believe `supervisor` will handle this properly.

[`scheduler:runMaster`](https://github.com/Baisang/quilt/blob/minion_role/minion/scheduler/master.go#L161) also seems to have checks for each of the minion config fields if they are null, as well as [`scheduler:runWorker`](https://github.com/Baisang/quilt/blob/minion_role/minion/scheduler/worker.go#L25).

I think `network.Run` checks if its `EtcdLeader()`? So I think that's separate from the `role`. And `Etcd.Run` would be also separate?

The only one I'm not sure of is `syncAuthorizedKeys`.... I think that can be cleaned up with this patch (removing `waitForMinion` and cleaning up `runOnce` to return if `m.AuthorizedKeys == nil` (although I believe it will do that now as it tries to read an empty authorized key file)

ejj: Ah sorry, by the plugin I meant SetMinionConfig in server.go

ejj:
also, needs to be rebased

Baisang:
was decided that `supervisor` should just be refactored with this patch change as well

Baisang:
@ejj ready for review again.

- `cluster` join changed to join on Role before Floating IP, PublicIP/PrivateIP. Additionally, `CloudID` is only assigned to a `db.Machine` when the foreman was able to determine the Role of the machine.Machine. This ensures that assigned `CloudID` joins always match (same behavior as before), but keeps Role join priority over other aspects (PublicIP, etc.).
- supervisor refactored into separate worker and master routines
- init flag removed
- minion self no longer returns error

ejj:
@Baisang please rebase when you have a chance.

Baisang:
@ejj rebased

minion/supervisor/supervisor.go
ejj: Ping

minion/supervisor/supervisor.go
```diff
@@ -176,6 +174,8 @@ func (sv *supervisor) updateWorker(IP string, leaderIP string, etcdIPs []string)
 	 * So, we need to restart the container when the leader changes. */
 	sv.Remove(Ovncontroller)
 	sv.run(Ovncontroller, "ovn-controller")
+
+	sv.SetInit()
```
ejj: Why this change in this patch?

minion/supervisor/supervisor.go
```diff
@@ -54,165 +50,38 @@ var images = map[string]string{
...
-	sv.runSystem()
-}
+func Run(_conn db.Conn, _dk docker.Client, _role db.Role) {
+	conn = _conn
+	dk = _dk
+	role = _role
 
-// Manage system infrstracture containers that support the application.
-func (sv *supervisor) runSystem() {
 	imageSet := map[string]struct{}{}
```
ejj: This imageSet stuff is only used on the worker right?

I'm not sure it makes sense to have a shared Run function,  Why not just have a RunWorker and RunMaster function and let the outside module call the appropriate one?  Slightly cleaner.

minion/supervisor/worker.go
```diff
@@ -0,0 +1,128 @@
...
+	setupWorker()
+	SetInit()
+	go runWorkerSystem()
+}
+
+func setupWorker() {
+	run(Ovsdb, "ovsdb-server")
+	run(Ovsvswitchd, "ovs-vswitchd")
+
+	for {
```
ejj: Both of these for loops are going to spin hard in the case of an error.  You need at least to sleep between iterations.

ejj: Also, have you tried booting a cluster and running the log test?  I'd be worried that it will return an error if ovsdb-server isn't ready yet.  May be fine though, not sure either way.

db/minion.go
```diff
@@ -10,7 +10,6 @@ type Minion struct {
 	Self           bool   `json:"-"`
 	Spec           string `json:"-" rowStringer:"omit"`
 	AuthorizedKeys string `json:"-" rowStringer:"omit"`
-	SupervisorInit bool   `json:"-"`
```
ejj: Do the previous patches guarantee that OVS is ready before anything else runs?  I.E. is it safe to remove the init flag?

Also this definitely needs a run of the log test locally to be confident there aren't warnings.

ejj:
@Baisang -- the series looks great.  Couple of small nits here and there.  I'd fix them up (and make sure you rebase) and then I'll merge it.

minion/supervisor/supervisor.go
Baisang: I think both worker and master use it (only in this function) to `dk.Pull` the container images they need. I had the `Run` be shared since they need to pull the same images (`Etcd`, `quilt/ovs`) before starting everything else up. So let me know what you think would be best.

minion/supervisor/supervisor.go
Baisang: So these changes are kind of unnecessary now since I ended up refactoring all of `supervisor` later, if you mean why I added `sv.SetInit()` back here it's because I later realized that the place I put it before was causing things to set up improperly a small percentage of time time, as `updateWorker` can return early without setting everything up properly.

db/minion.go
Baisang: Yeah, I was playing around with the location of calling `SetInit` and figured out that for Master, `ovsdb-server` needs to be running, and for Worker, the prerequisites are having `ovsdb-server`, `ovs-vswitchd`, and configuring the `quilt-int` bridge.

Will be sure to run log test

minion/supervisor/supervisor.go
Baisang: will be changed in the refactored supervisor, is it necessary to change it for this specific commit as well?

db/minion.go
Baisang: Ah, I get a bunch of WARNINGS from `Etcd`. Full output below:

https://gist.github.com/Baisang/00370a4948c8bf8a09add3913e2c6ba2

Will be taking a look now.

Baisang: Know where the problem is, will fix. Basically wasn't updating `oldEtcdIPs` when I should have been.

cluster/cloudcfg/cloudcfg.go
```diff
@@ -12,18 +14,22 @@ const (
...
 	err := t.Execute(&cloudConfigBytes, struct {
 		QuiltImage    string
 		UbuntuVersion string
 		SSHKeys       string
+		Role          string
 	}{
 		QuiltImage:    quiltImage,
 		UbuntuVersion: version,
 		SSHKeys:       strings.Join(keys, "\n"),
+		Role:          r,
```
kklin: Why not just `string(role)` here?

minion/supervisor/master_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/quilt/quilt/db"
+)
+
+func TestNone(t *testing.T) {
```
kklin: Is this test name still appropriate?

quiltctl/command/minion.go
```diff
@@ -1,16 +1,40 @@
...
+	"os"
 
+	"github.com/quilt/quilt/db"
 	"github.com/quilt/quilt/minion"
 )
 
 // Minion contains the options for running the Quilt minion.
-type Minion struct{}
+type Minion struct {
+	common *commonFlags
```
kklin: I don't think you actually use any of the options in `commonFlags`

quiltctl/command/minion.go
```diff
@@ -20,6 +44,24 @@ func (mCmd *Minion) Parse(args []string) error {
...
-	minion.Run()
+	if err := mCmd.run(); err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		return 1
+	}
 	return 0
 }
+
+func (mCmd *Minion) run() error {
+	if mCmd.role == "" {
```
kklin: We have a function that parses a role already, might want to just use that: https://github.com/quilt/quilt/blob/master/db/constants.go#L77

cluster/cluster.go
```diff
@@ -262,7 +262,10 @@ func (clst cluster) join() (joinResult, error) {
 			dbm := pair.L.(db.Machine)
 			m := pair.R.(machine.Machine)
 
-			dbm.CloudID = m.ID
+			if m.Role != db.None && m.Role == dbm.Role {
```
kklin: Just a question for my own understanding, but can a cloud machine's role ever be `None`?

kklin:
Super excited for this btw, makes the code cleaner _and_ is more correct.

cluster/cloudcfg/cloudcfg.go
Baisang: you're right, will fix. I think that was remnant for when I was debugging something.

minion/supervisor/master_test.go
Baisang: seems this test isn't necessary anymore, `TestMaster` should do the same

Baisang: or nevermind, I think this is now a case you are a Master but not the Leader. A name change is in order

quiltctl/command/minion.go
Baisang: Not sure what you mean? I thought I was passing role down like `quilt minion -role=master`

kklin: The commonFlags struct is for setting up the flags for some common options (like 'host'). It's not required for parsing other flags, e.g. 'Role'

Baisang: I see. I don't really understand how to set up the role flag then, since right now my implementation seems to be attaching it to `commonFlags`.

cluster/cluster.go
Baisang: Yeah, so `cluster` will try to get a list of machines from the CloudProviders, which will not have the `Role` field populated. So until the `foreman` is able to figure out their `Role`, the `Role` will be `db.None`

quiltctl/command/minion.go
kklin: I don't think your implementation uses `commonFlags` at all. It just references `mCmd.role`. I think if you just removed all references to `commonFlags` the parsing would still work.

Baisang:
ready again @ejj @kklin assuming travis is okay with it

kklin:
@ejj just to be clear, I'm not actually reviewing it, I was just reading it and noticed some things. You should merge whenever you're happy with it.

PR #765, opened 2017-02-07
Migrate OVS port creation to the plugin

ejj:
This PR refactors the code in preparation for the final patch that migrates the creating of OVS ports to the plugin.  By using ovs-vsctl to create the ports in the plugin, we can block container boot until OVS is configured.


tejasmanohar:
Will review this evening!

ejj:
I'm actually making some changes to the series for openflow.  Will reopen later.

PR #766, opened 2017-02-07
codecov: Ignore all files containing "mock"

ejj:
Now that we're using mockery more aggressively the codebase is strewn
with auto-generated mock files that shouldn't be counted toward code
coverage.


ejj:
quilt-bot smiles on @Baisang

PR #767, opened 2017-02-10
tests: Fix bandwidth bug where atoi parsed float

sirspinach:
The 15-bandwidth tests made an incorrect assumption that bandwidth
would always be returned as an integer string. Consequently, it
reported bandwidth incorrectly when the bandwidth had a decimal
part.


kklin:
In a similar vein to @kayousterhout, why not just make `bandwidthMbPerSec` a float? Seems most correct.

PR #769, opened 2017-02-12
Move some network configuration into the plugin

ejj:
This series moves OVSDB port creation, and some OpenFlow setup into the plugin.  Once this series is in, containers will be able to resolve their dependencies the first time they try.


api/client/mocks/Getter.go
```diff
@@ -1,7 +1,7 @@
 package mocks
 
-import "github.com/NetSys/quilt/api/client"
-import "github.com/stretchr/testify/mock"
+import client "github.com/NetSys/quilt/api/client"
+import mock "github.com/stretchr/testify/mock"
```
PenguinToast: Is there any reason we name the imports here?

minion/network/openflow/openflow.go
```diff
@@ -126,10 +127,17 @@ Table_2 {
...
-	VethPort  int
-	Mac       string
+// A Container that needs OpenFlow rules installed for it.
+type Container struct {
+	Veth  string
+	Patch string
+	Mac   string
+}
+
+type container struct {
```
kklin: I find the fact that there's an exported and unexported `container` with the same fields somewhat confusing. Maybe we could tweak the name?

api/client/mocks/Getter.go
ejj: Ah this is auto-generated code.  Apparently the library that generated it changed.

minion/network/openflow/openflow.go
ejj: Meh, I actually like the name.  It's literally the internal version of a container.

PR #770, opened 2017-02-13
nginx: Add setImage function

LuiseV:
This is needed for deploying the website and will eventually be needed for other users as well.


ejj:
meh these codecov failures are annoying

PR #773, opened 2017-02-14
specs: Migrate Nginx and Django to github.com/quilt

LuiseV:


LuiseV:
Yeah, I just need to rebase. Ethan removed `spec_tests.go`after I submitted my PR. 
But I realized afterward that the Django spec didn't work, so I decided to not update the PR until that's fixed.

PR #775, opened 2017-02-15
quiltctl: Added 'quilt switch' command

stevenlin598:
'quilt switch <namespace>' provides the ability to switch to another
namespace without manually calling 'quilt run <spec>' again.

### DIFF
--- a/quiltctl/command/switch.go
+++ b/quiltctl/command/switch.go
@@ -1,0 +1,113 @@
+package command
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/quilt/quilt/api"
+	"github.com/quilt/quilt/api/client"
+	"github.com/quilt/quilt/api/client/getter"
+	"github.com/quilt/quilt/cluster"
+	"github.com/quilt/quilt/cluster/machine"
+)
+
+// Switch contains the options for switching namespaces
+type Switch struct {
+	namespace    string
+	common       *commonFlags
+	clientGetter client.Getter
+}
+
+// NewSwitchCommand creates an instance of the Switch command.
+func NewSwitchCommand() *Switch {
+	return &Switch{
+		common:       &commonFlags{},
+		clientGetter: getter.New(),
+	}
+}
+
+// For mock testing
+var machineGetter = cluster.GetMachines
+
+// InstallFlags sets up parsing for command line flags.
+func (sCmd *Switch) InstallFlags(flags *flag.FlagSet) {
+	sCmd.common.InstallFlags(flags)
+
+	flags.Usage = func() {
+		fmt.Println("usage: quilt switch <namespace>")
+		fmt.Printf("`switch` switches to a different namespace\n")
+		flags.PrintDefaults()
+	}
+}
+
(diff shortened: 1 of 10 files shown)


ejj:
I'd like to see higher code coverage on this.  Btw, easy to check with `make coverage`

quiltctl/command/switch.go
```diff
@@ -0,0 +1,111 @@
...
+	return nil
+}
+
+// Run switches the current cluster to the specified namespace.
+func (sCmd *Switch) Run() int {
+	log.Info("Switching to namespace " + sCmd.namespace)
+
+	machines, err := machineGetter(sCmd.namespace)
+	if err != nil {
+		log.Error(err)
```
LuiseV: What's the reason for not returning 1 here?

stevenlin598: Fixed!

stevenlin598:
Added a couple more tests. Also, there's a bug where if the daemon is already running a spec and 'quilt switch' is executed, the foreman might still be operating on the old machines if the cluster hasn't updated the machine table yet. This happens if the namespace changes during a cluster run and clst.runOnce() aborts, but the foreman.runOnce() still executes.
https://github.com/quilt/quilt/blob/master/cluster/cluster.go#L82

kklin:
ping @LuiseV

LuiseV:
@kklin and @stevenlin598 , does the bug not matter? Should I review anyways?

kklin:
@LuiseV that bug is pre-existing, and will require some work to fix up. I think we should just get this patch in.

@ejj may disagree.

ejj:
I think we can go ahead and merge this, but we should file a high priority trello ticket on the issue.

LuiseV:
@ejj Don't merge quite yet. I'm reviewing and have a few nits.

stevenlin598:
Fixed!

kklin:
@ejj can you review this instead of me? I think it'd be good for him to get feedback from people other than me.

ejj:
Will do, but may take a while

ejj:
So overall the patch looks good, but I have some high level feedback:

Something about this feels a bit fragile.  I don't think it will work if the namespace changes while machines are booting, plus we have all of the issue of somehow guaranteeing the machine we query is actually a member of the namespace we think it is.

I had a thought that may turn out to be much simpler.  What if we just had the quilt daemon maintain a map from each namespace to its corresponding spec?  That way, if `quilt switch` is called we can just lookup the appropriate namespace in that map, and switch very easily.  The only downside to this is if the quilt daemon restarts, `quilt switch` won't work.  But that doesn't seem like much of an issue, we can simply say to do a `quilt run` (and in future patches we can store the mapping to a file).

Feels a bit more robust.  What do you think?

stevenlin598:
It definitely feels like a more reliable and simpler solution. We would also lose the ability to ```quilt switch``` to a spec started by a different daemon as well then?

ejj:
Yep, I think we'd lose that ability, but that seems fine to me.  In the default case people are going to leave the `quilt daemon` running.  Also, eventually we'll end up adding some sort of persistent storage (i.e. a manifest file of some sort) that could alleviate the problem.

ejj:
@steven, closing this for now.  Reopen when the next version is ready

PR #777, opened 2017-02-17
quiltctl: Make `quilt logs` error on scheduled container

PenguinToast:
Previously, when running `quilt logs` on a scheduled but not yet
running container, you would get a garbage error about docker. This
fix makes it clear to the user what the actual error is.


kklin:
@PenguinToast can you check if this is a problem for `quilt ssh` too?

PenguinToast:
@kklin Just checked, problem exists there too - I'll open up a new PR for that

PR #778, opened 2017-02-18
quiltctl: Show a reasonable error on trying to use an encrypted key

PenguinToast:
Previously, using `quilt ssh` or `quilt logs` with an encrypted private
key would show the error `asn1: structure error: length too large`,
which made the actual issue unclear. We now tell the user that
encrypted keys are unsupported and recommend they add the key to
`ssh-agent`.

### DIFF
--- a/quiltctl/ssh/native.go
+++ b/quiltctl/ssh/native.go
@@ -1,6 +1,7 @@
 package ssh
 
 import (
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


quiltctl/ssh/native.go
```diff
@@ -224,7 +225,15 @@ func signerFromFile(file string) (ssh.Signer, error) {
 		return nil, err
 	}
 
-	key, err := ssh.ParsePrivateKey([]byte(fileStr))
+	// Parse the key to give a helpul error when it's encrypted.
+	keyBytes := []byte(fileStr)
+	block, _ := pem.Decode(keyBytes)
+	if block != nil && block.Headers["Proc-Type"] == "4,ENCRYPTED" {
+		return nil, errors.New("ssh: password protected keys are " +
+			"not supported, try adding the key to ssh-agent first")
```
ejj: Is the command to add to ssh-agent consistent across platforms?  If so, it may be worth just saying what it is.   Most users don't likely remember it off the top of their head.

tejasmanohar: `eval "$(ssh-agent -s)"` should work everywhere

PenguinToast: I'll add `ssh-add` to the comment - I think `eval "$(ssh-agent -s)"` only runs `ssh-agent`, but doesn't actually add any keys.

PR #781, opened 2017-02-19
stitch: Omit ssh keys from machine hash calculation

mmussomele:


stitch/bindings.js
```diff
@@ -23,12 +26,19 @@ function Deployment(deploymentOpts) {
     this.invariants = [];
 }
 
+function omit_sshkey(key, value) {
```
Baisang: nit as we don't really follow a specific JS style yet, but if we're going to camelcase should be `omitSshKey` or something (man that looks ugly though)

stitch/bindings.js.go
```diff
@@ -27,12 +30,19 @@ function Deployment(deploymentOpts) {
     this.invariants = [];
 }
 
+function omitSSHKey(key, value) {
+    if (key == sshKeysField) {
```
kklin: I don't think `sshKeysField` needs to be a variable, it'd be slightly easier to read if you just used a string constant here IMO

quiltctl/command/run_test.go
```diff
@@ -38,7 +38,7 @@ func TestRunSpec(t *testing.T) {
 	stitch.DefaultImportGetter.Path = "/quilt_path"
```
kklin: Can you add why we made this change to the commit message?

stitch/bindings.js.go
```diff
@@ -27,12 +30,19 @@ function Deployment(deploymentOpts) {
...
+        return undefined;
+    }
+    return value;
+}
+
 // key creates a string key for objects that container a _refID, namely Containers
```
kklin: While you're in here, can you fix this typo? ("objects that _container_ a _refID" -> "objects that _have_ a _refID")

stitch/stitch_test.go
```diff
@@ -38,17 +38,40 @@ func TestMachine(t *testing.T) {
 				SSHKeys:  []string{"key1", "key2"},
 			}})
 
+	checkMachines(t, `deployment.deploy([new Machine({
```
kklin: I suspect we'll forget what this is testing, a comment would be helpful.

mmussomele:
@kklin updated with your comments.

PR #783, opened 2017-02-19
scheduler: Block container boots until OVN ready

kklin:
This patch delays booting containers on Workers until we can install
OpenFlow rules. Before, we logged warnings from trying to prematurely
install OpenFlow rules.


kklin:
@ejj can you take a look at this? It should fix the CI error we discussed on Slack.

PR #788, opened 2017-02-21
ci: Test POSTed data in mean test

Baisang:
Adds new test to POST data to the quilt/mean-service, and then ensures that this
data is available.

Basically all this test does is just post 10x from each public IP, and then checks on each public IP if all the data is available. Not super great since I think the haproxy instances can all just route to the same instance or something, but we can always increase the amount of times we post/get.

### DIFF
--- a/quilt-tester/tests/30-mean/check_mean.go
+++ b/quilt-tester/tests/30-mean/check_mean.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"os/exec"
@@ -12,6 +14,13 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 )
+
+// Response represents the JSON response from the quilt/mean-service
+type Response struct {
+	Text string `json:"text"`
+	ID   string `json:"_id"`
+	V    int    `json:"__v"`
+}
 
 func main() {
 	clientGetter := getter.New()
@@ -37,7 +46,17 @@ func main() {
 		log.WithError(err).Fatal("FAILED, couldn't query machines")
 	}
 
-	if logContainers(containers) && httpGetTest(machines, containers) {
+	publicIPs := getPublicIPs(machines, containers)
+	fmt.Printf("Public IPs: %s\n", publicIPs)
+	if len(publicIPs) == 0 {
+		log.Fatal("FAILED, Found no public IPs")
+	}
+
+	logTest := logContainers(containers)
+	getTest := httpGetTest(publicIPs)
+	postTest := httpPostTest(publicIPs)
+
+	if logTest && getTest && postTest {
 		fmt.Println("PASSED")
 	} else {
 		fmt.Println("FAILED")
@@ -58,14 +77,12 @@ func logContainers(containers []db.Container) bool {
 	return !failed
 }
 
-func httpGetTest(machines []db.Machine, containers []db.Container) bool {
-	fmt.Println("HTTP Get Test")
+func getPublicIPs(machines []db.Machine, containers []db.Container) []string {
 
 	minionIPMap := map[string]string{}
 	for _, m := range machines {
 		minionIPMap[m.PrivateIP] = m.PublicIP
 	}
-
 	var publicIPs []string
 	for _, c := range containers {
 		if strings.Contains(c.Image, "haproxy") {
@@ -78,10 +95,12 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
 		}
 	}
 
-	fmt.Printf("Public IPs: %s\n", publicIPs)
-	if len(publicIPs) == 0 {
-		log.Fatal("FAILED, Found no public IPs")
-	}
+	return publicIPs
+}
+
+func httpGetTest(publicIPs []string) bool {
+
+	fmt.Println("HTTP Get Test")
 
 	var failed bool
 	for i := 0; i < 25; i++ {
@@ -95,10 +114,86 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
 
 			if resp.StatusCode != 200 {
 				failed = true
+				fmt.Println(resp)
 			}
-			fmt.Println(resp)
 		}
 	}
 
 	return !failed
+}
+
+// checkInstances queries the todos for each instance and makes sure that all
+// data is available from each instance.
+func checkInstances(publicIPs []string, expectedTodos int) bool {
+	var todos []Response
+	failed := false
+	for _, ip := range publicIPs {
+		endpoint := fmt.Sprintf("http://%s/api/todos", ip)
+		resp, err := http.Get(endpoint)
+		if err != nil {
+			log.WithError(err).Error("HTTP Error")
+			failed = true
+			continue
+		}
+
+		if resp.StatusCode != 200 {
+			failed = true
+			fmt.Println(resp)
+		}
+
+		decoder := json.NewDecoder(resp.Body)
+		err = decoder.Decode(&todos)
+		if err != nil {
+			log.WithField("ip", ip).WithError(err).Error(
+				"JSON decoding error")
+			failed = true
+			continue
+		}
+
+		defer resp.Body.Close()
+		if len(todos) != expectedTodos {
+			log.WithFields(log.Fields{
+				"ip":       ip,
+				"expected": expectedTodos,
+				"actual":   len(todos),
+			}).Error(
+				"POSTed data not consistent for endpoint")
+			failed = true
+			continue
+		}
+	}
+
+	return !failed
+}
+
+// httpPostTest tests that data persists across the quilt/mean-service.
+// Data is POSTed to each instance, and then we check from all instances that
+// all of the data can be recovered.
+func httpPostTest(publicIPs []string) bool {
+
+	fmt.Println("HTTP Post Test")
+
+	var failed bool
+	for i := 0; i < 10; i++ {
+		for _, ip := range publicIPs {
+			endpoint := fmt.Sprintf("http://%s/api/todos", ip)
+
+			jsonStr := fmt.Sprintf("{\"text\": \"%s-%d\"}", ip, i)
+			jsonBytes := bytes.NewBufferString(jsonStr)
+
+			resp, err := http.Post(endpoint, "application/json", jsonBytes)
+			if err != nil {
+				log.WithError(err).Error("HTTP Error")
+				failed = true
+				continue
+			}
+
+			if resp.StatusCode != 200 {
+				failed = true
+				fmt.Println(resp)
+			}
+		}
+	}
+
+	return checkInstances(publicIPs, 10*len(publicIPs)) && !failed
 }


quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -96,9 +111,87 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
+	for _, ip := range publicIPs {
+		endpoint := fmt.Sprintf("http://%s/api/todos", ip)
+		resp, err := http.Get(endpoint)
+		if err != nil {
+			log.WithError(err).Error("HTTP Error")
+			failed = true
+			continue
+		}
+
+		if resp.StatusCode != 200 {
```
kayousterhout: Any reason not to log an error here too? If the test fails here, a log message seems useful / necessary to debug

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -96,9 +111,87 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
+		if err != nil {
+			log.WithError(err).Error("JSON decoding error")
+			failed = true
+			continue
+		}
+
+		defer resp.Body.Close()
+		if len(todos) != counter {
+			log.WithField("ip", ip).Error(
+				"POSTed data not consistent for endpoint")
```
kayousterhout: Include the actual + expected length here?

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -96,9 +111,87 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
+		}
+
+		if resp.StatusCode != 200 {
+			failed = true
+		}
+
+		decoder := json.NewDecoder(resp.Body)
+		err = decoder.Decode(&todos)
+		if err != nil {
+			log.WithError(err).Error("JSON decoding error")
```
kayousterhout: why not `.WithFiled("ip", ip)` here?

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -96,9 +111,87 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
+
+			if resp.StatusCode != 200 {
+				failed = true
+			}
+
+			fmt.Println(resp)
+		}
+	}
+
+	return checkInstances(publicIPs, counter) && !failed
```
kayousterhout: do !failed here first?  It seems like there's no point in doing checkInstances if posting the data failed

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -96,9 +111,87 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
+func httpPostTest(machines []db.Machine, containers []db.Container) bool {
+	fmt.Println("HTTP Post Test")
+	publicIPs := getPublicIPs(machines, containers)
+	fmt.Printf("Public IPs: %s\n", publicIPs)
+	if len(publicIPs) == 0 {
+		log.Fatal("FAILED, Found no public IPs")
+	}
+
+	var failed bool
+	counter := 0
```
kayousterhout: What about calling this nextTodoID instead?  That would make it more obvious how this is being used.

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -96,9 +111,87 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
 			fmt.Println(resp)
 		}
 	}
 
 	return !failed
 }
+
+// checkInstances queries the todos for each instance and makes sure that all
+// data is available from each instance.
+func checkInstances(publicIPs []string, counter int) bool {
```
kayousterhout: how about expectedNumTodos instead of counter?

quilt-tester/tests/30-mean/check_mean.go
Baisang: makes sense, will change

quilt-tester/tests/30-mean/check_mean.go
Baisang: the response still gets printed, so there'll be a 502/some other error in the output at least. but I see what you're saying, I was just copying the other test which doesn't have an error for that right now but we can do that if you still think it's a good idea.

kayousterhout: Yeah actually I was wondering if it would make sense to not print the response in general, and only print it when there's an error?  Not sure how noisy the quilt tests are already but it seems unnecessary to print the response even when things succeed.

Baisang: it's pretty noisy, so I think that would be a good change. will implement

Baisang:
@kayousterhout

Baisang:
seems to work locally with quilt-tester and mean example i just booted up

quilt-tester/tests/30-mean/check_mean.go
kayousterhout: One more Q here: should this be log.Error instead of a regular Println?

kayousterhout: or log.WithField("ip", ip) as below?

Baisang: Hmm, so what I perceive to be current pattern in tests is we `Println` if it's more "informational" and use `log.Error` if it's actually an error message. The reason why is because before we were just using `logrus` for everything in `check_mean.go` and it wasn't displaying newlines properly. So I made a fix to switch some of the more "informational" calls to `Println` so it could be displayed properly. (the reason why it wasn't displaying properly is because `logrus` outputs things in a weird format when it's not attached to a terminal)

Now obviously this distinction is pretty arbitrary so either would probably be fine (I don't perceive the HTTP responses to have content that needs to show up on multiple lines). Let me know what you think!

kayousterhout: I see that's helpful context -- since this is the pattern that's used elsewhere that seems fine.  Thanks for the explanation!

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -37,7 +46,8 @@ func main() {
 		log.WithError(err).Fatal("FAILED, couldn't query machines")
 	}
 
-	if logContainers(containers) && httpGetTest(machines, containers) {
+	if logContainers(containers) && httpGetTest(machines, containers) &&
```
kklin: I don't think the tests should short-circuit if a previous one failed. I'd like to gather as much information as possible in the tests.

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -95,10 +110,92 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
+			failed = true
+			continue
+		}
 	}
 
 	return !failed
 }
+
+// httpPostTest tests that data persists across the quilt/mean-service.
+// Data is POSTed from each instance, and then we check from all instances that
```
kklin: "POSTed _to_ each instance", right?

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -78,6 +86,13 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
 		}
 	}
 
+	return publicIPs
+}
+
+func httpGetTest(machines []db.Machine, containers []db.Container) bool {
+	fmt.Println("HTTP Get Test")
+
+	publicIPs := getPublicIPs(machines, containers)
```
kklin: Could we just make `httpGetTest` and `httpPostTest` take in the `publicIPs`, and generate them once in the caller? It'd remove some redundant code. BTW, I'd also pull the `client.QueryFoo` stuff into a method along with the error checking for number of IPs into a method that returns `[]string, error`, and then `log.Fatal` if an errors returned. Should clean up some boilerplate.

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -95,10 +110,92 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
+		log.Fatal("FAILED, Found no public IPs")
+	}
+
+	var failed bool
+	nextTodoID := 0
+	for i := 0; i < 10; i++ {
+		for _, ip := range publicIPs {
+			endpoint := fmt.Sprintf("http://%s/api/todos", ip)
+
+			jsonStr := fmt.Sprintf("{\"text\": \"%d\"}", nextTodoID)
```
kklin: What if we used "$ip-$i" and ditched `nextTodoID`? Slightly cleaner I think, and more informative if you're clicking around the app.

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -95,10 +110,92 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
+			}
+
+			if resp.StatusCode != 200 {
+				failed = true
+				fmt.Println(resp)
+			}
+		}
+	}
+
+	return !failed && checkInstances(publicIPs, nextTodoID)
```
kklin: I actually disagree with @kayousterhout here, I prefer collecting too much information in tests rather than not enough. For example, if we're failing to POST to one node, I'd still be interested in whether we can GET the data from the other nodes.

quilt-tester/tests/30-mean/check_mean.go
```diff
@@ -95,10 +110,92 @@ func httpGetTest(machines []db.Machine, containers []db.Container) bool {
...
+			failed = true
+			continue
+		}
+
+		if resp.StatusCode != 200 {
+			failed = true
 			fmt.Println(resp)
 		}
+
+		decoder := json.NewDecoder(resp.Body)
```
kklin: Just `json.Unmarshal`? I also have a small preference for the &todos to be declared near the decoding.

quilt-tester/tests/30-mean/check_mean.go
Baisang: sure, that sounds good to me

quilt-tester/tests/30-mean/check_mean.go
Baisang: yes

quilt-tester/tests/30-mean/check_mean.go
kayousterhout: Sorry to mis-direct you here Brian!

quilt-tester/tests/30-mean/check_mean.go
Baisang: Decoders work on streams (type `io.Reader`) while `Unmarshal` is for data already in memory. Response.body is a stream, we can convert it to a string in memory if you would like to though.

kklin: I see, that makes sense! No, I agree the decoder is better.

quilt-tester/tests/30-mean/check_mean.go
Baisang: not quite sure what you mean by pulling all of the client.QueryFoo into another method...how would this reduce boilerplate?

kklin: Eh, looking it over it actually wouldn't remove code, would just be slightly cleaner. I'm fine if we just move up the getPublicIPs call. But I was thinking

```
func main() {
  machines, containers, err := getMachinesAndContainers()
  if err != nil { ... }
  publicIPs, err := getPublicIPs(machines, containers)
}
```

Baisang: I see, the thing is that if `getMachinesAndContainers()` fails, I think we would want to know where exactly it failed. In that case, makes sense for us to exit early with `log.Fatal`, which makes putting it in as separate function that returns an error kinda redundant (i.e. it shouldn't reach the `if err != nil` in `main` at that point.

We can avoid exiting early by not using `log.Fatal`, but then it's not as immediately apparent what part failed.

kklin: Yeah, that's true. The normal approach would be to modify the return error with some context (`return nil, nil, fmt.Errorf("get connections: %s", err)`). I think the reason I have this instinct that logging the returned error is cleaner because it makes it possible to unit test -- you can't unit test code that exits.

I don't think it's worth the effort for this code, though, I'd much rather just get the new test in.

Baisang:
@kklin sorry forgot to ping you, ready to review again

Baisang:
@kklin updated, thanks!

PR #792, opened 2017-02-26
WIP -- network: Refactor NAT logic with IPTables library

sirspinach:
Last time I built a new quilt image using this code, I observed that iptables failed to install inside minion containers on worker machines. (`docker logs minion` contains "Couldn't find iptables"). Not sure if that was a configuration issue on my end.

Can check with `iptables -S nat` to see whether rules were properly updated. Someone will need to confirm that networking is working properly before this is merged.


kklin:
@sirspinach I saw that in the past when I wasn't using the most recent `quilt/ovs` image when building. Does `docker images quilt/ovs` show an old image?

BTW, I'm sort of out of the loop on your plans for Quilt. Would you like me to finish this up? It's relevant to me because I'm messing with iptables for something for @ejj, so I'd like to just get this merged soon if possible.

sirspinach:
that might have been what happened. Yeah, it would be great if you finished
this! I don't have enough time for both Quilt and Nefeli.

minion/network/nat.go
```diff
@@ -48,65 +55,84 @@ func updateNAT(containers []db.Container, connections []db.Connection) {
...
-			return nil, fmt.Errorf("failed to get current IP rules: %s", err)
+			return ipRuleSlice{}, err
 		}
-		rules = append(rules, rule)
-	}
 
-	if err := scanner.Err(); err != nil {
-		return nil, fmt.Errorf("scanner error while getting IP tables: %s", err)
+		for _, r := range rawRules {
+			// In iptables v1.6.0, policies for chains in the nat table
```
kklin: @sirspinach Does this change just make the code a bit simpler? Seems to me like we could just extend the logic to apply to chains also, and not have this assumption that may change in the future.

sirspinach: @kklin yes, the motivation behind this assumption was to remove these 4 policy rules: https://github.com/quilt/quilt/pull/792/files/e352fdb19b5bbd52b4fddaf0c51967211e97528d#diff-47a66ada7d55780d3bc7d9fc374099a2L102

kklin:
I'll open a new PR once it's ready.

PR #793, opened 2017-02-28
stitch: Support writing files to containers

kklin:


kayousterhout:
Also can you document this in the README as per slack discussion?

inspect/inspect_test.go
```diff
@@ -46,13 +46,13 @@ const testStitch = `var a = new Service("a", [new Container("ubuntu")]);
...
-        bf2a4d57f5842c52610dde4e093ef02815d5129c;
+        0df0539843c63da9000c01778dea06b22ecb365a;
+        d8c0112b665631b5eafdafd5113789a48759bd8e;
+        f90f6f4228ff2b6187a857ee0ca517986dd642e1;
         public;
     }
-    6ea7e010f11bb105bf84de4df7ddb1d06468f333 -> b7b8a22969314f4e0b7fd4935049d02233fada0b
-    b7b8a22969314f4e0b7fd4935049d02233fada0b -> bf2a4d57f5842c52610dde4e093ef02815d5129c
+    d8c0112b665631b5eafdafd5113789a48759bd8e -> 0df0539843c63da9000c01778dea06b22ecb365a
+    f90f6f4228ff2b6187a857ee0ca517986dd642e1 -> d8c0112b665631b5eafdafd5113789a48759bd8e
```
kklin: This is the output of the `quilt inspect` subcommand that Akshay worked on. The hashes are StitchIDs of the containers the test is visualizing.

Were you thinking a comment for the variable `expGraph`? Or comments within the graph that's outputted?

I'm wary of putting too much time into this, though, as I don't think anyone uses it.

kayousterhout: I just realized here "exp" is "expected" (I was somehow thinking exponent or experiment) -- so either changing this to expectedGraph or commenting `The expected graph returned by inspect` would make this much more obvious I think

kklin:
@kayousterhout Addressed your comments

kklin:
Oh, I noticed that `ToTar` is already defined in `util`, so I switched to that.

Also added coverage for `util.MapAsString`.

stitch/README.md
```diff
@@ -0,0 +1,31 @@
...
+```
+new Container("haproxy").withFiles({
+  "/etc/myconf": "foo"
+});
+```
+would create a `haproxy` instance with a text file `/etc/myconf` containing `foo`.
+
+If the files change after the container boots, Quilt does not restart the container.
+However, if the file content specified by `filepathToContent` changes, Quilt will
+destroy the old container and boot a new one with the proper filesystem.
```
kayousterhout: can you put "files" here instead of "filesystem"? "filesystem" implies ext3 or something which is a little misleading.

minion/scheduler/worker.go
```diff
@@ -17,6 +19,7 @@ import (
 const labelKey = "quilt"
 const labelValue = "scheduler"
 const labelPair = labelKey + "=" + labelValue
+const filesysKey = "filesys"
```
kayousterhout: As below filesys is misleading here -- it's really the file structure, and the file system isn't changing.  Maybe just filesKey?

PR #796, opened 2017-03-01
IGNORE ME: debugging codecov

kklin:
Don't review.


kklin:
Looks fine now, I think it was failing because of the S3 outage.

PR #798, opened 2017-03-02
ci: Store bandwidth result as a float

kklin:
It mildly simplifies the code by removing an assumption about overflow.


kklin:
@Baisang fixed

ejj:
Sort of annoying quilt-bot keeps assigning me to kevin prs ... I should fix that.

PR #800, opened 2017-03-02
Add test for outbound public traffic

kklin:


minion/network/nat.go
```diff
@@ -154,13 +160,48 @@ func routingRules(publicInterface string, containers []db.Container,
...
+
+	// Map each label to all ports on which it can send packets
+	// to the public internet.
+	portsToWeb := make(map[string]map[int]struct{})
+	for _, conn := range connections {
+		if conn.To != stitch.PublicInternetLabel {
+			continue
+		}
+
+		if _, ok := portsToWeb[conn.To]; !ok {
```
kklin: Ahh good catch.. Fixed, and added a unit test that would've caught it.

PR #801, opened 2017-03-02
stitch: Fix bug with multiple identical services

PenguinToast:
Previously, when multiple identital services were deployed, you would
end up with inconsistent StitchID assignment causing bugs.

A demo of this bug can be found at
https://github.com/PenguinToast/quilt-demo, by first running
`spark/run.js`, then `spark/run2.js`. `run.js` deploys a single Spark
cluster, and `run2.js` deploys another identical spark cluster. With the
bug, the master container for the second cluster will be assigned the
original StitchID of the first cluster's master, causing all the new
workers to connect to the first cluster's master.

The bug arose from the `_refID` fields of containers being strings,
causing a container with a `_refID` of `"10"` to be assigned a StitchID
before a container with a `_refID` of `"8"`. We fix this bug by making
`_refID` a number so that StitchIDs are assigned in the correct order.


kklin:
Are you going to write an integration test that would have caught this too?
It'd be cool if we had a test that ran the first file, then randomly shuffled the second file.

PenguinToast:
@kklin I could write an integration test, but I'm not sure what we could randomize and what this test would test for beyond the changes I've already made in the unit tests

stitch/bindings.js
```diff
@@ -4,6 +4,9 @@ var deployment = new Deployment({});
 // The label used by the QRI to denote connections with public internet.
 var publicInternetLabel = "public";
 
+// Global unique ID counter.
```
mmussomele: Just a few message nits:
Type in commit message, "identital".
Maybe a comma after "StitchID assignment" to make the meaning a bit more clear. 
Also I'm not sure the link to the demo is necessary in the message, but I don't feel strongly about it.

stitch/bindings.js
```diff
@@ -58,7 +66,7 @@ function setQuiltIDs(objs) {
     // If there are multiple references to the same object, there will be duplicate
     // refIDs.
     Object.keys(refIDs).forEach(function(k) {
-        refIDs[k] = _.uniq(refIDs[k]).sort();
+        refIDs[k] = _.sortBy(_.uniq(refIDs[k]), _.identity);
```
mmussomele: I not very familiar with javascript, but why this change? If `_refID` is now an integer, won't this work as is?

stitch/bindings.js
```diff
@@ -409,7 +417,7 @@ function Container(image, command) {
     // refID is used to distinguish deployments with multiple references to the
     // same container, and deployments with multiple containers with the exact
     // same attributes.
-    this._refID = _.uniqueId();
+    this._refID = uniqueID();
```
mmussomele: Is it necessary to define a new function? Or is it safe to just transform the result of `_.uniqueId()` to an int?

mmussomele:
It seems like your demo could easily be transformed into an integration test. I feel like it's ok if it doesn't always fail (fails randomly), since eventually an integration run will catch it. Plus that way it could be remove from the commit message, as it seemed a little out of place there.

kklin:
The reason I'd like an integration test is that your unit test ensures the one failure mode that you debugged never happens. Whereas an integration test would ensure that running one service and then another identical one _always works_. If this scenario starts failing for a different reason, we'd find out with the integration test and not the unit test.

Ah, ignore the random shuffling comment, I confused myself 😄 

BTW, I agree with @mmussomele about linking to the demo. I'd prefer a simple inline example demonstrating the bug as that link might become out of date.

stitch/bindings.js
PenguinToast: Javascript's built in array sort sorts things lexicographically, so it would still sort refIDs[k] as if it contained strings.

stitch/bindings.js
PenguinToast: It's not strictly necessary, but both Ethan and I think this is a more safe way of doing it so that we don't have to rely on Underscore.js always returning integers.

PenguinToast:
@kklin Added the integration test - could you review it?

quilt-tester/tests/61-duplicate-cluster/check_duplicate_cluster.go
```diff
@@ -0,0 +1,76 @@
...
+	clnt, err := clientGetter.Client(api.DefaultSocket)
+	if err != nil {
+		log.WithError(err).Fatal("FAILED, couldn't get quiltctl client.")
+	}
+	defer clnt.Close()
+
+	leader, err := clientGetter.LeaderClient(clnt)
+	if err != nil {
+		log.WithError(err).Fatal("FAILED, couldn't get leader client.")
+	}
```
PenguinToast: Fixed

quilt-tester/tests/61-duplicate-cluster/check_duplicate_cluster.go
```diff
@@ -0,0 +1,76 @@
...
+
+	containers, err := leader.QueryContainers()
+	if err != nil {
+		log.WithError(err).Fatal("FAILED, couldn't query containers.")
+	}
+
+	containersPretty, _ := exec.Command("quilt", "containers").Output()
```
PenguinToast: Fixed

quilt-tester/tests/61-duplicate-cluster/check_duplicate_cluster.go
```diff
@@ -0,0 +1,76 @@
...
+		if workerCount != 3 {
+			failed = true
+			log.WithFields(log.Fields{
+				"master":       master,
+				"worker count": workerCount,
+			}).Error("FAILED, expected 3 workers connected to master")
+		}
+
+		fmt.Printf("`quilt logs %s` output:\n", master)
+		fmt.Println(string(logs))
```
PenguinToast: Fixed

Makefile
```diff
@@ -46,6 +46,8 @@ COV_SKIP= /api/client/mocks \
 	  /quilt-tester/tests/40-stop \
 	  /quilt-tester/tests/100-logs \
 	  /quilt-tester/tests/75-network \
+	  /quilt-tester/tests/60-duplicate-cluster-setup \
```
PenguinToast: Fixed

quilt-tester/tests/61-duplicate-cluster/check_duplicate_cluster.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	containersPretty, _ := exec.Command("quilt", "containers").Output()
+	fmt.Println("`quilt containers` output:")
+	fmt.Println(string(containersPretty))
+
+	var masters []string
+	for _, dbc := range containers {
+		if strings.Join(dbc.Command, " ") == "run master" {
+			id := fmt.Sprint(dbc.StitchID)
```
PenguinToast: Good question - I copied most of this test from `check_spark.go`. Fixed

quilt-tester/tests/61-duplicate-cluster/check_duplicate_cluster.go
```diff
@@ -0,0 +1,76 @@
...
+			masters = append(masters, id)
+			println("found id ", id)
```
PenguinToast: Fixed

quilt-tester/tests/61-duplicate-cluster/check_duplicate_cluster.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+	if len(masters) != 2 {
+		log.WithFields(log.Fields{
```
PenguinToast: Fixed

quilt-tester/tests/61-duplicate-cluster/check_duplicate_cluster.go
```diff
@@ -0,0 +1,76 @@
...
+	for _, master := range masters {
+		logs, err := exec.Command("quilt", "logs", master).CombinedOutput()
+		if err != nil {
+			log.WithError(err).Fatal(
+				"FAILED, unable to get Spark master logs.")
+		}
+
+		logsStr := string(logs)
+		workerCount := strings.Count(logsStr, "Registering worker")
+		if workerCount != 3 {
```
PenguinToast: Fixed

quilt-tester/tests/61-duplicate-cluster/duplicate-cluster.js
```diff
@@ -0,0 +1,15 @@
+var spark = require("github.com/quilt/spark");
+var infrastructure = require("github.com/quilt/quilt/quilt-tester/config/infrastructure");
+
+var deployment = createDeployment({});
+deployment.deploy(infrastructure);
+
+var sprk = new spark.Spark(1, 3);
+publicInternet.connect(8080, sprk.masters);
+var sprk2 = new spark.Spark(1, 3);
+publicInternet.connect(8080, sprk.masters);
```
PenguinToast: Fixed

PenguinToast:
@kklin Addressed PR comments

kklin:
Thanks @PenguinToast, did you see my comment about collapsing the tests into one?

PenguinToast:
Ah, missed that comment - collapsing the tests would be a little ugly, since we'd have to
 - Change how `quilt-tester` so that it chooses the correct spec to run first
 - Hard-code or somehow pass the test path to the test itself so that it can run the second spec

It's definitely do-able, but some discussion would be needed to figure out the "right" way to do the above two points

kklin:
Nah, I don't think it's worth the effort. Thanks for looking into it!

PR #805, opened 2017-03-07
tests: Add README and spec tests back in

LuiseV:
We temporarily removed the tests when we moved the Quilt specs to the new GitHub organization. 
This patch adds back in a modified version of the tests that work with the remote specs.


stitch/get.go
```diff
@@ -41,8 +41,7 @@ func GetQuiltPath() string {
 type ImportGetter struct {
 	Path         string
 	AutoDownload bool
-
-	repoFactory func(repo string) (repo, error)
```
kklin: I'd rather keep this hidden, and add a `NewImportGetter` function that's public and sets the `repoFactory` in the `ImportGetter` that it returns. `repoFactory` is really a quirk for unit testing.

LuiseV: Good point

scripts/specs-tester/specs-tester.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"os"
+	"sort"
+
+	"github.com/quilt/quilt/scripts/specs-tester/spec_tests"
+)
+
+func main() {
+	defer os.RemoveAll(spec_tests.QuiltPath)
```
kklin: Did you try using an `afero.MemMapFS`?

scripts/specs-tester/tests/test_specs.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/quilt/quilt/stitch"
+
+	"github.com/robertkrimen/otto"
+)
+
+// QuiltPath is the QUILT_PATH used for testing.
+var QuiltPath = "/tmp/.quilt"
+
+func configRunOnce(specPath string) error {
+	stitch.HTTPGet = func(url string) (*http.Response, error) {
```
kklin: Since we're hitting the network, I think we might as well let the github keys resolve.

scripts/specs-tester/tests/test_specs.go
```diff
@@ -0,0 +1,87 @@
...
+		resp := http.Response{
+			Body: ioutil.NopCloser(bytes.NewBufferString("")),
+		}
+		return &resp, nil
+	}
+
+	getter := stitch.NewImportGetter(QuiltPath)
+	getter.AutoDownload = true
+
+	if remote := isRemote(specPath); remote {
```
kklin: Can be simply `if isRemote(specPath)`

kklin:
I can setup the travis thing if you're busy

LuiseV:
Addressed your comments, @kklin. It would be great if you can set up the Travis. Thanks!

PR #810, opened 2017-03-11
ci: Add test for etcd

kklin:
This patch adds a test that boots an etcd cluster, and ensures the
cluster is healthy.

Whoever reviews this should also review https://github.com/quilt/etcd/pull/1. The `quilt/etcd` PR should be merged before this one.


quilt-tester/tests/etcd/etcd.js
```diff
@@ -0,0 +1,11 @@
+var etcd = require("github.com/quilt/etcd");
+var infrastructure = require("github.com/quilt/quilt/quilt-tester/config/infrastructure")
+
+var deployment = createDeployment({});
+deployment.deploy(infrastructure);
+
+var nWorker = deployment.machines.filter(function(m) {
+    return m.role == "Worker"
+}).length;
```
PenguinToast: It seems to me that it would make more sense to expose the `nWorkers` variable from `infrastructure.js` instead of doing this.

kklin: I added some commits to facilitate this, can you do another round?

PR #814, opened 2017-03-17
ovs: Build multiple kernel module versions in OVS container

PenguinToast:
Previously, the OVS container would only contain kernel modules for one
specific kernel version. This would cause the minion containers to fail
to install the OVS modules.

We now build kernel modules for multiple kernel versions, and install
the module corresponding to the kernel version on minion boot.


cluster/cloudcfg/template.go
```diff
@@ -16,9 +16,9 @@ initialize_ovs() {
 	ExecStartPre=/sbin/modprobe gre
 	ExecStartPre=/sbin/modprobe nf_nat_ipv6
 	ExecStart=/usr/bin/docker run --rm --privileged {{.QuiltImage}} \
-	bash -c "insmod /modules/openvswitch.ko \
-	         && insmod /modules/vport-geneve.ko \
-	         && insmod /modules/vport-stt.ko"
+	bash -c "insmod /modules/$(uname -r)/openvswitch.ko \
```
kklin: It'd be nice if we could print an error if we don't support the machine's kernel. Doesn't have to be in this commit though. Could you create a trello card on it? (or do it, if you feel it)

ovs/Dockerfile
```diff
@@ -1,20 +1,27 @@
 FROM ubuntu:16.04
 
-RUN ovs_ver="2.6.1" linux="linux-headers-4.4.0-62-generic" \
-build_deps="${linux} build-essential libssl-dev python python-six wget" \
+ENV ovs_ver="2.6.1"
+ENV kernel_versions="4.4.0-38-generic 4.4.0-62-generic"
```
kklin: Oh, does adding kernel versions make the image much bigger? If it does, I think we should just build 4.4.0-62 until we actually need another version.

PenguinToast: Only adds about 13 MB (271 -> 284 MB) - we could still remove it since there isn't any need to build 2 versions right now (except for having it as an example to make adding versions easier in the future).

kklin: Yeah, let's just remove it. Kind of misleading to have a version we don't need in here.

cluster/cloudcfg/template.go
PenguinToast: Created a trello card.

ovs/Dockerfile
PenguinToast: Done

PR #815, opened 2017-03-18
quiltctl: Add debug-logs command

mmussomele:
When developing quilt, we often have a need to see logs from several
different machines and containers, but it's a pain to log into each
and extract the logs we need. This patch adds a command that fetches
logs from a set of machines and containers and stores them in files
on the user's machine.

This has the nice side effect of giving users an easy way to get
logs that they can send us in the case that they're encountering
bugs that they can't figure out.


mmussomele:
Changes made.

quiltctl/command/debug.go
```diff
@@ -0,0 +1,363 @@
...
+		conn, err := dCmd.sshGetter(t.ip, dCmd.privateKey)
+		if err != nil {
+			errno++
+			log.Error(err)
+			continue
+		}
+
+		for name, cmd := range cmds {
+			dCmd.log(fmt.Sprintf("\tDownloading log: %s", name))
+			if t.container {
```
kklin: I get the general sense that this `target` data structure isn't quite right. It feels kind of fragile. I'd like to get rid of the `t.container` bool if possible.

Perhaps we could make `cmds` an attribute of the `target`? This way, we'd just generate different `cmds` in the `machinesToTargets` and `containersToTargets`.

quiltctl/command/debug.go
```diff
@@ -0,0 +1,364 @@
...
+		}
+
+		containers = append(containers, conts...)
+	}
+
+	mTargets := machinesToTargets(machines)
+	cTargets := containersToTargets(containers, ipMap)
+
+	var targets []logTarget
+	if dCmd.all || (dCmd.machines && dCmd.containers) {
```
mmussomele: It makes testing annoying if done in `Parse`. But it's cleaner to do that, so I'll set them in `Run`.

kklin: Why does it make testing annoying?

mmussomele: It makes sense to be able to create a Debug struct with `all` true and `machines` and `containers` false and then `Run` it for the tests. If we convert in Parse, then all tests for `Run` require `machines` and `containers` to explicitly be true.

mmussomele: Eh

mmussomele: Actually now that I think about it, having the tests set that explicitly would be fine.

kklin: Ah. I could see an argument setting it in `Run` because then Debug{all: true}.Run() would behave as expected

mmussomele: Yea that was my original thought, it's not a big deal though.

kklin: Your call

quiltctl/command/debug_test.go
```diff
@@ -434,3 +493,43 @@ func filesFor(id string, container bool) []string {
 
 	return exp
 }
+
+// A simple mock of filepath.Walk. This function is not a perfect implementation, but
+// should be good enough for our purposes.
+func virtualWalk(root string, walkFn filepath.WalkFunc) error {
```
kklin: Can we use this? https://godoc.org/github.com/spf13/afero#Walk

mmussomele:
Note: Coverage is low because of the untested `tarball` function. It'll be tested once my [Afero PR](https://github.com/spf13/afero/pull/118) is merged.

quiltctl/command/debug.go
```diff
@@ -0,0 +1,360 @@
...
+			continue
+		}
+
+		for _, cmd := range t.cmds {
+			log.Debugf(fmt.Sprintf("Downloading log '%s' for target %s",
+				cmd.name, t.stitchID))
+
+			result, err := conn.CombinedOutput(cmd.cmd)
+			if err != nil {
+				log.WithError(err).Debugf("Failed to get log '%s'"+
```
kklin: Why is this at `Debug` and the `WriteFile` error at `Error`? I think we might as well just log this at `ERROR` as well.

quiltctl/command/debug.go
```diff
@@ -0,0 +1,360 @@
...
+					" from target %s", cmd.name, t.stitchID)
+				continue
+			}
+
+			logFile := filepath.Join(path, cmd.name)
+			if err := util.WriteFile(logFile, result, 0666); err != nil {
```
kklin: Same here re permissions

quiltctl/command/debug.go
```diff
@@ -0,0 +1,360 @@
...
+	for _, m := range machines {
+		if m.PublicIP == "" {
+			continue
+		}
+
+		t := logTarget{
+			ip:       m.PublicIP,
+			dir:      machineDir,
+			stitchID: m.StitchID,
+			cmds:     []logCmd{},
```
kklin: Nit, but I'd just leave `cmds` as `nil`

kklin: Or even simpler, `cmds: machineCmds`

quiltctl/command/debug.go
```diff
@@ -0,0 +1,360 @@
...
+		if !ok {
+			log.Errorf("No machine with private IP %s", c.Minion)
+			continue
+		}
+
+		t := logTarget{
+			ip:       ip,
+			dir:      containerDir,
+			stitchID: c.StitchID,
+			cmds:     []logCmd{},
```
kklin: Same here re `nil`.

quiltctl/command/debug_test.go
```diff
@@ -0,0 +1,436 @@
...
+			ContainerReturn: test.containers,
+		}
+
+		mockClientGetter := new(mocks.Getter)
+		mockClientGetter.On("Client", mock.Anything).Return(mockLocalClient, nil)
+		testCmd.clientGetter = mockClientGetter
+
+		assert.Equal(t, test.expReturn, testCmd.Run())
+
+		if test.expReturn == 0 {
```
kklin: Seems weird to me to have a special case for this, why not just specify the `ps` file in `expFiles`?

quiltctl/command/debug.go
```diff
@@ -19,6 +22,7 @@ import (
 	"github.com/quilt/quilt/util"
 )
 
+// Stored in variables to be mocked out for the unit tests.
```
kklin: Should be in the previous commit

quiltctl/command/debug.go
```diff
@@ -358,3 +370,53 @@ func findTarget(targets []logTarget, id string) (logTarget, error) {
...
+		log.WithError(err).Error("Failed to close tar writer")
+		return 1
+	}
+
+	if err := util.RemoveAll(root); err != nil {
+		log.WithError(err).Error("Failed to remove temporary log directory")
+		return 1
+	}
+
+	if err := util.WriteFile(root+".tar", buf.Bytes(), 0666); err != nil {
```
kklin: Same here re permissions

quiltctl/command/debug_test.go
```diff
@@ -419,7 +479,6 @@ func TestDebug(t *testing.T) {
 }
 
 func filesFor(id string, container bool) []string {
-	debugFolder := "debug_logs_Mon_Jan_01_00-00-00"
```
kklin: Wrong commit?

quiltctl/command/debug.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	dCmd.machines = dCmd.machines || dCmd.all
+	dCmd.containers = dCmd.containers || dCmd.all
+
+	var targets []logTarget
+	mTargets := machinesToTargets(machines)
+	cTargets := containersToTargets(containers, ipMap)
+	if !(dCmd.machines || dCmd.containers) {
+		var err error
```
mmussomele: ah good catch

quiltctl/command/debug.go
```diff
@@ -0,0 +1,360 @@
...
+		errno++
+		log.Error(err)
+	} else if err := util.WriteFile(psFile, result, 0666); err != nil {
+		errno++
+		log.Error(err)
+	}
+
+	for _, t := range targets {
+		path := filepath.Join(rootDir, t.dir, t.stitchID)
+		if err := util.Mkdir(path, 0755); err != nil {
```
mmussomele: It will but there should never be two targets with the same stitchID

quiltctl/command/debug.go
mmussomele: If a log failed to download it could be a network error or something specific to the container/machine you were fetching from and the rest might still work. If a local disk write failed, all the writes are likely to fail, so I considered it more serious. Anyway, I think both at `Error` is fine though, failing to fetch from a machine could be a problem

quiltctl/command/debug.go
mmussomele: I'm a bit wary of just assigned as `machineCmds`. It's fine for now since we only read the slice, but for future proofing a copy feels safer.

quiltctl/command/debug.go
kklin: Oh right, makes sense

quiltctl/command/debug.go
kklin: I think `cmds: machineCmds` makes a copy of `machineCmds` before assigning? Is that what you were concerned about?

kklin:
@mmussomele is this ready for another round? Can you ping the afero folks btw?

mmussomele:
@kklin ah yes it is ready, sorry

kklin:
@mmussomele the build failed

mmussomele:
Ah I imported `github.com/quilt/quilt/minion/supervisor`, which can't build on windows. Should be fixed up now.

ejj:
Talked with Matt, he seems to think it's ready (and that Kevin would approve).  Since Kevin is out of town, I'm just going to merge it.

kklin:
Fine with me

mmussomele:
@ejj it should be ready now

PR #816, opened 2017-03-18
minion: Check that minions are running the same quilt as the daemon

mmussomele:
If the minions and the daemon are not running the same version of
quilt, strange errors can occur, such as stitch JSON failing to parse
remotely but working fine locally. To help mitigate this, the daemon
now sends a hash of the source it was compiled with to the minions,
who also have a hash of the source they were compiled with. If the
hashes don't match, the daemon will log a warning to assist debugging
efforts.


Makefile
```diff
@@ -15,14 +16,19 @@ LINE_LENGTH_EXCLUDE=./api/pb/pb.pb.go \
...
+all: sha
 	cd -P . && go build .
 
-install:
+install: sha
 	cd -P . && go install .
 
+sha: $(MAIN_SOURCE)
+	cat $(VERSION_HASH_TEMPLATE).txt > $(VERSION_HASH_TEMPLATE).go
+	echo -e "\nconst Hash = \`$$(shasum -a 256 $^ | sort -k2 | shasum -a 256 | cut -d' ' -f1)\`" >> $(VERSION_HASH_TEMPLATE).go
```
Baisang: nit but space after `-d` ?

also what does `$^` do? not sure how that works

mmussomele: The space tells `cut` to use spaces as delimiters when splitting. Basically the output of the shasum is `<hash> -` and I just want the hash, so `cut` extracts the first column, where the columns are delimited by spaces.

`$^` expands to all prerequisites for the make target.

Baisang: oh, I meant you should have a space after the `-d` flag, so it'd be `cut -d ' '`.

so `$(MAIN_SOURCE)` gets all of the relevant files and prints out their contents as well, then you take the hash of each file (?). so now you have hash of each file and filename, you sort by filename and then hash everything. makes sense

mmussomele: No particular reason for the flag, but it works either way.

As for your second point, the command that produces MAIN_SOURCE basically gets the path to every go source file in the tree that isn't a unit test, integration test, vendor file, or the version file itself. Those are then passed to shasum and then it does the rest like you said.

Baisang: what does `perl -ne 'print if /^.*(?<!test)\.go$$/'` do?

it seems like you're just getting the file names, and not actually getting the file contents? unless I am mistaken.

kklin:
I wonder if we could just use the git commit hash somehow. Seems much simpler, although it wouldn't account for as much as directly hashing the files.

cluster/foreman/foreman.go
```diff
@@ -213,9 +217,23 @@ func (c clientImpl) getMinion() (pb.MinionConfig, error) {
 func (c clientImpl) setMinion(cfg pb.MinionConfig) error {
 	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
 	_, err := c.SetMinionConfig(ctx, &cfg)
+
+	// gRPC mangles the error message, so we need to check the suffix, not the entire
+	// message.
+	if isVersionError(err) {
+		log.Warn(err)
```
kklin: I actually expect it to be common for the versions to differ if you're developing, so this seems like it'd get annoying very quickly. Not sure what would be better though.

Makefile
mmussomele: So `find`'s regex doesn't support negative lookbehinds, but I need that to only match go source files that don't end in `_test`, since tests are irrelevant to the behavior of quilt. The perl filters the go source files so that only non-tests are fed to `shasum`. And yup, just filenames, that's what `shasum` takes as arguments.

cluster/foreman/foreman.go
mmussomele: I thought about that, and that's why I decided to just have it warn (as opposed to error). Plus, it doesn't actually stop anything if the versions differ, quilt continues running as normal.

As for the git commit hash, I also considered that, but I figured a common situation that would cause the errors @luise encountered would happen while developing, and I wanted the warning to not depend on having remembered to commit before building.

mmussomele: Hmmm... maybe in order to ease development when you know your changes won't affect things, the daemon could take a `-ignore-versions` flag that suppressed this error?

Makefile
Baisang: don't we want to hash the file contents, not just the file names? otherwise if we change a file locally and `make install` then our local version will be different than whatever is on the minions, but the hash will be the same

Baisang: oops, for some reason I thought you were only hashing filenames for some reason. my bad!

Makefile
```diff
@@ -1,6 +1,7 @@
 export GO15VENDOREXPERIMENT=1
 PACKAGES=$(shell govendor list -no-status +local)
 NOVENDOR=$(shell find . -path -prune -o -path ./vendor -prune -o -name '*.go' -print)
+MAIN_SOURCE=$(shell find . -name "*.go" -not -path "./vendor/*" -not -path "./quilt-tester/*" -not -path "./minion/version/*" | perl -ne 'print if /^.*(?<!test)\.go$$/')
```
Baisang: how about `grep -v _test.go`? instead of the `perl` filtering command

Baisang: or `grep -v test`, etc.

mmussomele: Good point, would be simpler

Makefile
mmussomele: ah ok, yea definitely hashing the contents (and the filenames too I guess since there's a second call to `shasum`).

kklin:
I'm not sure how I feel about this patch overall. I have this gut feeling that it's too aggressive (I think the error will come up fairly regularly while developing, and I don't want us to get desensitized), but I'm not really sure.
I've requested that @ejj review it.

ejj:
So one thing, I'd say just use the `git` commit ID.  The sha1 is somewhat overkill.

I have some more thoughts, but switching to slack.

mmussomele:
Closing based on offline discussion.

kklin:
We can tweak and reopen this now that we have semantic versioning.

ejj:
Yeah, basically I think it comes down to saying that the "master" version can work with anything

kklin:
Why would we want that? I'd expect "master" daemon to just run with "master" minion.

ejj:
Ah, you were just worried about developers getting messed up.  But that's a good point, probably doesn't matter.

mmussomele:
Now in progress, not ready for merge.

mmussomele:
@ejj ready for another look

kklin:
@mmussomele Is this still ready for review?

ejj:
Closing for now.  @mmussomele please re-open when you're back.  Enjoy your summer!

PR #818, opened 2017-03-18
ci: Post test results in the new slack

kklin:
We're migrating to quilt-dev.slack.com.


kklin:
@luise friendly reminder that the quilt-bot deployment will also need to be updated to use this endpoint.

PR #826, opened 2017-03-24
quiltctl: Add help message for ssh commands that need to allocate pty

Baisang:
`quilt ssh` can fail with a tty error if a pty is not allocated for commands that need one.

Unfortunately, there is no good way to catch this command. Printing out the error returned from the native SSH client gives `Process exited with status 1` (lol)

So, this just adds a small help message in `quilt ssh -h` for users to be able to read.

### DIFF
--- a/quiltctl/command/ssh.go
+++ b/quiltctl/command/ssh.go
@@ -145,6 +145,8 @@ func (sCmd SSH) Run() int {
 		if exitErr, ok := err.(exitError); ok {
 			log.WithError(err).Debug(
 				"SSH command returned a nonzero exit code")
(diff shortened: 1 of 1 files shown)


kayousterhout:
Would it make sense for quilt ssh to always add the -t flag?

Baisang:
(for context, we discussed in person yesterday that aliasing `ssh` to `ssh -t` could cause some scripts to break if they check if there has been a pty allocated)

updated to now display an error message when ssh fails, will not always be applicable but can help catch common user mistakes

https://i.fluffy.cc/WNVTCMFd1VdT5Hs5qhGvH5N7qVqWnPcb.png

PR #827, opened 2017-03-24
amazon: Support reserved instances 

kklin:


cluster/amazon/amazon.go
```diff
@@ -29,15 +29,18 @@ type Cluster struct {
...
-// DefaultRegion is the preferred location for machines which haven't a user specified
-// region preference.
-const DefaultRegion = "us-west-1"
+const (
+	// DefaultRegion is the preferred location for machines which haven't a
+	// user specified region preference.
+	DefaultRegion = "us-west-1"
+
+	namespaceTagKey = "namespace"
+	spotPrice       = "0.5"
```
ejj: Btw, we should file a trello card.  The spot price should also be a machine attribute.  Doesn't need to happen int his PR though

kklin: Sounds good, is spot price an attribute that's common to all machines? Or would we need a way to specify provider-specific attributes?

cluster/google/google.go
```diff
@@ -148,6 +148,10 @@ func (clst *Cluster) Boot(bootSet []machine.Machine) error {
 	// XXX: should probably have a better clean up routine if an error is encountered
 	var names []string
 	for _, m := range bootSet {
+		if !m.Reserved {
+			return errors.New("ephemeral instances are not yet implemented")
```
ejj: I actually think the word "preemptable" is better.  That's what the google documentation calls them

cluster/amazon/amazon.go
```diff
@@ -104,91 +110,179 @@ func (clst *Cluster) Boot(bootSet []machine.Machine) error {
...
-					ImageId:          aws.String(amis[clst.region]),
-					InstanceType:     aws.String(br.size),
-					UserData:         &cloudConfig64,
-					SecurityGroupIds: []*string{aws.String(groupID)},
-					BlockDeviceMappings: []*ec2.BlockDeviceMapping{
-						blockDevice(br.diskSize),
-					},
-				},
-				InstanceCount: &count,
+		if br.reserved {
```
ejj: I might pull the bodies of the if and else statement into their own helper functions.  Not sure though, may not make it easier to read.

cluster/amazon/amazon.go
```diff
@@ -104,91 +110,179 @@ func (clst *Cluster) Boot(bootSet []machine.Machine) error {
...
+			return err
+		}
+
 	}
 
-	instIds := []string{}
-	for _, spot := range spots.SpotInstanceRequests {
-		if spot.InstanceId != nil {
-			instIds = append(instIds, *spot.InstanceId)
+	if spots != nil {
```
ejj: Can you pull this into the above if block?

cluster/amazon/amazon.go
```diff
@@ -104,91 +110,179 @@ func (clst *Cluster) Boot(bootSet []machine.Machine) error {
...
+			},
 		})
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	if err := clst.wait(ids, false); err != nil {
-		return err
+	for _, spot := range spotsResp.SpotInstanceRequests {
```
ejj: So I definitely had issues at one point with amazon losing tags in some edge cases.  Will this code be resilient to that?  I.E. if a spot request loses its tag, will we pick it up when the instance boots later?

stitch/README.md
```diff
@@ -80,3 +80,24 @@ can be reached.
...
+The Machine object represents a machine to be deployed.
+
+Its attributes are:
+- `role`: The Quilt role the machine will run as. *required*
+    - Master
+    - Worker
+- `provider`: The machine provider. *required*
+    - Amazon
+    - Google
+    - Vagrant
```
kklin: I'll remove it

stitch/README.md
```diff
@@ -80,3 +80,24 @@ can be reached.
 
 If multiple containers have the same hostname, an error is thrown during the
 vetting process.
+
+## Machine
+The Machine object represents a machine to be deployed.
+
+Its attributes are:
```
kklin: I think adding the types would be helpful, I'll do it.

cluster/amazon/amazon.go
kklin: Yeah, if we can match up an instance ID with a spot request, we still pick it up. We have a unit test that covers this https://github.com/kklin/quilt/blob/5437b3f44630295135a244b38c150f4520189d06/cluster/amazon/amazon_test.go#L100

PR #830, opened 2017-03-26
Digital Ocean!

ejj:
This patch implements a first pass at the Digital ocean cloud provider.


cluster/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,231 @@
...
+		}(m)
+	}
+
+	var err error
+	for range bootSet {
+		if e := <-errChan; e != nil {
+			err = e
+		}
+	}
+	return err
```
vivi: any reason why the last reported error is returned (instead of any other)?

vivi: any reason why the last reported error is returned (instead of any other)?

cluster/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,231 @@
...
+
+	_, _, err = clst.client.ListDroplets(&godo.ListOptions{})
+	return clst, err
+}
+
+// Creation is broken out for unit testing.
+var newDigitalOcean = func(namespace, region string) (*Cluster, error) {
+	// Underscores are illegal. Uppercase is illegal in volume names.
+	if strings.Contains(namespace, "-") || strings.Contains(namespace, "_") {
+		log.Warning("DigitalOcean namespaces cannot have underscores, hypens, " +
```
Baisang: It seems that hyphens should be allowed?

Also, I'm not sure if it's more intuitive to just log a warning here vs. forcing users to correct the namespace. I guess since they should be shown the changed namespace later it'll be okay either way.

kklin:
Please add digital ocean to `provider` in the docs (https://github.com/quilt/quilt/blob/master/stitch/README.md#machine)

cluster/digitalocean/digitalocean.go
ejj: Totally arbitrary.  I don't think it's ideal, but it's not clear what would be better either.

cluster/digitalocean/digitalocean.go
ejj: Ah yeah, that's left over kruft.

PR #831, opened 2017-03-27
quilt: Add version number

kklin:
This patch adds a version const, and a subcommand to print it.
Versioning will be useful for coordinating releases.


ejj:
FWIW, I think it's also important that the daemon and the minion log the version at boot.

kklin:
Oh, right, forgot about that. Done.

PR #833, opened 2017-03-28
amazon: Don't include pending instances in `List`

kklin:
Before, we would include pending reserved machines in `List`. This
resulted in a bug where the cluster would prematurely try to update
floating IPs on instances that had not yet fully booted.

We now block machine boot and stops until they are no longer pending.
If we time out before machines transition to the "running" state, we
destroy them to avoid leaking machines that might not ever properly
transition.


kklin:
@kayousterhout I updated the patch to try to cleanup machines that never properly transitioned from pending to running, can you take another look?

cluster/amazon/amazon.go
```diff
@@ -188,41 +207,58 @@ func (clst *Cluster) Stop(machines []machine.Machine) (err error) {
...
-		}
+		err = clst.stopSpots(spotIDs)
+	}
 
-		for _, spot := range spots.SpotInstanceRequests {
-			if spot.InstanceId != nil {
-				instIDs = append(instIDs, *spot.InstanceId)
-			}
-		}
+	if err == nil && len(instIDs) > 0 {
```
kayousterhout: Why do you need to check for no error here?  An error with stopping the spot instances seems like it could / should be unrelated to stopping the reserved instances

cluster/amazon/amazon.go
```diff
@@ -188,41 +207,58 @@ func (clst *Cluster) Stop(machines []machine.Machine) (err error) {
...
+	if len(instIDs) != 0 {
+		if err := clst.stopInstances(instIDs); err != nil {
 			return err
 		}
 	}
 
-	if len(instIDs) > 0 {
-		_, err = clst.client.TerminateInstances(&ec2.TerminateInstancesInput{
-			InstanceIds: aws.StringSlice(instIDs),
+	_, err = clst.client.CancelSpotInstanceRequests(
```
kayousterhout: I know this logic is mostly left from before -- but any reason not to try to cancel the spot instance requests, even if there was an error stopping the instances?  Seems like you could have an error cancelling the instances, but still have outstanding requests that can be cancelled

cluster/amazon/amazon.go
```diff
@@ -297,6 +333,11 @@ func (clst *Cluster) listInstances() (instances []awsMachine, err error) {
 				Name:   aws.String("instance.group-name"),
 				Values: []*string{aws.String(clst.namespace)},
 			},
+			{
+				Name: aws.String("instance-state-name"),
```
kayousterhout: what does this do?

cluster/amazon/amazon_test.go
```diff
@@ -538,6 +538,72 @@ func TestBoot(t *testing.T) {
 	mc.AssertExpectations(t)
 }
 
+func TestBootUnsuccessful(t *testing.T) {
```
kayousterhout: Can you add a comment about what this test is doing?  Something like

Boots both a preemptible and non-preemptible instance where <X bad thing happens so it's unsuccessful> and verifies that <Y>.

(what are X and Y?)

cluster/amazon/amazon_test.go
```diff
@@ -606,7 +672,13 @@ func TestStop(t *testing.T) {
...
 	mc.AssertCalled(t, "TerminateInstances",
 		&ec2.TerminateInstancesInput{
-			InstanceIds: aws.StringSlice([]string{reservedIDs[0], "inst1"}),
+			InstanceIds: aws.StringSlice([]string{"inst1"}),
+		},
+	)
+
+	mc.AssertCalled(t, "TerminateInstances",
+		&ec2.TerminateInstancesInput{
+			InstanceIds: aws.StringSlice([]string{reservedIDs[0]}),
```
kayousterhout: Why doesn't this also check that CancelSpotInstanceRequests was called for the 2nd spot instance that didn't get fully started?

kayousterhout:
Also can you update the commit + PR name + description?

cluster/amazon/amazon.go
kklin: It filters the returned machines to only have machines that are in the "running" state

cluster/amazon/amazon_test.go
kklin: I don't follow, we call `CancelSpotInstanceRequests` on all the spot instances IDs?

kklin:
@kayousterhout I updated the comment in the PR to match the commit message, if that's what you meant. Otherwise, I wasn't sure what else needs to be updated.

cluster/amazon/amazon.go
kklin: It just simplified the error handling a bit, but I've gone ahead and made it try both before returning.

cluster/amazon/amazon.go
kklin: It was to simplify the error handling, but I agree that we should try both. Fixed.

kklin:
@kayousterhout addressed your comments as best as I could

kayousterhout:
Re: update commit / PR description, I meant to mention the change to also cleanup / stop machines that never transitioned to running

kklin:
Oh, I had already done that. Did you want me to elaborate more?

cluster/amazon/amazon_test.go
kayousterhout: Sorry misread this -- thought that req. spot instances had been mocked to return 2 spot instance but I see it's just 1

cluster/amazon/amazon_test.go
```diff
@@ -538,6 +538,76 @@ func TestBoot(t *testing.T) {
...
+	)
+	mc.On("DescribeAddresses", mock.Anything).Return(
+		&ec2.DescribeAddressesOutput{}, nil,
+	)
+	mc.On("DescribeSpotInstanceRequests", mock.Anything).Return(
+		&ec2.DescribeSpotInstanceRequestsOutput{
+			SpotInstanceRequests: nil,
+		}, nil,
+	)
+	mc.On("TerminateInstances", &ec2.TerminateInstancesInput{
```
kayousterhout: Does this mock call also automatically verify that the function was actually called? (wondering how this test checks that the instances got stopped)

kayousterhout:
No new one looks good

cluster/amazon/amazon_test.go
kklin: Sigh, this is sloppy. I forgot the `mc.AssertExpectations(t)`, which ensures that all those `On` definitions are actually called. If you add it in, you'll see that the unit test actually fails . . .

kklin:
@ejj Could you have someone finish this up? I'm too rushed to finish this up properly.

There's two things that I think need to happen:
- @kayousterhout pointed out offline that rather than just returning one of the errors and potentially ignoring the other, we really should return an error that represents both of the errors. I do this in `clst.Stop` and `clst.stopSpots`, at a glance.
- The unit test should use `mc.AssertExpectations(t)` to assert that the required functions are actually called, and should be fixed (I would split it into two separate unit tests -- one just for spot machines, and one just for reserved machines).

ejj:
@kklin, I'll do it.

ejj:
Replaced with this PR: https://github.com/quilt/quilt/pull/843

PR #836, opened 2017-03-29
cluster: Log the reason a provider was unreachable

kayousterhout:
In the existing code, we log issues connecting to each provider
to debug (so most people won't see them) and then log a vague
message later if a provider the user tries to use couldn't be
setup correctly; e.g.:

WARNING [Mar 29 15:12:54.447] Provider Amazon is unavailable in us-west-1.

This commit saves the setup error so that we
can log it later on to help the user debug; e.g.:

WARNING [Mar 29 15:12:54.447] Provider Amazon is unavailable in us-west-1: AWS
failed to connect (No access credentials were found! Make sure your AWS
credentials are in ~/.aws/credentials.): NoCredentialProviders: no valid
providers in chain. Deprecated.
    For verbose messaging see aws.Config.CredentialsChainVerboseErrors


cluster/cluster.go
```diff
@@ -90,18 +98,13 @@ func newCluster(conn db.Conn, namespace string) *cluster {
...
 		conn:      conn,
-		providers: make(map[instance]provider),
+		providers: make(map[instance]maybeProvider),
 	}
 
 	for _, p := range allProviders {
 		for _, r := range validRegions(p) {
 			prvdr, err := newProvider(p, namespace, r)
-			if err != nil {
-				log.Debugf("Failed to connect to provider %s in %s: %s",
```
kayousterhout: I'm not totally sure this is the right thing to do.  I don't think this debug message is useful for providers you're not using, and if you are using the provider, the message will get logged at a higher level elsewhere -- but wanted to flag it in case I'm missing something or others think it is useful.

cluster/cluster.go
```diff
@@ -332,17 +336,21 @@ func (clst cluster) syncACLs(adminACLs []string, appACLs []db.PortRange,
...
 
-	for inst, prvdr := range clst.providers {
-		// For providers with no specified machines, we remove all ACLs.
-		// Otherwise we set acls to what's specified.
-		var setACLs []acl.ACL
-		if _, ok := prvdrSet[inst]; ok {
-			setACLs = acls
-		}
+	for inst, maybePrvdr := range clst.providers {
+		if maybePrvdr.setupErr == nil {
```
ejj: Would be more go-ish to do a `!= nil` and a `continue`.

cluster/cluster.go
```diff
@@ -433,12 +441,14 @@ func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
 
 func (clst cluster) get() ([]machine.Machine, error) {
 	var cloudMachines []machine.Machine
-	for _, p := range clst.providers {
-		providerMachines, err := p.List()
-		if err != nil {
-			return []machine.Machine{}, err
+	for _, maybePrvdr := range clst.providers {
+		if maybePrvdr.setupErr == nil {
```
ejj: Same here.

ejj:
Looking through this, I suppose I have a nagging feeling that a simpler solution exists.  Unfortunately, it's not entirely clear to me what it is.

Here's one idea.  What if, instead of initializing the providers when the cluster is created.  We instead do so lazily when they're needed?  So basically the updateCloud function would initialize the providers that haven't been initialized yet, and are actually needed by the spec.  Would that be cleaner/simpler?

kayousterhout:
We'd discussed your idea last week but thought it was too much work, but turns out it was super simple.  There's one big downside though, which is that is you start a new quilt daemon and run an empty spec (to shutdown existing machines), the provider won't be setup yet, so we won't do anything.  This still seems probably better than the current approach, and at this point it also seems straightforward to add a config for "providers we care about" as we've discussed in the past. 

Let me know what you think and then I can fix up the tests accordingly.

kayousterhout:
Updated.  I can squash the commits once we've decided on the approach.

kklin:
Personally, I think being unable to stop machines after restarting the daemon is pretty unacceptable.

ejj:
FWIW, I like the new version a lot better.  I think you should go ahead and squash.

I'd say high level notes, may be good to make a "getProvider" function that returns or initializes.  Also may be nice to unit test if it's not too much trouble.

ejj:
Replaced with this PR: https://github.com/quilt/quilt/pull/846

PR #837, opened 2017-03-29
build: Build windows executable

ejj:
We hope to allow quilt to be used from windows desktops.  In service
of that goal, this patch supports cross-compiling the binary for
windows.  The resulting binary hasn't been tested, and may not work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,9 @@ clean:
 	rm -f *.cov.coverprofile cluster/*.cov.coverprofile minion/*.cov.coverprofile
 	rm -f *.cov.html cluster/*.cov.html minion/*.cov.html
 
+windows:
+	cd -P . && GOOS=windows GOARCH=386 go build .
+
 COV_SKIP= /api/client/mocks \
 	  /api/pb \
 	  /cluster/provider/mocks \
--- a/quiltctl/quiltctl.go
+++ b/quiltctl/quiltctl.go
@@ -9,12 +9,12 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
+// Note the `minion` command is in quiltclt_posix.go as it only runs on posix systems.
 var commands = map[string]command.SubCommand{
 	"daemon":  command.NewDaemonCommand(),
 	"get":     &command.Get{},
 	"inspect": &command.Inspect{},
 	"logs":    command.NewLogCommand(),
-	"minion":  command.NewMinionCommand(),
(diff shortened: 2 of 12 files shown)


ejj:
To be clear, this should be considered an RFC ... need to clean up something

ejj:
New version is better.  Ready for review.

quiltctl/quiltctl.go
```diff
@@ -9,15 +9,13 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
+// Note additional commands are in posix_cmds.go.
```
mmussomele: I think this comment should elaborate on why the commands are separate slightly more.

Makefile
```diff
@@ -32,6 +32,9 @@ clean:
 	rm -f *.cov.coverprofile cluster/*.cov.coverprofile minion/*.cov.coverprofile
 	rm -f *.cov.html cluster/*.cov.html minion/*.cov.html
 
+windows:
+	cd -P . && GOOS=windows GOARCH=386 go build .
```
kklin: Do we need multiple binaries to support 32 vs 64 bit?

quiltctl/ssh/resize_windows.go
```diff
@@ -0,0 +1,7 @@
+package ssh
+
+import "os"
+
+func signalResize(sig chan os.Signal) {
+	// Unimplemented
```
kklin: Is it worth logging here?

quiltctl/ssh/native.go
```diff
@@ -170,7 +169,7 @@ func (p *pty) Request() error {
 		return err
 	}
 
-	signal.Notify(p.resizeSignal, syscall.SIGWINCH)
+	signalResize(p.resizeSignal)
```
kklin: This function name makes me think that we're sending a resize signal at this line. Maybe something like `setupResizeHandler` would be more appropriate.

Makefile
ejj: Nah, 32 bit binaries run on 64bit systems, and we aren't performance sensitive or using all that much memory, so there's no point.

quiltctl/ssh/resize_windows.go
ejj: I think it will just be confusing personally.

kklin: I'm fine without it.

PR #838, opened 2017-03-30
cluster: Fix join shuffling and FloatingIP discrepancies

Baisang:
Before, it was possible for `machine.Machine`s to flip-flop between `db.Machine`s
in the cluster join. The join has been updated to now break ties between machines
by prioritizing joining machines that have the same PublicIP and other attributes.

Additionally, assigning FloatingIPs was updated. FloatingIPs are now only assigned
once the CloudID of a machine has been assigned. This ensures that FloatingIPs will
only be assigned once the machine has fully booted and been stabilized in the join
(meaning, it will not join with any other machine).

Tests have been corrected and added to reflect these changes.

==============================================================

Seemed fine locally with the logs test, but I haven't really been able to test this with floating IP outside of unit tests, so let me know if anything seems a little shaky.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -253,6 +253,7 @@ func (clst cluster) join() (joinResult, error) {
 		}
 
 		res.machines = view.SelectFromMachine(nil)
+		cloudMachines = foreman.GetMachineRoles(cloudMachines)
 
 		dbResult := syncDB(cloudMachines, res.machines)
 		res.boot = dbResult.boot
@@ -356,15 +357,13 @@ type syncDBResult struct {
 
 func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
 	ret := syncDBResult{}
-	cms = foreman.GetMachineRoles(cms)
 
 	pair1, dbmis, cmis := join.Join(dbms, cms, func(l, r interface{}) int {
 		dbm := l.(db.Machine)
 		m := r.(machine.Machine)
 
-		// Don't join if m.ID and dbm.CloudID are empty.
-		if m.ID != "" && dbm.CloudID == m.ID &&
-			dbm.Provider == m.Provider && dbm.Preemptible == m.Preemptible &&
+		if dbm.CloudID == m.ID && dbm.Provider == m.Provider &&
+			dbm.Preemptible == m.Preemptible &&
 			dbm.Region == m.Region && dbm.Size == m.Size &&
 			(m.DiskSize == 0 || dbm.DiskSize == m.DiskSize) &&
 			(m.Role == db.None || dbm.Role == m.Role) {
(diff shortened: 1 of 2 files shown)


Baisang:
lol forgot to lint, one sec

cluster/cluster.go
```diff
@@ -356,13 +357,12 @@ type syncDBResult struct {
...
 func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
 	ret := syncDBResult{}
-	cms = foreman.GetMachineRoles(cms)
 
 	pair1, dbmis, cmis := join.Join(dbms, cms, func(l, r interface{}) int {
 		dbm := l.(db.Machine)
 		m := r.(machine.Machine)
 
-		// Don't join if m.ID and dbm.CloudID are empty.
+		// Don't join if m.ID is empty (unit tests)
```
ejj: Why this (unit tests) note?

cluster/cluster.go
```diff
@@ -378,25 +378,29 @@ func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
...
 			dbm.Region != m.Region ||
 			dbm.Size != m.Size ||
 			dbm.Preemptible != m.Preemptible ||
 			(m.DiskSize != 0 && dbm.DiskSize != m.DiskSize) ||
-			(m.Role != db.None && dbm.Role != m.Role):
+			(m.Role != db.None && dbm.Role != m.Role) {
 			return -1
-		case m.ID != "" && dbm.CloudID == m.ID:
+		}
+		if m.ID != "" && dbm.CloudID == m.ID {
```
ejj: Why do we need this case at all then?

cluster/cluster.go
```diff
@@ -378,25 +378,29 @@ func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
...
 		}
+
+		score := 10
+		if dbm.Role != db.None && m.Role != db.None && dbm.Role == m.Role {
+			score -= 4
+		}
+		if dbm.FloatingIP == m.FloatingIP {
+			score -= 2
+		}
+		if dbm.PublicIP == m.PublicIP && dbm.PrivateIP == m.PrivateIP {
```
ejj: Personally, I think that the IP being correct should be a bigger weight than the floating IP being correct.  That way you'd be preferring fundamental machine attributes over things that can be moved to that machine

ejj:
LGTM, just a couple of little nits

cluster/cluster.go
kklin: +1

cluster/cluster.go
Baisang: `m.ID` shouldn't really be empty in real life, it only is for some unit tests. The alternative to this is correcting all of the unit tests that lack `m.ID` currently, I'm not sure what is best

cluster/cluster.go
Baisang: I left it in from previous, we can remove it if you think not necessary

cluster/cluster.go
Baisang: Makes sense, I'll try it out

cluster/cluster.go
ejj: Doesn't look necessary to me, but I'd like you to convince yourself either way =)

Baisang:
@ejj ready for review again

PR #840, opened 2017-04-02
amazon: Continue blocking boot/stop when List fails

kklin:
Before, we exited cleanly from `wait` if the List failed. However, a
failure to list machines is most likely an indication of an error -- it
should not be treated as equivalent to the machines being in the desired
state.


kklin:
@kayousterhout added a commit to change the log message, did you want to take another look?

PR #843, opened 2017-04-04
amazon: Don't include pending instances in `List`

ejj:
Before, we would include pending reserved machines in `List`. This
resulted in a bug where the cluster would prematurely try to update
floating IPs on instances that had not yet fully booted.

We now block machine boot and stops until they are no longer pending.
If we time out before machines transition to the "running" state, we
destroy them to avoid leaking machines that might not ever properly
transition.

### DIFF
--- a/cluster/amazon/amazon.go
+++ b/cluster/amazon/amazon.go
@@ -126,25 +126,22 @@ func (clst *Cluster) Boot(bootSet []machine.Machine) error {
 		bootReqMap[br] = bootReqMap[br] + 1
 	}
 
-	var ids []string
 	for br, count := range bootReqMap {
-		var newIDs []string
 		if br.preemptible {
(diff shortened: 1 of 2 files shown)


ejj:
This has already been reviewed.  I'm just going to merge it once travis passes.

ejj:
Also, see the original PR: https://github.com/quilt/quilt/pull/833

kayousterhout:
Did you see Kevin's comment about the tests? 

"The unit test should use `mc.AssertExpectations(t)` to assert that the required functions are actually called, and should be fixed (I would split it into two separate unit tests -- one just for spot machines, and one just for reserved machines)."

(the first part seems most important to fix)

ejj:
Newest version added AssertExpectations to the new test.

PR #845, opened 2017-04-06
stitch: Move to Node.js (RFC)

PenguinToast:
This patch replaces Otto with Node.js for running Quilt specs. We now
require Node.js to be installed and on your PATH, and you need to run
`npm install` first to install all the Node.js dependencies. All Quilt
specs will also need to be modified to use Node.js. Then, `quilt run`
can be used as before.

I'm creating this PR simply to get comments on the general idea around
my implementation; no unit tests have been modified in this PR thus
running `make check` will fail. quilt/nginx#2 accompanies this PR so
 that we can see how the new specs will look.


PenguinToast:
@kklin

kklin:
I'll open a new PR once it's ready

PR #846, opened 2017-04-06
cluster: Set up providers on demand

ejj:
In addition to setting  up all of the providers for all of the regions
when the Quilt daemon is started, this commit additionally attempts to
setup a provider when it's needed by a spec.  This makes error
handling easier, and also means that Quilt will attempt to fix setup
errors by re-connecting to the provider in every iteration of the loop
(whereas currently, if Quilt fails to connect to a provider, it will
never re-try).

Co-Authored-By: Ethan J. Jackson <ejj@eecs.berkeley.edu>


ejj:
Newer version of this PR: https://github.com/quilt/quilt/pull/836

PR #847, opened 2017-04-07
build: Simple debian packaging for quilt

Baisang:
This patch adds functionality to create debian packaging for quilt
(i.e. a .deb file that can be installed easily by users). Output is
generated within the dist/ directory.

It does not completely comply with debian packaging policy as we
use dependencies that are not in debian yet, and the package is
currently not run through lintian. It should be sufficient for
test distribution for the time being.

Also, we might want to just put this in some sort of `build` branch or something, since the `debian` folder can be extra clutter. I don't think this necessarily belongs in `master`.


debian/control
```diff
@@ -0,0 +1,44 @@
...
+ .
+ The Quilt.js JavaScript framework allows for development, versioning, and
+ testing of infrastructure in the same way we do for application code.
+ Additionally, Quilt.js code is shareable, reusable and composable,
+ making it easy to set up and manage systems without being an expert in
+ system administration.
+ .
+ Quilt is a research project out of UC Berkeley. It is currently under
+ heavy development, but please try it out - we are eager for feedback!
+ Example: Deploying a MEAN Stack App with Quilt The MEAN stack (MongoDB,
```
luise: We should delete `Example:` and everything after that.

debian/control
```diff
@@ -0,0 +1,44 @@
+Source: quilt
+Section: admin
+Priority: extra
+Maintainer: Ethan Jackson <ejj@eecs.berkeley.edu>
```
luise: Ethan J. Jackson

debian/control
```diff
@@ -0,0 +1,44 @@
...
+  debhelper (>= 9),
+  dh-golang,
+  golang-any
+Standards-Version: 3.9.8
+Homepage: https://github.com/quilt/quilt
+Vcs-Browser: https://github.com/quilt/quilt.git
+Vcs-Git: https://github.com/quilt/quilt.git
+XS-Go-Import-Path: github.com/quilt/quilt
+
+Package: golang-github-quilt
```
luise: I assume `quilt` is  taken? If so, what about `quilt.js` or `quilt-js`? Or do you have a good reason for the current name? I think JS is more descriptive of us than Go or GitHub.

debian/rules
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/make -f
+
+export DH_VERBOSE=0
+
+export DH_GOPKG := github.com/quilt/quilt
```
luise: Why `:=`?

debian/rules
```diff
@@ -0,0 +1,34 @@
...
+	rm -rf $(GOPATH)/quilt-build-deps*.deb
+	rm -rf $(GOPATH)/bin/* $(GOPATH)/pkg/*
+	rm -f $(GOPATH)/goinstall.log
+
+binary-arch: clean
+	dh_prep
+	dh_installdirs
+	cd $(GOPATH)/src/$(DH_GOPKG) && go install
+	mkdir -p $(PKGDIR)/usr/bin
+	cp $(GOPATH)/bin/* $(PKGDIR)/usr/bin/
```
luise: Could `$(GOPATH)/bin/*` contain stuff that isn't related? If so, could we use `$(GOPATH)/bin/quilt/*` or something instead?

Makefile
```diff
@@ -125,5 +125,17 @@ docker-push-quilt:
 docker-build-ovs:
 	cd -P ovs && docker build -t ${REPO}/ovs .
 
+builddeb-docker-image:
+	${DOCKER} build -t quilt-build -f Build_Dockerfile .
+
+builddeb-docker: builddeb-docker-image
+	mkdir -p dist
```
luise: Isn't `-p` is redundant since `dist` is just a single directory?

Makefile
```diff
@@ -125,5 +125,17 @@ docker-push-quilt:
...
+	${DOCKER} run -v $(PWD):/mnt/src/github.com/quilt/quilt quilt-build
+
+builddeb:
+	dpkg-buildpackage -us -uc -b
+	mkdir -p dist
```
luise: Same with `-p` here?

luise:
@Baisang  I have a few questions and comments, but I don't need to look at it again. @ejj or @kklin might have more of an opinion on the setup, directory structure, and details of the packaging...

debian/control
Baisang: standard name for go packages in debian is that format, we could probably try to do `quilt-js`. the name isn't too big of an issue right now since this won't be "in debian" for a long time

debian/rules
Baisang: let me investigate, this `debian/rules` was adapted off of another debian go project

debian/control
Baisang: oh but yes `quilt` by itself is taken, that's the name of a tool in debian

Baisang: ah I am wrong about this, accidentally was looking at the section for libraries instead of binary-only packages. we can do `quilt-js`

http://pkg-go.alioth.debian.org/packaging.html

Makefile
Baisang: if it already exists, -p won't throw an error

kklin:
We're going to close this for now because we want to distribute binaries through `npm` at first. Debian packaging doesn't seem worth the overhead right now, especially considering the effort needed to get it merged upstream.

PR #853, opened 2017-04-10
docs: Added developer guide to Quilt

kayousterhout:
The new guide describes how to define new containers and
deploy them with Quilt.

This is a first cut -- I'm sure we'll need to refine this more, but
I wanted to check in what I have as @luise thinks more about
the README / etc.


stitch/developers-guide.md
```diff
@@ -0,0 +1,131 @@
+# Developers Guide to Quilt
```
ejj: One high level comment, and I'm curious what others thing, but when I here "developers guide" I think of someone developing the actual quilt system rather than specs.  Wonder if we should call this a "spec writers" guide or a "Quilt.js developers" guide.

Also, I think we should go ahead and link to this in the README so it's easier to find.

stitch/developers-guide.md
```diff
@@ -0,0 +1,131 @@
...
+
+This guide describes how to write the Quilt spec for a new application, using the lobste.rs application as an example.  lobste.rs is an open source project that implements a reddit-like web page, where users can post content and vote up or down other content.
```
ejj: Ah also, please word wrap at 80

stitch/developers-guide.md
kayousterhout: Good point -- done.  Let me know what you think now.

kayousterhout:
@ejj just want to make sure this hadn't been forgotten (Will asked about something like this guide on Slack so would be nice to get it up)

PR #854, opened 2017-04-10
Fix whitespace in README

timanglade:
Fenced code blocks don’t need to be indented.


ejj:
Thanks Tim!  LGTM, will merge after travis is happy.

PR #855, opened 2017-04-11
build: Add explicit targets for multiple platforms

ejj:
The upcoming release will publish binaries for windows, mac and linux.
This patch makes sure build targets are implemented for all three.


ejj:
Not sure this needs a review.  Just going to merge if the build passes.

PR #858, opened 2017-04-12
stitch: Add dirExists() for checking existence of a directory

luise:
This commit adds the function dirExists() that checks if a directory exists.
This is useful for specs that use readDir(), since importing such specs will
throw an error if the path given to `readDir` doesn't exist.


stitch/stitch.go
```diff
@@ -319,6 +322,20 @@ func readDirImpl(call otto.FunctionCall) (otto.Value, error) {
 	return call.Otto.ToValue(filesJS)
 }
 
+func dirExistsImpl(call otto.FunctionCall) (otto.Value, error) {
+	path, err := getPath(call)
+	if err != nil {
+		return otto.Value{}, err
+	}
+
+	exists, err := afero.Afero{Fs: util.AppFs}.DirExists(path)
```
mmussomele: One small nit, it feels weird that we wrap `util.AppFs` here instead of in a function in the `util` package, but given that the same thing is done above, not a big deal.

luise: The right thing would have been to use `util.FileExists` instead... I didn't realize that function existed.

luise: The right thing would have been to use `util.FileExists` instead... I didn't realize that function existed.

luise: The right thing would have been to use `util.FileExists` instead... I didn't realize that function existed.

luise: Sorry, confused this with our Slack discussion. I think this is fine - as you say, the same thing is done right above.

PR #862, opened 2017-04-24
amazon: Stop tagging spot requests

ejj:
This patch reimplements commit 46b2cc4 which was reverted due to bugs.
The original commit message below:

When the amazon provider was first written, there was no way to filter
spot requests based on the security group they were in.  So, instead,
it tagged spot requests with their namespace in lieu of checking the
security group directly.

Not only did this approach complicate the code, it also lead to a
brief window of time where a spot request was created, but not yet
tagged.  During this window all quilt daemons running on a particular
aws account would consider those spot requests as superfluous members
of in their own namespace, and would thus cancel them.  This patch
does away with this issue, and simplifies the code, by simply
filtering spot requests on their security group when listed.


ejj:
fixed

cluster/amazon/amazon_test.go
```diff
@@ -87,14 +87,8 @@ func TestList(t *testing.T) {
 				// A spot request with tags and a corresponding instance.
```
kayousterhout: Can you update the comments here too?

ejj:
done

cluster/amazon/amazon.go
```diff
@@ -293,33 +287,21 @@ func (clst *Cluster) stopInstances(ids []string) error {
...
-					Values: trackedSpotStates,
-				},
-			},
-		})
+func (clst *Cluster) listSpots() (machines []awsMachine, err error) {
+	input := ec2.DescribeSpotInstanceRequestsInput{Filters: []*ec2.Filter{{
+		Name:   aws.String("state"),
+		Values: trackedSpotStates,
+	}, {
+		Name:   aws.String("launch.group-name"),
```
Baisang: Does this filter work? I can't seem to find it anywhere in their docs, so I'm confused.

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSpotInstanceRequests.html

(they don't have an anchor for this link, but ctrl-f "DescribeSpotInstanceRequest")
https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/

ejj: Ah I'm glad you ask!

So the filter is not documented, but I tested it and it works.  `instance.group-name` is a documented filter used in `describe-instances`. For the spot requests they seem to have documented `launch.group-id` as meaning the ID of the security group for the booted instance analogous to `instance.group-id` when doing describe instances.  Thus, I just guessed that `launch.group-name` would work and it appears to.

Baisang: I see. I would normally be worried about using an undocumented filter, but we tried using a documented one and it failed anyway ¯\_(ツ)_/¯

It's strange they have that and it's not documented, but as long as it works for us that sounds good!

PR #865, opened 2017-04-30
Ovs upgrade

Baisang:
Upgrade OVS to 2.7 and make `ovn-controller` not restart every time

Both commits were tested using quilt-tester locally once, all tests passed

### DIFF
--- a/minion/supervisor/worker.go
+++ b/minion/supervisor/worker.go
@@ -95,9 +95,6 @@ func runWorkerOnce() {
 		return
 	}
 
-	/* The ovn controller doesn't support reconfiguring ovn-remote mid-run.
-	 * So, we need to restart the container when the leader changes. */
-	Remove(images.Ovncontroller)
 	run(images.Ovncontroller, "ovn-controller")
 }
 
--- a/ovs/Dockerfile
+++ b/ovs/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:16.04
 
-ENV ovs_ver="2.6.1"
+ENV ovs_ver="2.7.0"
 ENV kernel_versions="4.4.0-62-generic 4.4.0-77-generic"
 
 RUN build_deps="build-essential libssl-dev python python-six wget" \


kklin:
Did you test changing ovn-remote?

Baisang:
@kklin not explicitly, what would be a good way to do that?

kklin:
Not sure, I'm not sure there's a _good_ way. Ideally, you'd trigger a leader change by booting two masters, and killing the leader or something. But I bet other code around leader failover is broken.

(BTW, I'm fine merging it for now if we can't figure out a way to test it. Fixing leader failover is an entire project in itself)

minion/supervisor/worker.go
```diff
@@ -69,7 +69,6 @@ func runWorkerOnce() {
 	}
 
 	oldEtcdIPs = etcdIPs
-	oldIP = IP
```
mmussomele: Since machines don't change roles anymore, isn't this assignment ineffectual now? A bit of a nit, but it feels like it deserves it's own patch

Baisang: This was a line that was ineffectual I put in from a previous patch, was just going to trim it with this one. We still need some form of 'state variables' in supervisor currently as it doesn't sync with the state of the real world yet.

Baisang:
@kklin , I tested booting with 2 masters and killed a master on aws. quilt then tried to boot another master, an election was had, and then it removed and replaced/restarted all containers. Everything seemed to work after that.

Let me know if you want the log outputs or anything

kklin:
@Baisang nah, that sounds fine to me

kklin:
Actually, shouldn't ovn-controller never restart on workers now?

Baisang:
@kklin I believe so, unless the container dies for some reason

kklin:
Oh, when you said "replaced/restarted all containers" do you mean the application containers? So not ovn-controller?

Baisang:
yes, i meant the application containers. I didn't check if ovn-controller restarted, I can do that if you'd like

minion/supervisor/worker.go
kklin: I think you should put it in its own patch. Then in the commit description you could describe why the line is ineffectual, and thus can be removed. It's helpful to do it in a separate commit because if it ends up that the line _is_ necessary, it'll be easier to pinpoint the offending change.

kklin:
I think it's alright, we'll have to do more thorough failover testing soon anyways.

Baisang:
@kklin moved the ineffectual assignment to another patch

PR #868, opened 2017-05-04
Support arbitrary project IDs for GCE

kklin:


cluster/google/google.go
```diff
@@ -566,16 +566,13 @@ func (clst *Cluster) firewallPatch(name string,
 	return clst.gce.PatchFirewall(clst.projID, name, firewall)
 }
 
-func newComputeService(ctx context.Context) (*compute.Service, error) {
-	client, err := google.DefaultClient(ctx, compute.ComputeScope)
+func newComputeService(keyfile string) (*compute.Service, error) {
```
mmussomele: Just a nit, keyfile still feels like the name of a variable holding a path to me.

mmussomele: Otherwise LGTM

PR #871, opened 2017-05-11
openflow: Only replace flows if necessary

kklin:
The `ovs-ofctl replace-flows` subcommand is buggy because it treats some
flows as different even though they are actually equivalent. This
results in certain flows being replaced every time `replace-flows` is
called, and might have caused the network failures seen in integration
tests.

To work around this before the problem is fixed upstream, we only run
`replace-flows` if `diff-flows` shows any changes.


minion/network/openflow/openflow.go
```diff
@@ -242,7 +249,7 @@ func openflowPorts() (map[string]int, error) {
 
 var ofctl = func(action string, flows []string) error {
 	cmd := exec.Command("ovs-ofctl", "-O", "OpenFlow13", action,
-		ipdef.QuiltBridge, "-")
+		ipdef.QuiltBridge, "/dev/stdin")
```
kayousterhout: why this change?

minion/network/openflow/openflow_test.go
```diff
@@ -20,11 +20,13 @@ func TestAddReplaceFlows(t *testing.T) {
 		return client, nil
 	}
 
-	var action string
-	var flows []string
+	actions := map[string][]string{}
```
kayousterhout: can you give this a better name like actionToFlows or similar?

kklin: Done

minion/network/openflow/openflow.go
kklin: replace-flows doesn't support using `-` to specify stdin.

PR #872, opened 2017-05-11
Fixes for debug-logs

kklin:


quiltctl/command/debug.go
```diff
@@ -59,35 +59,45 @@ type logCmd struct {
...
-}
+	// A list of commands to output various machine logs.
+	machineCmds = []logCmd{
+		{"docker-ps", "docker ps -a"},
+		{"minion", "docker logs minion"},
+		{images.Etcd, "docker logs " + images.Etcd},
+		{images.Ovsdb, "docker logs " + images.Ovsdb},
+		{"syslog", "sudo cat /var/log/syslog"},
+		{"journalctl", "sudo journalctl -xe"},
+		{"uname", "uname"},
```
Baisang: why not `uname -a`? unless this is already outputting everything we want

kklin: I think `uname -a` would be useful, I just tacked on another commit as this one was just shuffling around the structure.

PR #873, opened 2017-05-12
Build OVS kernel modules at runtime

kklin:


kklin:
This feature is kind of odd in that it's not a part of the standard code path, but if it doesn't work, then the project will look really broken (specifically, when DigitalOcean updates their Ubuntu image, the network will never come up).

I have a couple things I'd like feedback on:

1) I wasn't sure how best to inform the user that the kernel module is being built at runtime. I think the log message I have is easily missed.

2) How will we know when we will have to update the kernel image bundled in `quilt/ovs` though? Will it just have to wait until we have integration tests for all the providers, and then we'd grep the log output?

3) I don't think there's a good way to test this on a regular basis..

kklin:
Also, note to self: after merging, if `quilt/quilt` gets built before `quilt/ovs`, `/bin/bootstrap` won't be present in the `quilt/quilt` image. Not sure if we have this problem in our build setup.

ovs/bootstrap
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+set -e
+
+build_dir="/build"
+ovs_ver="2.6.1"
```
Baisang: should be 2.7.0 after the upgrade patch

kklin: Fixed

kklin:
@Baisang did you have any other comments?

@ejj @kayousterhout any thoughts on the 3 questions I posted in the above comment?

ejj:
Kevin -- RE your above comments:

1) I actually think just a log message is fine.  It's not something actionable by the user.  Things are just slower and there's nothing they can do about it.

2) Yep I think an integration test is a good idea.  As mentioned offline yesterday, I think we should also just figure out Ubuntu's release schedule for new kernel modules, and proactively update the quilt/ovs image.  Ideally we could just make a calendar event that reminds us monthly.

3) Well, with (2), when a new release goes out, we could probably test it locally with vagrant or something before upgrading to the new image.  Or alternatively could try out a new minion or something.  But yeah, in general, features like this are somewhat painful as they get stale.

kayousterhout:
This looks good to me and I agree that we should have so kind of integration test, although I don't understand enough about this to know what that should be.

ovs/bootstrap
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+set -e
+
+build_dir="/build"
+ovs_ver="2.7.0"
+build_deps="build-essential libssl-dev python python-six curl"
+
+function setup_build() {
+    apt-get update && apt-get install -y ${build_deps}
```
Baisang: If `apt-get update` fails, will it fail silently?

kklin: No, if anything exits non-zero, the entire script exits (because of the `set -e`). The way our systemd service is setup, if the OVS service fails, the minion doesn't try to start. Logs are accessible through journalctl.

PR #879, opened 2017-05-18
Migrate to Node.js

kklin:


kklin:
The spec test is expected to fail because the PRs in the other repos have not yet been merged.

One person should review all the PRs.

minion/engine.go
```diff
@@ -19,12 +19,42 @@ func updatePolicy(view db.Database, spec string) {
...
+// listening on the same public port get placed on the same machine.
+func portPlacements(connections []db.Connection) (placements []db.Placement) {
+	ports := make(map[int][]string)
+	for _, c := range connections {
+		if c.From != stitch.PublicInternetLabel {
+			continue
+		}
+
+		min := c.MinPort
+		ports[min] = append(ports[min], c.To)
```
ejj: Btw, I think this code is a bit misleading ... Do public connections support port ranges?  If they do, then the code is wrong because using the min alone won't detect overlaps.  If they don't, then the code deserves a comment, as it's confusing.

ejj:
`quilt get` appears to still be mentioned in the `quilt` usage string.

.travis.yml
```diff
@@ -4,15 +4,19 @@ sudo: required
 
 services:
     - docker
+
 go:
     - 1.x
 
+before_install:
+    - nvm install 7.10.0
```
ejj: I don't think this is the recommended way to get node installed for travis.  Seems like they have a node_js target built in: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/

Makefile
```diff
@@ -58,7 +63,7 @@ COV_SKIP= /api/client/mocks \
 	  /version
 
 COV_PKG = $(subst github.com/quilt/quilt,,$(PACKAGES))
-coverage: $(addsuffix .cov, $(filter-out $(COV_SKIP), $(COV_PKG)))
```
ejj: Why not leave this target, and have it depend on `go-coverage` and `js-coverage`?  Seems more intuitive to me, and we don't have to change the travis target.

Makefile
```diff
@@ -68,6 +73,9 @@ coverage: $(addsuffix .cov, $(filter-out $(COV_SKIP), $(COV_PKG)))
 	go test -coverprofile=.$@ .$*
 	go tool cover -html=.$@ -o .$@.html
 
+js-coverage:
+	npm run-script cov > bindings.lcov
```
ejj: Out of curiosity, does codecov.io understand this?

docs/GettingStarted.md
```diff
@@ -2,6 +2,12 @@
 This guide explains how to install Quilt, and also serves as a
 brief, hands-on introduction to some Quilt basics.
 
+## Install Node
+Quilt specs are written in Node. Only Node version v7.10.0 has been tested.
```
ejj: I'd be a little bit more forma, and say `Node.js`, context may not be clear to a new user.

docs/GettingStarted.md
```diff
@@ -60,13 +55,19 @@ We suggest you read
...
 
-### Import Nginx Spec
-Before we can run the Nginx spec, we need to import it into our `QUILT_PATH`.
-To do that, execute the following command from your shell:
+### Acquire the Nginx Spec
+In order to run the Nginx spec, we'll have to download it first. We'll simply
+clone it:
 ```bash
-quilt get github.com/quilt/nginx
+git clone https://github.com/quilt/nginx
```
ejj: Is it possible to do this with npm directly somehow?  Seems like you can point npm at a URL for example: https://docs.npmjs.com/cli/install.

Not sure.  What's the most canonical option here?

docs/GettingStarted.md
```diff
@@ -60,13 +55,19 @@ We suggest you read
...
+cd nginx
 ```
 
+### Install Spec Dependencies
+The Nginx spec depends on the `@quilt/quilt` module. More complicated specs
```
ejj: This may be a dumb question form a node newbie, but is it possible to get the specs to just depend on `@quilt` instead of `@quilt/quilt`?

quiltctl/command/run_test.go
```diff
@@ -301,6 +181,7 @@ func TestConfirm(t *testing.T) {
 }
 
 func TestPromptsUser(t *testing.T) {
+	t.Skip("remove node dep")
```
ejj: What's this for?  I'd prefer to just remove a test rather than skip it unless their is some strange circumstance.

stitch/bindings.js
```diff
@@ -1,5 +1,29 @@
+const crypto = require('crypto');
```
ejj: Should this have use strict?  Or is this prepended to the specs automagically?

.travis.yml
tejasmanohar: used that before, just runs `nvm` under the hood but is cleaner

README.md
```diff
@@ -37,9 +37,10 @@ web proxy:
 
 [//]: # (b1)
 ```javascript
-var Node = require("github.com/quilt/nodejs");
-var HaProxy = require("github.com/quilt/haproxy").Haproxy;
-var Mongo = require("github.com/quilt/mongo");
+const {Machine, createDeployment, Range, githubKeys} = require("@quilt/quilt");
```
tejasmanohar: kinda weird to mix var and const when everything's constant

const doesn't add much value in JS IMO, just indicates that the reference won't change... doesn't even concern value, but it's trendy now so maybe just use it across the board

docs/GettingStarted.md
```diff
@@ -2,6 +2,12 @@
 This guide explains how to install Quilt, and also serves as a
 brief, hands-on introduction to some Quilt basics.
 
+## Install Node
+Quilt specs are written in Node. Only Node version v7.10.0 has been tested.
+
+Installation instructions for various operating systems are available
+[here](https://nodejs.org/en/download/package-manager/).
```
tejasmanohar: i'd like this https://nodejs.org/en/download/ ... has point-and-click installations

docs/GettingStarted.md
tejasmanohar: you can `npm install` git URIs... that's nicer than cloning manually though just publishing `quilt-nginx` (or @quilt/nginx -- i thing orgs are a thing now) on NPM sounds ideal

docs/GettingStarted.md
tejasmanohar: i think you can get `quilt`, not `@quilt`-- `@quilt` is an org

tejasmanohar: nvm https://www.npmjs.com/package/quilt quilt is taken... maybe you can get `@quilt/_` or something though `quilt/quilt` may just be simpler

docs/GettingStarted.md
```diff
@@ -89,7 +92,7 @@ var baseMachine = new Machine({
 
 ### Deploying `quilt/nginx/main.js`
 In one shell, start the Quilt daemon with `quilt daemon`. In another shell,
-execute `quilt run github.com/quilt/nginx/main.js`. Quilt will set up several
+execute `quilt run ./main.js`. Quilt will set up several
 Ubuntu VMs on your cloud provider as Workers, and these Workers will host Nginx
 Docker containers as specified in
```
tejasmanohar: unrelated to this pr but it may be worth explaining what docker is or linking somewhere sometime... i don't think the average node hobbyist knows what it is

minion/engine.go
kklin: Public connections don't support ranges. I'll add a comment.

.travis.yml
kklin: Unfortunately, travis doesn't support multiple languages(https://github.com/travis-ci/travis-ci/issues/4090)

Makefile
kklin: Yep

quiltctl/command/run_test.go
kklin: Ah, that was from when I was cleaning up the tests.. I'll just remove the test because `quilt run` depends on Node now.

kklin: Actually, the dependency is easy to mock out. I just fixed up the test.

README.md
kklin: I switched it to be all `var`s.

docs/GettingStarted.md
kklin: Made a trello card.

docs/GettingStarted.md
kklin: From a quick search, `npm` always installs into the `node_modules` directory. So we could `npm install @quilt/nginx`, but the source would end up in `./node_modules/@quilt/nginx` rather than just being cloned into the current directory.

Please correct me @tejasmanohar if you know a way to clone and install dependencies through npm.

tejasmanohar: you are correct, but shouldn't that be good enough? shouldn't the user's spect just need to require it and write the deployment code themselves?

kklin: This is describing how to get the example deployment code.

kklin: So it's not really downloading a dependency, more like downloading project code.

tejasmanohar: makes sense. i think it could be worth reorganizing the guide so that users dont think git clone is the way they download specs... that's what "In order to run the Nginx spec, we'll have to download it first" makes it sound like IMO

maybe asking them to copy example code from the doc to their editor is better or cloning the repo and cd'ing to example/ or something to get the point across

PR #881, opened 2017-05-22
travis: Automatically publish binaries for tags

kklin:
As part of our release process, we publish binaries of `quilt` for
various operating systems. Manually compiling and uploading the binaries
is very cumbersome, so we now have travis automatically update the
release when changes to a tag are made.


kklin:
@Baisang they get published to the release page: https://github.com/quilt/quilt/releases. A new release is automatically created for each tag.

PR #882, opened 2017-05-22
stitch: Provide a useful error when Node isn't installed

kklin:
To compile Quilt specs, we shell out to Node. Before, if Node wasn't
installed, we displayed only a blank error message. We now check that
Node is installed before shelling out to it, and return a more helpful
error if it is not installed.


stitch/stitch.go
```diff
@@ -109,6 +109,10 @@ func (stitchr Range) Accepts(x float64) bool {
 
 // FromFile gets a Stitch handle from a file on disk.
 func FromFile(filename string) (Stitch, error) {
+	if err := exec.Command("command", "-v", "node").Run(); err != nil {
```
kklin: @ejj this breaks compatibility on Windows... Should we just not support Windows for now?

kayousterhout: Does it make sense to print the actual err here too?

kklin: I think the error would be redundant, and since this is user facing it'd be better to just have a simpler output. The error would be something like "exited with code 1".

kayousterhout: Ah I see that does seem pretty useless.

kklin:
Huh, it's failing to find Node on travis for some reason. Investigating.

kayousterhout:
Ok since this is failing anyway -- can you make this slightly more helpful? like "Failed to find Node; ensure that Node.js is installed"? (helpful to have more of a suggestion for how the user should fix it)

kklin:
@kayousterhout I made a couple changes, could you just review it again?

kklin:
No.. This code is kind of tough to unit test because it's basically just shelling out to Node.

kayousterhout:
I'm just wondering if we should have a super simple test that reads a very minimal stitch and makes sure that there isn't an error? Since otherwise the existing test isn't that useful, since that test would pass even if the code *always* exited with an error.

kklin:
I see your point, but I'd like the unit tests to not depend on Node. What do you think of mocking out the `exec` call to Node, and replacing it with a function that always returned the same JSON string during unit tests?

kayousterhout:
That makes sense re: not wanting to depend on Node.  Mocking out the call seems fine to me, or also fine to just skip this if you'd prefer.

kklin:
It's nice to have more coverage:  I'll add it to my todo list.

PR #883, opened 2017-05-23
travis: Don't run `docker-build-quilt` and `release` in parallel

kklin:
The `-j 2` flag caused `make docker-build-quilt` and `make release` to
run in parallel. This resulted in a race condition where `make release`
would write the Darwin version of the Quilt binary to `./quilt`, and the
Dockerfile would copy it in (rather than the Linux version). The Darwin
version isn't runnable on Linux, so the minion container was completely
broken.

This patch avoids this problem by running `make release` after `make
docker-build-quilt`.


ejj:
FWIW, I think the correct solution to this problem, is actually to have the platform specific build targets tag their binaries.  I.e. `quilt_darwin`, `quilt_linux` etc.  We still have a bug in our Makefile, this patch just changes it so that travis doesn't hit it.

PR #891, opened 2017-05-26
changelog: Prepare for 0.0.2 release

kklin:


kklin:
This is just my first stab at the changelog notes.. I wasn't sure what tone to use, and what resolution the changes should be described in. I basically generated the changes by doing `git log v0.0.1..`.

Looking for feedback from @luise @kayousterhout @ejj in particular.

CHANGELOG.md
```diff
@@ -1,6 +1,31 @@
...
+- Fix a bug where inbound and outbound public traffic would get randomly
+dropped.
+- Support floating IP assignment in DigitalOcean.
+- Support arbitrary GCE projects.
+- Upgrade to OVS2.7.
+- Fix a race condition where the minion boots before OVS is ready.
+- Build the OVS kernel module at runtime if a pre-built version is not
+available.
+- Evaluate specs using Node.js.
+
```
luise: Just for clarity I'd add `Release 0.0.1` - just without comments, so there isn't this weird gap.

CHANGELOG.md
```diff
@@ -1,6 +1,31 @@
 Quilt Change Log
 ================
 
+Release 0.0.2
+-------------
+
+Release 0.0.2 most notably modifies `quilt run` to evaluate Quilt specs using
+Node.js, rather than within a Javascript implementation written in Go. This
+enables users to make use of many great Node features, such as module
+distribution, versioning, unit testing, a rich ecosystem of modules, and `use
```
luise: Are "module distribution" and "rich ecosystem of modules" similar enough that they could be combined?

CHANGELOG.md
```diff
@@ -1,6 +1,31 @@
 Quilt Change Log
 ================
 
+Release 0.0.2
+-------------
+
```
luise: Overall, I like the combination of the initial paragraph explaining Node and the bullet points for the smaller fixes. In general, I think it would be nice to have most updates in the bullet form, but offer a short explanation of major changes, like using Node.js, that dramatically change the user's workflow or experience.

CHANGELOG.md
```diff
@@ -1,6 +1,31 @@
...
+Release 0.0.2 most notably modifies `quilt run` to evaluate Quilt specs using
+Node.js, rather than within a Javascript implementation written in Go. This
+enables users to make use of many great Node features, such as module
+distribution, versioning, unit testing, a rich ecosystem of modules, and `use
+strict`. In order to facilitate this, we now require `node` version 7.10.0 or
+greater as a dependency to `quilt run`.
+
+This release also improves cloud provider support.
```
luise: I think we could leave this out and rely on the bullets below.

CHANGELOG.md
```diff
@@ -1,6 +1,31 @@
...
+
```
luise: I'd add a comment like "What's new:,  "Updates:", or something like that before the bullets.

CHANGELOG.md
kklin: I meant "rich ecosystem of modules" to mean modules written by the open source community, and "module distribution" to mean `npm`.. I'll just call "module distribution" "package management".

kklin:
@luise addressed your comments

kayousterhout:
Overall this looks good (sorry for the slow review).  One thing I noticed in Spark's changelog is that they explicitly separate API changes (one section) from performance / stability fixes (in another section).  I think it would be good to do this, so that users can easily skin the top API changes part to see what potentially non-backward-compatible changes occurred (seems especially important as we do some major changes in the beginning).

Also, should we be updating this as new bugs are fixed (e.g., have a "future release" section that we keep up to date)?

kklin:
I'm not sure what we'd consider API changes. I definitely think we should make breaking changes obvious.

It would be nice to have some sort of sections to make the changelog more readable. E.g. client, network, cloud providers.

I think updating the changelog as we go will make the contribution process difficult. I'm imagining people submitting PRs, and us having to go through a back and forth telling contributors to update the changelog. I don't feel strongly though.

Anyways, I'm going to merge this as is now because I want to get the release out, and we can refine the changelog format in the next release.

kayousterhout:
Oh yeah that's a good point re: contribution process.  I agree that it's better to trade some additional work for whoever does the release for an easier process for contributors.

ejj:
FWIW, RE change log.  I think we internally should do it, but we shouldn't gate contributions on it.  I.E. if some community member submits a PR, we could submit another PR on top of it that fixes up the change log.

PR #892, opened 2017-05-28
Vagrant

Baisang:
So one major thing is the network patch that @kklin made that doesn't restrict NAT rules by interface that is part of this PR. I'm not too sure what the specific consequences of that may be. I'll try to run `quilt-tester` on this with aws, but booting random specs seems to work on vagrant. I will write up/update the docs on that, possibly in a second PR.

But this patch should contain the general fixes to make Vagrant work and some small immediate improvements.

### DIFF
--- a/cluster/vagrant/api.go
+++ b/cluster/vagrant/api.go
@@ -55,13 +55,6 @@ func initMachine(cloudConfig string, size string, id string) error {
 	path := vdir + id
 	os.Mkdir(path, os.ModeDir|os.ModePerm)
 
-	_, stderr, err := shell(id, `vagrant --machine-readable init coreos-beta`)
-	if err != nil {
-		log.Errorf("Failed to initialize Vagrant environment: %s", stderr)
-		destroy(id)
-		return errors.New("unable to init machine")
-	}
-
 	err = util.WriteFile(path+"/user-data", []byte(cloudConfig), 0644)
 	if err != nil {
 		destroy(id)
--- a/cluster/vagrant/api.go
+++ b/cluster/vagrant/api.go
@@ -8,66 +8,76 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"text/template"
 
 	log "github.com/Sirupsen/logrus"
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/quilt/quilt/util"
 )
 
-var vagrantCmd = "vagrant"
-var shCmd = "sh"
+const vagrantCmd = "vagrant"
+const shCmd = "sh"
+const cloudConfigPath = "/user-data"
+const sizePath = "/size"
+const vagrantFilePath = "/Vagrantfile"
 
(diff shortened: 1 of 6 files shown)


luise:
@Baisang Is this ready for review or should I wait?

Baisang:
@luise it's ready to review, but there may need to be discussion on the NAT change

kklin:
Yeah, I'm not sure what the implications are of removing the interface from the NAT rules. I think the rule will apply to all interfaces, which may behave weirdly with the container networking stacks? @ejj ?

Maybe we should apply the rule on `quilt-int` for outbound connections? Did you test outbound connections with this change btw, @Baisang?

Also, can you modify my email to be kevin@quilt.io in the NAT commit? Thanks!

ejj:
FWIW, I don't understand NAT in linux.  No idea what will happen.

Baisang:
@kklin will do, I haven't tested outbound connections. would this be just going onto a worker machine and trying to curl google.com or something?

kklin:
@Baisang yep

Baisang:
@kklin, going into a worker machine and curling google worked. Same with going into the minion container. Going into an application container (e.g. node.js) and curling did not work (this is intentional, right?).

The outbound_public test worked also

cluster/vagrant/api.go
```diff
@@ -14,60 +15,80 @@ import (
...
+	}
 
-const boxVersion = "20170515.0.0"
+	return vagrantFileBytes.String()
+}
 
+// initMachine creates the files necessary to initialize a vagrant machine.
 func initMachine(cloudConfig string, size string, id string) error {
+
+	// By default, we store vagrant information in ~/.vagrant
```
kklin: This comment doesn't seem very useful to me, and could get stale easily.

cluster/vagrant/api.go
```diff
@@ -14,60 +15,80 @@ import (
...
+		panic(err)
+	}
 
-const boxVersion = "20170515.0.0"
+	return vagrantFileBytes.String()
+}
 
+// initMachine creates the files necessary to initialize a vagrant machine.
 func initMachine(cloudConfig string, size string, id string) error {
+
```
Baisang: my habit is to put a newline before comments, but will just remove the comment

cluster/vagrant/api.go
```diff
@@ -14,60 +15,80 @@ import (
...
+	vagrantFile := createVagrantFile(ram, cpus)
+
+	err = util.WriteFile(path+vagrantFilePath, []byte(vagrantFile), 0644)
 	if err != nil {
 		destroy(id)
 		return err
 	}
 
-	err = util.WriteFile(path+"/size", []byte(size), 0644)
+	err = util.WriteFile(path+sizePath, []byte(size), 0644)
```
Baisang: i see what you're saying. in that case I think just having the file and making the vagrantfile read from the file would be best.

Baisang:
@kklin made the changes you suggested, please review again

PR #897, opened 2017-05-31
ovs: Update prebuilt kernel version for DigitalOcean

kklin:
DigitalOcean updated the kernel version in its Ubuntu image.


kklin:
Will need to retrigger the build after `quilt/ovs` builds on DockerHub (we need to fix this)

PR #899, opened 2017-06-01
version: Change default version name to `dev`

kklin:
This change makes the image associated with the master branch have the
tag `dev`. This is better because it matches the tag we use to track the
HEAD of the master branch. Furthermore, this will cleanly fix a build bug
introduced by `travis: Push Docker images for tags rather than branches`
because we won't require special logic for the `master` branch.


kklin:
fyi @luise right now, merges to master won't properly update the minion, so this is fairly important

PR #901, opened 2017-06-02
Renamed spec to blueprint in all code

kayousterhout:
We've decided that blueprint is a better name than spec, so
this commit renames all code and comments to use blueprint rather
than spec.


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Quilt Change Log
 ================
 
+Release 0.1.1 (unreleased)
```
kklin: I have an unmerged commit  that calls this "Up Next" or something like that, without the specific release number. I think the release number should get assigned when we're actually doing the release because some features won't go into the next patch release.

Baisang:
as ethan said just don't forget travis

PR #903, opened 2017-06-06
Fix readme / spec tests' dependence on execution order

kayousterhout:


kklin:
@kayousterhout not sure if you noticed, but Travis failed.

kayousterhout:
Hm let's see if Travis likes this version better. It seems like Getwd didn't work correctly on Travis, for reasons I don't understand, but this solution is better anyway I think, since it's more consistent with how the other tests get run.

PR #904, opened 2017-06-08
stitch: Usability improvements

kayousterhout:
This PR combines various usability improvements to stitch to give better error messages to the user (e.g., to fix the problem @mchang6137 ran into with connect) and to rename connect to allowFrom.  See the individual commit messages for more detail.


stitch/stitch.go
```diff
@@ -129,6 +129,9 @@ func FromFile(filename string) (Stitch, error) {
 	if err != nil {
 		return Stitch{}, errors.New(stderr.String())
 	}
+	// If there wasn't an error, still print stderr, in case there were any
+	// warnings or other non-fatal errors.
+	fmt.Print(stderr.String())
```
kklin: We should probably print to stderr as well.

luise:
@kayousterhout I think the `allowFrom` example in the commit example should swap `a` and `b`.

stitch/bindings_test.js
```diff
@@ -468,6 +468,12 @@ describe('Bindings', function () {
             expect(() => publicInternet.connect(new PortRange(80, 81), foo)).to
                 .throw('public internet cannot connect on port ranges');
         });
+        it('connect to non-service', function () {
+            expect(() => foo.connect(10, 10)).to
+                .throw('Services can only connect to other services. ' +
+                    'Check that you\'re connecting to a service, and not ' +
```
kklin: Nit - can you just double quote this string so we don't need the escape in "you're"?

stitch/bindings.js
```diff
@@ -26,7 +26,10 @@ function githubKeys(user) {
 // The default deployment object. createDeployment overwrites this.
 global._quiltDeployment = new Deployment({});
 
-// The label used by the QRI to denote connections with public internet.
+// The name used to refer to the public internet when describing network
+// connections to the QRI (connections to other services are referenced by
```
kklin: I had to reread this a couple times -- "network connections to the QRI" sounds like it's creating some sort of network ACL between the QRI and the deployment.

stitch/bindings.js
```diff
@@ -297,24 +300,43 @@ Service.prototype.deploy = function(deployment) {
...
+            "Check that you're allowing connections from a service, and " +
+            "not from a Container or other object.");
+    }
+    this.allowedInboundConnections.push(
+        new Connection(sourceService, portRange));
 };
 
 // publicInternet is an object that looks like another service that can be
 // connected to or from. However, it is actually just syntactic sugar to hide
 // the connectToPublic and connectFromPublic functions.
```
kklin: This comment still references `connectToPublic` and `connectFromPublic`.

stitch/bindings.js
```diff
@@ -323,18 +345,32 @@ var publicInternet = {
 
 // Allow outbound traffic from the service to public internet.
 Service.prototype.connectToPublic = function(range) {
+    console.warn("Warning: connectToPublic is deprecated; switch to using " +
+        "allowToPublic.");
+    this.allowToPublic(range);
+}
+
+Service.prototype.allowToPublic = function(range) {
```
kklin: I had to reread the name `allowToPublic` a couple times. I'd prefer `allowOutboundPublic`, what do you think?

stitch/bindings_test.js
```diff
@@ -407,7 +406,79 @@ describe('Bindings', function () {
...
+                .throw('public internet can only connect to single ports and not to port ranges');
+        });
+        it('connect from publicInternet port range', function () {
+            expect(() => foo.allowFrom(publicInternet, new PortRange(80, 81))).to
+                .throw('public internet can only connect to single ports and not to port ranges');
+        });
+        it('allowFrom non-service', function () {
+            expect(() => foo.allowFrom(10, 10)).to
+                .throw('Services can only connect to other services. ' +
+                    'Check that you\'re allowing connections from a service, and not ' +
```
kklin: Same thing here re escaping `you're`

stitch/bindings_test.js
kayousterhout: The JS style guides I've seen suggest choosing a quote style, either single or double quotes, and using that consistently (the two I looked at said single, which is what's already in this file, e.g., https://github.com/felixge/node-style-guide#use-single-quotes) -- but I checked the Google JS style guide, which seems like a reasonable one to follow, and that said to use a template string to avoid escaping (https://google.github.io/styleguide/jsguide.html#features-string-literals) so I switched to that

stitch/bindings.js
kayousterhout: Makes sense -- take a look at the new comment and let me know if that's better

stitch/bindings.js
kayousterhout: I agree that's better -- I also thought allowToPublic sounded weird but couldn't come up with an alternative.

kayousterhout: Do you think allowFrom should be allowInboundPublic? Either name has an inconsistency -- I'm not sure which inconsistency is better.

kklin: I think it's fine leaving it as is. These aren't user-facing methods, so I'm not too concerned.

PR #905, opened 2017-06-08
cluster: Speed up machine listing

kklin:


cluster/cluster.go
```diff
@@ -432,14 +433,32 @@ func syncDB(cms []machine.Machine, dbms []db.Machine) syncDBResult {
...
+
 func (clst cluster) get() ([]machine.Machine, error) {
-	var cloudMachines []machine.Machine
+	var wg sync.WaitGroup
+	cloudMachinesChan := make(chan listResponse, len(clst.providers))
 	for inst, p := range clst.providers {
-		providerMachines, err := p.List()
-		if err != nil {
-			return []machine.Machine{}, fmt.Errorf("list %s: %s", inst, err)
+		wg.Add(1)
```
Baisang: There's no difference between adding 1 to the wait group each time vs. adding them all at once, right?

kklin: I think adding 1 each time is slightly better. I could imagine that someone in the future adds something like
```
if (something) {
  continue // Don't check machines on this provider because it's broken.
}
```

If that patch didn't modify the wait group, the code would hang forever. To put it another way, I like adding to the wait group as close as possible to where the go function is spawned.

PR #907, opened 2017-06-09
registry: Don't attempt to run registry logic on Workers

kklin:
Only Masters require running the registry module. Before, Workers never
exited the `bootWait` loop because Workers don't run the registry
container.


minion/registry/registry.go
```diff
@@ -28,19 +28,17 @@ restart containers running the custom image.
 
 // Run builds Docker images according to the Image table.
 func Run(conn db.Conn, dk docker.Client) {
-	bootWait()
+	if conn.MinionSelf().Role != db.Master {
```
kayousterhout: Can you add a comment about why this check is here, or alternately (better I think), change the method comment to say "Run builds Docker... table, if this is the master, and otherwise does nothing."

minion/registry/registry.go
```diff
@@ -28,19 +28,17 @@ restart containers running the custom image.
...
+		return
+	}
 
+	bootWait()
 	for range conn.TriggerTick(30, db.ImageTable).C {
 		runOnce(conn, dk)
 	}
 }
 
 func runOnce(conn db.Conn, dk docker.Client) {
```
kayousterhout: As long as you're fixing things in here, can you add a method comment for runOnce? It was hard me to figure out what this and run each do.

minion/registry/registry_test.go
```diff
@@ -10,35 +10,18 @@ import (
 	"github.com/quilt/quilt/minion/docker"
 )
 
-func TestRunMaster(t *testing.T) {
+func TestBuild(t *testing.T) {
```
kayousterhout: Can this just be "TestRunOnce", so it's obvious that it's testing the "runOnce" method -- which then is consistent with the other test names in the class (or alternatively, should "Run" be renamed to "Build"? Or not because then it's inconsistent with the "Run" methods in other classes?

kklin:
@kayousterhout addressed your comments!

PR #910, opened 2017-06-13
quilt: Add option for TLS-encrypted GRPC communication

kklin:


kklin:
Should someone other than @Baisang look at this since it's a big PR, and he's busy with Yelp?
I'd like @kayousterhout or @ejj to look over this as well anyways, since it changes a lot of code.

kayousterhout:
I can look at this

quiltctl/command/daemon.go
```diff
@@ -42,6 +42,11 @@ func (dCmd *Daemon) Parse(args []string) error {
 	return nil
 }
 
+// FlagsParsed makes any necessary post-parsing transformations.
+func (dCmd *Daemon) FlagsParsed() error {
```
kayousterhout: Actually what about BeforeRun() and AfterRun()?

kklin: I like that a lot!

quiltctl/command/flag.go
```diff
@@ -20,24 +22,44 @@ type flagParser interface {
...
 	if !cf.initLocalClient {
 		return nil
 	}
 
-	cf.localClient, err = clientGetter.Client(cf.host)
+	cf.localClient, err = clientGetter.Client(cf.host, cf.creds)
 	return err
 }
+
+func authFromDirectory(dir string) (auth.Auth, error) {
```
kklin: I agree with the sentiment that it'd be better if as much of the TLS stuff was in the same package as possible, but treating a lack of a directory as `auth.Insecure` seems like a very command-parsing specific thing. It seems like if we moved it into the `auth` package some of the quirks of the command line processing would leak into that package.

It's just a gut feeling though, and I don't feel that strongly about it. What do you think?

api/client/client.go
```diff
@@ -90,6 +90,12 @@ func New(lAddr string) (Client, error) {
 	cc, err := grpc.Dial(addr, grpc.WithDialer(dialer), grpc.WithInsecure(),
 		grpc.WithBlock(), grpc.WithTimeout(connectTimeout))
 	if err != nil {
+		if err == context.DeadlineExceeded {
+			err = daemonTimeoutError{
+				host:         lAddr,
+				connectError: err,
+			}
+		}
```
kklin: We could definitely give more helpful errors, but I think of that as part of the usability patches that will have to follow after the main PR has been merged.

api/client/client.go
```diff
@@ -87,8 +88,8 @@ func New(lAddr string) (Client, error) {
 	dialer := func(dialAddr string, t time.Duration) (net.Conn, error) {
 		return net.DialTimeout(proto, dialAddr, t)
 	}
-	cc, err := grpc.Dial(addr, grpc.WithDialer(dialer), grpc.WithInsecure(),
-		grpc.WithBlock(), grpc.WithTimeout(connectTimeout))
+	cc, err := grpc.Dial(addr, grpc.WithDialer(dialer), grpc.WithBlock(),
+		grpc.WithTimeout(connectTimeout), creds.Client())
```
kklin: I think this sounds much cleaner! Something felt off about the way `auth.Auth` was being passed around, but I didn't think of pulling out all of the client getting.

```
I don't know if foreman could use the same getClient function? Is there a reason it includes less complexity around the dialer?
```
The foreman was just written earlier than the API client. They could use the same connection settings (and it's probably better that they do).

```
This would move almost all of the complexity (everything except Close) of authentication and grpc to one place.
```
I think it's unavoidable for the caller to call `Close`. Unless you wrapped the client method calls with something that got a client, made the call, and closed it. But it seems to be a common Go pattern for the caller to `defer client.Close()`, so I'm not too worried.

auth/tls.go
```diff
@@ -0,0 +1,109 @@
...
+		VerifyPeerCertificate: tlsAuth.verifyPeer,
+		// Skip the built-in TLS verification, and just use our custom
+		// VerifyPeerCertificate function.
+		InsecureSkipVerify: true,
+	}))
+}
+
+// The built-in client TLS implementation verifies the hostname of the certificate
+// matches the connection. However, it's cumbersome to generate certificates with
+// proper hostnames because doing so would require sending keys and certificates
```
kklin: Eventually, it could if you generated the keys at runtime (after the machines have booted, and you know their IPs). I wouldn't expect that to happen immediately though -- I was thinking that we would first implement a secure key distribution system (i.e. SCP) that still distributed the pre-generated keys, and then we could talk about auto-generating keys.

cluster/cloudcfg/cloudcfg.go
```diff
@@ -33,16 +49,46 @@ func Ubuntu(keys []string, role db.Role) string {
...
+		TLSDirectory:  tlsDirectory(creds),
 	})
 	if err != nil {
 		panic(err)
 	}
 
 	return cloudConfigBytes.String()
 }
+
+func tlsDirectory(creds auth.Auth) TLSDirectory {
```
kklin: I could see that, do you have any ideas for how we could make generating the `TLSDirectory.Enabled` without using a type switch? Is the only way to make the `auth.Auth` interface have a `IsTLS` method? Maybe this will solve itself when we switch to the alternate `auth` structure you proposed in the other comment.

cluster/cloudcfg/cloudcfg.go
```diff
@@ -33,16 +49,46 @@ func Ubuntu(keys []string, role db.Role) string {
...
+	switch a := creds.(type) {
+	case auth.Insecure:
+		return TLSDirectory{Enabled: false}
+	case auth.TLS:
+		return TLSDirectory{
+			Enabled: true,
+			Root:    "/etc/quilt/tls",
+			Files: []File{
+				{
+					Name:     auth.CAPrefix.CertName(),
```
kklin: The worker needs to verify that inbound connections are from a trusted source (i.e. they have a certificate signed by the CA).

quiltctl/command/flag.go
kayousterhout: Good point re: directory -- I agree that the lack of directory logic should stay here, so just those 3 lines could move to the auth package.

docs/Security.md
```diff
@@ -0,0 +1,52 @@
+# Security
+
+## TLS
+Quilt uses [grpc](http://www.grpc.io/) for communication with the daemon and
+deployed clusters. Functionality exposed through `grpc` include deploying new
+Stitches and querying deployment information. Thus, TLS should be enabled for
```
Baisang: s/Stiches/blueprints ?

docs/Security.md
```diff
@@ -0,0 +1,52 @@
...
+all non-experimental deployments.
+
+### Quickstart
+```
+# Generate the necessary TLS files.
+$ quilt setupTLS ~/.quilt/tls
```
Baisang: nit but can the command be called `setup-TLS` instead? since that's how most commands split 2 words. eh but that would be awkward since we have` -tlsDir` already and it'd have to be `--tls-dir` or something, so maybe don't do this

docs/Security.md
```diff
@@ -0,0 +1,52 @@
...
+
+# Start the daemon with TLS enabled.
+$ quilt daemon -tlsDir ~/.quilt/tls
+
+# Interacting with the daemon or cluster requires TLS authentication.
+$ quilt run -tlsDir ~/.quilt/tls specs/example.js
```
Baisang: I think this will make it harder for users to just query and see what quilt is doing. Would it be possible to have the `tlsDir` be configured somehow, so that users won't have to pass it in every time for the commands? e.g. check in a default directory unless overridden by the flag

quiltctl/command/flag.go
```diff
@@ -16,14 +18,26 @@ type flagParser interface {
...
 }
 
-func (cf *commonFlags) FlagsParsed() error {
-	return nil
+func (cf *connectionFlags) FlagsParsed() (err error) {
+	return cf.setupClient(getter.New())
+}
+
+func (cf *connectionFlags) setupClient(clientGetter client.Getter) (err error) {
+	if !cf.initLocalClient {
```
kklin: It's cleaner to inject dependencies via method arguments when possible IMO. If we injected it via an instance variable, we would have to ensure that all users of `connectionFlags` obtained one that has the proper `clientGetter` configured. The down side is we can't cover this line (https://github.com/quilt/quilt/pull/910/files/82693a6aa4e514f025c4c9ec97d422a302e829f6..19d21fc37caf89c63a273fe2adca4b67d8d0c18f#diff-4718997e54c3828de49063271ed2350fR33), but I think that's fine.

Makefile
```diff
@@ -2,6 +2,7 @@ export GO15VENDOREXPERIMENT=1
 PACKAGES=$(shell govendor list -no-status +local)
 NOVENDOR=$(shell find . -path -prune -o -path ./vendor -prune -o -name '*.go' -print)
 LINE_LENGTH_EXCLUDE=./api/pb/pb.pb.go \
+		    ./api/client/mocks/Getter.go \
```
kklin: It's autogenerated

kayousterhout: OHH did not realize that.  Thanks.

docs/Security.md
kklin: We're actually going to make the daemon proxy connections to the cluster, and make it so unix sockets are unencrypted. This will basically remove the need for the `tlsDir` flag.

docs/Security.md
kklin: I didn't think of that, `setup-tls` would certainly be more consistent with `debug-logs`.

PR #911, opened 2017-06-14
minor: Fix typo in comment

kayousterhout:


kklin:
(cc @ejj)

We haven't used this `minor` category before, but I kind of like it. (I think before, we would have used `quilt: ...` because it's in `quilt.go`). Worth discussing if we should make this a standard.

kayousterhout:
One other alternative is "docs:" -- but I like minor as a signal to people skimming the commits that it's safe to ignore

kklin:
I'd like to save docs for our documentation folder (or is that getting moved?)

kayousterhout:
Makes sense (and it won't get moved)

ejj:
I don't particularly have an opinion on the minor thing.  I likely won't use it, but I think its fine.

PR #912, opened 2017-06-14
cluster: Readability improvements

kayousterhout:
This PR does 2 renamings and one comment update in an attempt to make the cluster and db code easier to understand.


api/server/server.go
```diff
@@ -68,7 +68,7 @@ func (s server) Query(cts context.Context, query *pb.DBQuery) (*pb.QueryReply, e
 	var rows interface{}
```
kklin: Would `FilterMachines` be a better name? I don't think `MachinesWithFilter` is very intuitive.

My other concern is that a lot of the database tables have this `SelectFromFoo` method.. we should probably rename all of them for consistency.

kayousterhout:
The Select change is much broader, as you brought up, so I removed it from this PR for now.

PR #914, opened 2017-06-16
minion: Upgrade to version 17.05.0 of Docker engine

kklin:
Although there is no specific feature that we require, using the latest
version of Docker ensures that we have the latest bug fixes, and will
make upgrading in the future easier.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Up Next
 - Package the OVS kernel module for the latest DigitalOcean image to speed up
 - Renamed specs to blueprints.
 - Load balancing.
+- Upgraded to the latest Docker engine version.
```
kayousterhout: can you specify what version?

kklin: Done

PR #915, opened 2017-06-16
ipdef: Fix load balancer IP

kklin:
A previous commit set it to 10.0.0.1, which conflicted with the gateway.


kklin:
Sloppy on my part..

PR #916, opened 2017-06-19
Simplify client getting

kklin:
This PR contains various changes to the API for obtaining Quilt clients to facilitate enforcing TLS encryption.

The commits are basically the same as the previous (unmerged) TLS PR, with the exception of the new commit to proxy connections through the daemon.


kklin:
The associated quilt-tester PR will need to be merged at the same time as this https://github.com/quilt/tester/pull/41

kklin:
@kayousterhout given that you already read through most of this series, I think you should review it.

quiltctl/command/flag.go
```diff
@@ -34,7 +33,7 @@ func (cf *connectionFlags) InstallFlags(flags *flag.FlagSet) {
 }
 
 func (cf *connectionFlags) BeforeRun() error {
-	return cf.setupClient(getter.New())
```
kayousterhout: I wonder if it makes sense to do this 1-line change, and then the corresponding changes in the tests, in your earlier PR (instead of the comment I left there about defining something like client.New here)?

api/server/server.go
```diff
@@ -64,23 +73,69 @@ func Run(conn db.Conn, listenAddr string) error {
...
+		}
+		defer leaderClient.Close()
+
+		rows, err = actions.proxy(leaderClient)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		if actions.noProxy == nil {
+			return nil, fmt.Errorf("unrecognized table: %s", query.Table)
```
kayousterhout: (this is irrelevant if you implement my other suggestion above)

quiltctl/command/daemon.go
```diff
@@ -42,6 +42,16 @@ func (dCmd *Daemon) Parse(args []string) error {
 	return nil
 }
 
+// BeforeRun makes any necessary post-parsing transformations.
+func (dCmd *Daemon) BeforeRun() error {
+	return dCmd.common.BeforeRun()
```
kklin: My original thinking was that if it was an anonymous member variable, we wouldn't be able to override it in `InstallFlags`, and still call the `common` implementation.

But I was doing something similar in some other code today, and I think you can just reference `dCmd.<struct_name>`, so I'll try it out!

quiltctl/command/flag.go
```diff
@@ -9,6 +9,15 @@ import (
 type flagParser interface {
 	// InstallFlags sets up parsing for command line flags.
 	InstallFlags(*flag.FlagSet)
+
+	// BeforeRun is called after command line flags have been parsed, but
+	// before the Run method is called. It gives commands an opportunity to
+	// transform parsed flags into a more consumable form.
+	BeforeRun() error
```
kklin: I'm not sure what's better. If we did that approach, then the `Run()` method of the `connection` would be a no-op, which feels funny to me.

quiltctl/command/flag.go
```diff
@@ -20,18 +22,33 @@ type flagParser interface {
 	AfterRun() error
 }
 
-type commonFlags struct {
-	host string
+type connectionFlags struct {
+	host   string
+	client client.Client
+
+	initClient bool
```
kklin: This is necessary because some subcommands don't run as clients (namely `daemon`, which uses the host flag to spawn a server). Do you think that the daemon should handle parsing the host in a special way, and that the other subcommands should use this helper?

quiltctl/command/log.go
```diff
@@ -86,15 +86,9 @@ func (lCmd *Log) Parse(args []string) error {
 
 // Run finds the target container or machine minion and outputs logs.
 func (lCmd *Log) Run() int {
-	c, err := lCmd.clientGetter.Client(lCmd.common.host)
```
kklin: Yeah.. the whole `clientGetter` stuff was generally terrible.

quiltctl/command/flag.go
kayousterhout: Does connection need to have a Run() method? I don't think it does since it's never used as a Command?

quiltctl/command/flag.go
kayousterhout: Ah I didn't notice that.  Yeah I do think it's better for the daemon to just parse the host itself, since esp. once you add the TLS stuff, it's going to be getting more and more unnecessary goop by using this class

quiltctl/command/flag.go
kklin: Ah, you're right, I was confused

api/server/server.go
```diff
@@ -64,23 +73,69 @@ func Run(conn db.Conn, listenAddr string) error {
 	return nil
 }
 
+func newAPIServer(conn db.Conn, shouldProxy bool) server {
+	s := server{
+		conn:        conn,
+		shouldProxy: shouldProxy,
+	}
+	s.queryHandlers = map[db.TableType]queryAction{
+		db.MachineTable: {
```
kklin: This made the code much easier to follow, thanks for the suggestion!

quiltctl/command/flag.go
kklin: Hm.. I'm wavering on this because it'd be nice to ensure consistency in the name of common flags (specifically, `--host` and `--tlsDir`). I suppose we could take the embedding a step further and do

```
type connectionFlags struct {
  host, tlsDir string
}

type connectionHelper struct {
  client client.Client
  connectionFlags
}
```

The daemon would just embed `connectionFlags`, while stuff like `ps` would embed the `connectionHelper`.

What do you think?

api/server/server.go
```diff
@@ -137,3 +192,100 @@ func (s server) Version(_ context.Context, _ *pb.VersionRequest) (
...
+			return nil, resp.err
+		}
+		containers = append(containers, resp.containers...)
+	}
+	return containers, nil
+}
+
+// updateContainers returns a list of containers from the hash join
+// of the leader's view of containers and the workers' views of containers.
+func updateContainers(lContainers []db.Container, wContainers []db.Container) (
```
kklin: I renamed it "updateLeaderContainerAttrs"

quiltctl/command/flag.go
kayousterhout: Hmm I see the issue -- especially once tlsDir is added.  I think the further embedding you suggested sounds like the best approach.

quiltctl/command/flag.go
kklin: I think you meant to have `setupClient` take `client.New` rather than `clientGetter`? There's no type yet for the `client.New` signature at this point in the history (it's inside the `client.Getter` interface). I don't think it's worth it to introduce the type, and make the change here because all we gain is a slightly cleaner diff in the later commit right?

api/server/server.go
```diff
@@ -24,18 +26,25 @@ import (
...
 
 type server struct {
-	conn db.Conn
+	conn          db.Conn
+	shouldProxy   bool
+	queryHandlers map[db.TableType]queryAction
+}
+
+type queryAction struct {
+	noProxy func() interface{}
```
kklin: n/a because I implemented your `QueryWithProxy` approach instead!

api/server/server.go
kklin: n/a because I implemented your `QueryWithProxy` approach instead!

kklin:
@kayousterhout I made your changes, but I realized that we shouldn't merge this until https://github.com/quilt/tester/pull/42 is merged because it's a lot easier to just apply https://github.com/quilt/tester/pull/41 in this PR.

I'll ping you once the other PR is merged, and this is ready for another round

kklin:
ok @kayousterhout, this is ready for another review

PR #917, opened 2017-06-20
Move integration tests from quilt/tester to quilt/quilt

kklin:
This PR preserves the history of the moved files from quilt/tester.


kklin:
@kayousterhout this is ready for review. The Travis build is expected to fail until the PR is merged because the blueprints tester is pointing at the yet-unmerged specs.

quilt-tester/junit.go
```diff
@@ -8,12 +8,14 @@ import (
 	"github.com/Sirupsen/logrus"
 )
 
+// JUnitReport defines the XML output schema for a test run.
```
kayousterhout: These new comments are awesome!  What is a test run (is it running all of the test cases?)?  If easy to describe, would be helpful to include in this comment.

quilt-tester/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@quilt/tester-dependencies",
+  "version": "0.0.1",
+  "license": "MIT",
+  "dependencies": {
+    "@quilt/quilt": "quilt/quilt",
```
kayousterhout: Can you alphabetize these?  I think it would be good to get in the habit of doing that for these dependency lists, since otherwise in long lists like this, it's easy to accidentally end up with duplicates

scripts/blueprints-tester/tests/test_blueprints.go
```diff
@@ -42,52 +42,61 @@ func run(name string, args ...string) error {
 }
 
 type blueprint struct {
-	repo, path string
```
kayousterhout: It's a little repetitive to specify all of the packageJsonFolders when they seem to all be the first directory in the blueprintPath -- it would be nice to either extract the folder by taking the first thing in the path (looks possibly difficult) or making the blueprintPath not include the folder and having line 30 join the folder and the path.

kklin: It's not quite always the first directory in the blueprintPath:
When it's a integration test spec, it's in the subdirectory `quilt-tester` of the `quilt/quilt` repo
Otherwise, it's in the root of the git folder.

IMO it's ugly, but I'm not _as_ concerned about the quality of this code.

kklin:
@kayousterhout address your comments!

PR #919, opened 2017-06-22
Docs: Move Quilt docs to use Slate

kayousterhout:
This series of commits adds Slate to our repository, which is used for translating Markdown files to HTML files, and then converts our docs to rely on Slate to be built.

Note that there is a big re-write of the docs that will come after this PR; the goal of this one is to move over to Slate and avoid major typos / issues with understanding the docs.


ejj:
I'm not sure how I feel about committing the slate repo directly to ours.  Is that really necessary?  Can't we have travis clone it (or a specific version) at build time?

ejj:
One high level comment.   A lot of people will find the docs by clicking around the source repo instead of the website.  I know I, personally, almost always do that.  One thing I don't like about embedding slate in the repo, is it muddies the docs directory, and makes it difficult to find the relevant content for people clicking around.

ejj:
Also, another issue with embedding it in our source, is it will be hard to incorporate updates from them.  I.E. if they had some nifty new features we can't just rebase them into our website.

kayousterhout:
The updates from them is the issue I was most concerned about at first -- but then I noticed that the Slate repo has changed fairly little after the initial implementation.  I also noticed that other users of Slate seem to slowly get farther and farther from the original Slate functionality, as they add more sophisticated formatting etc., which is the direction I'd expect our docs to move over time as well.

That being said, it sounds like you'd like the Slate code to be in its own repo for a variety of reasons, so I'll look into doing that tomorrow!

ejj:
I'm actually not sure what I want =)  I think I would be quite happy if, for example, the slate stuff was in it's own sub-directory of docs (not co-mingled with the actual documentation), and then we had a make target that copied our documentation into the slate sub-directory for compilation.  That would be a bit cleaner, seems pretty easy to do (i.e. a lot less work than pulling into a separate repo).

kayousterhout:
The copy didn't occur to me but seems kind of hack-y / prone to weird bugs...what about just making a sym link?

ejj:
Funny, for some reason I think a copy is actually less error-prone.  It's more inline with the Makefile-ish way to do things.  That said I don't have a strong argument for it, and either way I think would be cleaner than what we have now.  Up to you.

kayousterhout:
I just remembered one other reason I had for this directory structure -- I think it's good to somewhat discourage folks from looking at the raw markdown files, because if those are super obviously accessible, I'm worried folks will look at those instead of the more nicely compiled / formatted version at docs.quilt.io.  Do you think that's something we shouldn't be concerned about, and if they find / use the markdown version that's fine?

ejj:
Ah interesting, I actually was thinking precisely the opposite.  We should make the docs easy to access either in markdown form, or on the website.  I'm thinking of the use case where a) someone has cloned the repo and is in a terminal in which case the markdown is much easier to access than opening a browser and b) people who found us from, and are comfortable in, the github UI in which case they're going to prefer looking at the markdown.

kayousterhout:
That's tricky for us to do because the links are different: any place we point to another place in the docs, the links will only work in one version or the other (or we could change Slate to re-write the links? But that's getting kind of complicated)

ejj:
Eww =)

I think this is getting complicated enough, let's just talk about it offline when we have a moment.  There's a lot of tradeoffs here -- worth talking it through.

kayousterhout:
@ejj I've updated this to tear out the Slate-specific code; that's now in it's own repository, and I submitted a [PR there](https://github.com/quilt/slate/pull/1) with the changes I made to Slate (all of which were previously in this PR).  Let me know what other issues you see!

docs/_Dev.md
```diff
@@ -170,3 +98,161 @@ defined in [Makefile](../Makefile). Set `REPO` to your own repository
...
 After the above setup, you're good to go - just remember to build and push your
 image first, whenever you want to run the `minion` with your latest changes.
+
+## Contributing Code
+
+We highly encourage contributions to Quilt from the Open Source community!
+Everything from fixing spelling errors to major contributions to the
+architecture is welcome.  If you'd like to contribute but don't know
+where to get started, feel free to reach out to
+[Ethan](http://people.eecs.berkeley.edu/~ejj/contact.html) for some guidance.
```
ejj: Maybe change this to a link to the contact page, and say "us" instead of "Ethan"

docs/_Dev.md
```diff
@@ -170,3 +98,161 @@ defined in [Makefile](../Makefile). Set `REPO` to your own repository
...
+
+Commits should have proper author attribution, with the full name of the commit
+author, capitalized properly, with their email at the time of authorship.
+Commits authored by more than one person should have a `Co-Authored-By:` tag at
+the end of the commit message.
+
+### Submitting Patches
+Patches are submitted for inclusion in Quilt using a Github Pull Request or
+(though not preferred), they may be sent directly to
+[Ethan](http://people.eecs.berkeley.edu/~ejj/contact.html) using
```
ejj: I'd ditch this git too.  Let's just say you have to use a Pull Request for now.

docs/_Dev.md
```diff
@@ -170,3 +98,161 @@ defined in [Makefile](../Makefile). Set `REPO` to your own repository
...
+`git-format-patch` and `git-send-email`.
+
+A pull request is a collection of well formed commits that tie together
+in some theme, usually the larger goal they're trying to achieve.  Completely
+unrelated patches should be included in separate pull requests.
+
+Pull requests go through a two stage review process.  In the first stage,
```
ejj: I don't think this is correct any longer.

docs/_Dev.md
```diff
@@ -170,3 +98,161 @@ defined in [Makefile](../Makefile). Set `REPO` to your own repository
...
+multiple changes to the code.  After each round of review, the original author
+is expected to update the pull request with appropriate changes.  These changes
+should be incorporated into the patches in their most logical places.  I.E.
+they should be folded into the original patches or, if appropriate inserted as
+a new patch in the series.  Changes _should not_ be simply tacked on to the end
+of the series as tweaks to be squashed in later -- at all stages the PRs should
+be ready to merge without reorganizing commits.
+
+## The Quilt Daemon
+Two processes need to be running for Stitches to be enforced:  `quilt daemon` and
```
ejj: Some of the language in this paragraph is out of date btw.  I.E. Stitches->Blueprints

docs/_Dev.md
```diff
@@ -170,3 +98,161 @@ defined in [Makefile](../Makefile). Set `REPO` to your own repository
...
+`quilt run`. `quilt daemon` does the heavy lifting -- it's responsible for enforcing
+Stitches.  `quilt run` is responsible for compiling Stitches and sending them to
+the daemon to be enforced.
+
+## Code Structure
```
ejj: I wonder if we really should include this section at all ...  It's bound to get out of date, and none of our users will care.  What do you think?

If you want to keep it, we should do a pass over it to make sure it is still correct.

docs/_GettingStarted.md
```diff
@@ -37,10 +37,12 @@ successful, then the `quilt` command should execute successfully in your shell.
 
 ## Configure A Cloud Provider
 
-Below we discuss how to setup Quilt for Amazon EC2. Google Compute Engine is
-also supported. Since Quilt deploys systems consistently across providers, the
-details of the rest of this document will apply no matter what provider you
-choose.
+Quilt currently supports Amazon EC2, Digital Ocean, and Google Compute
+Engine; support for running locally with Vagrant is coming soon. Refer to
```
ejj: I would say Vagrant is "experimental" ... it should work

docs/_GettingStarted.md
```diff
@@ -54,6 +56,91 @@ aws_access_key_id = <YOUR_ID>
...
+deployment.deploy(new Machine({
+  provider: "DigitalOcean",
+  region: "sfo1",
+  size: "512mb",
+  floatingIp: "8.8.8.8",
+  role: "Worker"
+}));
+```
+
+Creating a floating IP is slightly unintuitive. Unless there are already
```
ejj: Btw, a lot of these comments I'm making, were in the original code, and could be left for a future PR.  But anyways, another issue here ... this paragraph is very confusing.  Specifically the bit about a datacenter I don't really understand.

docs/index.html.md
```diff
@@ -13,3 +13,18 @@ includes:
 search: true
 ---
 
+# Quilt
+
+Quilt is a simple way to use JavaScript to deploy and manage software
```
ejj: s/simple/approachable

kayousterhout:
@ejj I updated to this to address all but two of your comments.  Did you see my comment at the top that there's another big PR coming with a re-write of the docs, so this one just just focused on converting everything to Slate and fixing major issues?  I fixed most of your comments except for two (the two I didn't thumbs-up), which I think should be addressed as part of the major docs update.  I'd like to do that holistically rather than doing spot-fixes, because there are some major structural issues that need to be fixed, and I think it's more efficient to do that all at once (also pragmatically, I've already done part of that).  Let me know if you'd like either of those fixed as part of this PR (otherwise I've added them to my TODOs for the bigger re-write).

ejj:
Yep makes sense re delaying the fixes for a more comprehensive rewrite. Ready for me to merge then?

Ethan (Phone)

kayousterhout:
Just pushed one last commit for simple -> approachable (just seeing that
comment)! Now ready to merge.

PR #920, opened 2017-06-23
Fix Vagrant inbound connections

kklin:
This PR fixes inbound connections on Vagrant by allowing the Vagrant provider to override the interface that NAT rules should be applied to. While in the cluster code, I cleaned up various things about the `machine.Machine` struct to make it cleaner.


cluster/amazon/amazon.go
```diff
@@ -10,7 +10,6 @@ import (
 	"github.com/quilt/quilt/cluster/acl"
```
ejj: In the commit message, did you mean "these attribute are unknown by the provider"

cluster/cluster.go
```diff
@@ -217,13 +217,21 @@ func (clst cluster) updateCloud(machines []machine.Machine, act action) {
 	}
 }
 
+type joinMachine struct {
```
ejj: I think I'd just call it `machine`

cluster/amazon/amazon.go
kklin: Thanks, I meant "These attributes are known by the cluster"

cluster/cluster.go
kklin: That conflicts with the package `github.com/quilt/quilt/cluster/machine` unfortunately.

PR #923, opened 2017-06-24
vagrant: Refuse to boot preemptible machines

kklin:
Preemptible machines are not defined on Vagrant, so this commit modifies
the Vagrant provider to throw an error when requested to boot them.
Otherwise, the cluster would continuously create and destroy machines
because Vagrant's `List()` would return machines with the `preemptible`
flag set to false, thus causing the cluster join to not match them.


cluster/vagrant/vagrant.go
```diff
@@ -26,6 +26,13 @@ func New(namespace string) (*Cluster, error) {
 
 // Boot creates instances in the `clst` configured according to the `bootSet`.
 func (clst Cluster) Boot(bootSet []machine.Machine) error {
+	for _, m := range bootSet {
+		if m.Preemptible {
+			return errors.New(
+				"preemptible instances are not defined on Vagrant")
```
ejj: I don't like this error message much, sounds confusing.  Perhaps just "Vagrant does not support preemptible instances"

kklin: Done

PR #926, opened 2017-06-25
google: Create only one network per namespace

kklin:
This PR makes it so we only create one network per Google namespace, and also does some small cleanups to the Google provider.


ejj:
ooooh excited for this!

cluster/google/google.go
```diff
@@ -565,7 +605,7 @@ func (clst *Cluster) createNetwork() error {
 
 	log.Debug("Creating network")
 	op, err := clst.gce.InsertNetwork(&compute.Network{
-		Name:      clst.ns,
+		Name:      clst.networkName,
```
luise: Shouldn't the argument to `networkExists` (line 596)  also change to `clst.networkName`?

cluster/google/google.go
```diff
@@ -295,17 +295,50 @@ func (clst *Cluster) instanceNew(name string, size string,
...
+		fwApplies := false
+		for _, tag := range fw.TargetTags {
+			if tag == clst.zone {
+				fwApplies = true
+				break
+			}
+		}
+
+		if fwApplies {
+			fws = append(fws, *fw)
```
kklin: No reason, just didn't think of it! Good catch!

cluster/google/google.go
kklin: Good catch, thanks! (The reason why it worked is that we happen to make the network name the namespace)

PR #927, opened 2017-06-26
Various small fixes in preparation for the counter PR.

ejj:


quiltctl/ssh/resize/resize.go
```diff
@@ -0,0 +1,15 @@
+// +build !windows
```
kklin: I thought we aren't targeting windows anymore?

minion/docker/docker.go
```diff
@@ -77,7 +77,6 @@ type RunOptions struct {
 type client interface {
 	StartContainer(id string, hostConfig *dkc.HostConfig) error
 	UploadToContainer(id string, opts dkc.UploadToContainerOptions) error
-	DownloadFromContainer(id string, opts dkc.DownloadFromContainerOptions) error
```
kklin: Is there an implementation to delete as well?

quiltctl/ssh/resize/resize.go
ejj: You know, I wrote this patch a long time ago, and the problem it was intended to fix, I don't seem to be having anymore.  I'll likely just drop the patch.

kklin:
LGTM assuming travis passes

PR #930, opened 2017-06-27
quilt: Use exponential backoff waiting

kklin:


cluster/google/google.go
```diff
@@ -167,86 +168,53 @@ func (clst *Cluster) Stop(machines []machine.Machine) error {
...
-			exists := false
-			for _, ist := range instances {
-				if name == ist.ID {
-					exists = true
-				}
-			}
-			if live == exists {
-				names = append(names[:0], names[1:]...)
+		for _, id := range ids {
+			if _, live := liveMachines[id]; live != shouldLive {
```
Baisang: I think this part isn't super clear since you're comparing a `bool` to `struct{}{}`. Maybe have a comment explaining what's going on? From my understanding the `id` is in the map only if the machine was in the `machines` array, and if it `shouldLive` that's okay but if it shouldn't then we return false. Is that right?

ejj:
Does anything outside of the cluster use the waitfor stuff?  I might pull it in there and just hardcode the values.  That will prevent people from doing crazy things like a 500 milisecond initial check time.

Alternatively, you could make a cluster.WaitFor that just calls util.WaitFor with default values.  Just an idea.

cluster/google/google.go
kklin: It's actually comparing a `bool` to a `bool`. `live` represents whether `id` was in the `liveMachines` map.

Does that clear things up? Or do you think a comment in the code would still be helpful?

Baisang: Ohhh, I see. Yeah I think a comment would be helpful here, I was a bit thrown off since I'm used to seeing
``` go
_, ok := someMap[key]
```

But I tend to prefer more comments than other people when I code, so that might be it.

kklin:
quilt-tester uses the WaitFor stuff, and there's other places in the project that probably should use it, but are still using a hand-rolled version. I created a cluster implementation with reasonable defaults that calls the more general version.

PR #934, opened 2017-07-03
Various Cleanups

ejj:
This PR does three more-or-less unrelated things that are all necessary for the upcoming counters series.

* It updates to tool chain and re-runs `make generate`.  Note that mockery 1.0 does not play nicely with go lint.  The problem is fixed by  this PR https://github.com/vektra/mockery/pull/167, but unfortunately until that's merged we'll need to make sure we're cloning the updated version locally before re-running.

* The PR refactors the Amazon client a bit.

* The PR changes api/client to use mockery instead of the manually implemented solution we had.


cluster/amazon/client/client.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return resp.SecurityGroups, err
+}
+
+func (ac awsClient) CreateSecurityGroup(name, description string) (string, error) {
+	csgResp, err := ac.client.CreateSecurityGroup(&ec2.CreateSecurityGroupInput{
+		GroupName:   &name,
+		Description: aws.String("Quilt Group")})
```
luise: Seems like you can get rid of the `description` argument.

cluster/amazon/client/client.go
```diff
@@ -0,0 +1,173 @@
...
+	_, err := ac.client.DisassociateAddress(&ec2.DisassociateAddressInput{
+		AssociationId: &associationID})
+	return err
+}
+
+func (ac awsClient) DescribeVolumes(id string) ([]*ec2.Volume, error) {
+	resp, err := ac.client.DescribeVolumes(&ec2.DescribeVolumesInput{
+		Filters: []*ec2.Filter{{
+			Name:   aws.String("volume-id"),
+			Values: []*string{&id}}}})
```
luise: `stringSlice`?

cluster/amazon/client/client.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return resp.Volumes, err
+}
+
+// New creates a new Client.
+func New(region string) Client {
+	session := session.New()
+	session.Config.Region = &region
```
luise: `aws.String`

cluster/amazon/client/client_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func TestErrors(t *testing.T) {
+	ac := New("junk")
+
+	// Disable HTTP requesting for unit tests
+	ac.(awsClient).client.Client.Handlers.Clear()
+	ac.(awsClient).client.Client.Handlers.Send.PushBack(func(r *request.Request) {
+		r.Error = errors.New("test")
+	})
+
```
luise: Is there not a mock for this? Or is it possible to generate one? Seems like these tests don't add much value (other than a slightly happier Cov report :) )

minion/ovsdb/ovsdb_test.go
```diff
@@ -155,12 +155,13 @@ func TestCreateSwitchPort(t *testing.T) {
 			newMutation("ports", "insert", ovs.UUID{GoUUID: "qlsportadd"}),
 		},
 		Where: newCondition("name", "==", "lswitch")}}
-	api.On("Transact", "OVN_Northbound", ops).Return(nil, errors.New("err")).Once()
+	api.On("Transact", "OVN_Northbound", ops[0], ops[1]).Return(
```
ejj: Mockery updated and the new version doesn't support the old syntax (it panics)

cluster/amazon/client/client_test.go
ejj: Believe it or not, this is actually the recommended way to test the aws API.

I don't think they add a lot of value, but the code their testing doesn't have a lot of logic in it anyways.  Really all we need to verify is that the underlying aws API is actually called, which is precisely what these tests do.  I think its worth the coverage bump

minion/ovsdb/ovsdb_test.go
```diff
@@ -532,10 +537,10 @@ func TestListAddressSets(t *testing.T) {
 	api := new(mockTransact)
 	odb := Client(client{api})
 
-	ops := []ovs.Operation{{
+	ops := ovs.Operation{
```
ejj: Done

cluster/amazon/client/client.go
```diff
@@ -0,0 +1,173 @@
...
+			Filters:                filters})
+	return resp.SpotInstanceRequests, err
+}
+
+func (ac awsClient) RequestSpotInstances(spotPrice string, count int64,
+	launchSpec *ec2.RequestSpotLaunchSpecification) (
+	[]*ec2.SpotInstanceRequest, error) {
+
+	resp, err := ac.client.RequestSpotInstances(&ec2.RequestSpotInstancesInput{
+		SpotPrice:           &spotPrice,
```
ejj: Meh, I think the aws.String thing is stupid.  It's convenient occaisionally, but it's super weird and not all that useful.

cluster/amazon/client/client.go
ejj: Good catch

cluster/amazon/client/client.go
```diff
@@ -0,0 +1,173 @@
...
+// New creates a new Client.
+func New(region string) Client {
+	session := session.New()
+	session.Config.Region = &region
+	return awsClient{ec2.New(session)}
+}
+
+func stringSlice(slice []string) []*string {
+	if slice == nil {
+		return nil
```
ejj: Actually this is a bit confusing.  So the amazon API makes a distinction between nil which means "this parameter was omitted" and an empty slice which means "this parameter was provided with no elements".  If you use aws.StringSlice directly, or make the change you suggest, then `nil` will be changed to []*string{} which will actually cause the code to stop working.  The reason that CancelSpotInstanceRequestsInput requires it to be non-nil, is because semantically you can't cancel a spot request without passing in an ID.  I.E nil is semantically wrong.

In short, the point of string slice is to preserve the semantics of the argument, unlike aws.StringSlice which clobbers it.

ejj: Btw, I added a comment clarifying.

cluster/amazon/amazon.go
```diff
@@ -311,57 +297,33 @@ func (clst *Cluster) parseDiskSize(inst ec2.Instance) (int, error) {
...
-	}
-
-	volumeInfo, err := clst.client.DescribeVolumes(
-		&ec2.DescribeVolumesInput{
-			Filters: filters,
-		})
-	if err != nil {
+	volumeID := *inst.BlockDeviceMappings[0].Ebs.VolumeId
+	volumes, err := clst.client.DescribeVolumes(volumeID)
+	if err != nil || len(volumes) == 0 {
```
ejj: Sure, that's probably true, but I don't think this is the code to check it.  If someone upstream depends on the volumes being zero, it should enforce that.  The code that checks what the volume size is, should just return whatever the answer is.  At any rate, I could see a problem if, for example, someone deleted the volume through the Amazon API -- it may cause is to delete the VM and reboot it.  That said, I'm not sure how bad a problem that is.  I'm also definitely sure, this isn't the patch to change it.

ejj:
@luise, pushed a new version in case you want to look over the changes.

cluster/amazon/amazon.go
luise: Ok. I'm only asking because before your change, it would not error. If it errors now, it would cause `List()` to log a warning which might be undesired for cosmetic reasons.

cluster/amazon/client/client.go
luise: Hmm, ok. I just find it weirdly inconsistent that we sometimes use `aws.String` and sometimes just `&`.

PR #935, opened 2017-07-04
Pull interface mocks out of the production binaries

ejj:
For several packages we were including the mocks used for unit testing in with the production binary.  This caused them to include the httptest package which added a nonsense flag to our `quilt --help` output.


ejj:
Well we want everyone the interface instead of the struct directly so we can mock it out. Thus we keep the struct private and return be interface type. 
		

		
	_____________________________

kklin:
@ejj do you think this fix deserves to be in the Changelog?

PR #938, opened 2017-07-06
ovs: Bundle newest DigitalOcean kernel version

kklin:
The Ubuntu image provided by DigitalOcean now uses the linux kernel
version 4.4.0-81.


kklin:
The auto-build worked, I just noticed the machines took longer than usual to connect (6 minutes).

Baisang:
I think we've gone past the vagrant kernel version we've pinned a while ago. Would we want to update either?

Baisang:
I.e our ovs docker file version > the version in the pinned vagrant box

kklin:
I think the Vagrant version is pinned to `4.4.0-62` (as is Amazon and Google).

Baisang:
Oh I think you're right, I remember now. I had specifically found that version to make sure it was the same as the Amazon and Google ones.

PR #940, opened 2017-07-07
Update all dependencies except for vishvananda/netlink

kklin:


ejj:
Why not netlink?

kklin:
@ejj I included a note in the commit message: "The latest version vishvananda/netlink currently has an error where it fails to compile on Mac."

ejj:
Ah, we should submit a PR.  I had to fix up several such errors to get it working in the first place.

PR #941, opened 2017-07-07
Usage string updates and cleanups

luise:
This PR tweaks usage strings, word-wraps and standardizes them, and adds `-v` and `-verbose` flags for debug logging.


ejj:
Overall comment, 1) I think you should just use backticks instead of of trying to construct these huge usage strings using concatenation.  2)  I think we should wordwrap at 80-ish characters.  It's much easier to read in a terminal that way.

luise:
@ejj , did you see the word wrapping in 32089e7? I wanted to get rid of the manual wrapping with the back ticks. Do you prefer the manual way? I could make the strings in the code more readable some other way.

kklin:
I prefer backticks. It's easy to miss newlines or spaces when concatenating strings in my experience.

luise:
Ok, sounds good, I'll do the wrapping with back ticks instead.

ejj:
The other nice thing about backticks, is it's easy to have a sense of what the output will look like just by looking at the code.

luise:
@ejj , updated to use back ticks

quilt.go
```diff
@@ -21,22 +21,22 @@ func main() {
...
 			"for deploying it as specified.  Alternatively, quilt may be\n" +
-			"instructed to stop all deployments in a given namespace,\n" +
-			"or the default namespace if none is provided.\n")
+			"instructed to stop all deployments or containers in a given\n" +
+			"namespace, or in the default namespace if none is provided.\n")
 		flag.PrintDefaults()
-		fmt.Println("        Valid logger levels are:\n" +
-			"            debug, info, warn, error, fatal or panic.")
 	}
+	var logLevelInfo = "level to set logger to (debug, info, warn, error, " +
```
ejj: This is very inelegantly worded.

quiltctl/command/ssh.go
```diff
@@ -33,18 +34,15 @@ func NewSSHCommand() *SSH {
 	return &SSH{sshGetter: ssh.New}
 }
 
-var sshUsage = `usage: quilt ssh <id> [command]
-
-Create a SSH session with the specified id.
-Either a container or machine ID can be supplied.
-If no command is supplied, a login shell is created.
+var sshCommands = "quilt ssh <id> [command]"
+var sshExplanation = `Create a SSH session with the specified id. Either a container or
```
ejj: "an SSH"

quiltctl/command/stop.go
```diff
@@ -22,25 +22,24 @@ func NewStopCommand() *Stop {
 	return &Stop{}
 }
 
+var stopCommands = `quilt stop [-H=<daemon_host>] [-containers] [-namespace=<namespace> |
+<namespace>]`
+
+var stopExplanation = "`stop` " + `creates an empty Stitch for the given namespace, and
```
ejj: FWIW, I'm not necessarily a big fan of backticks in usage strings -- it's a bit unusual.  That said, I can't think of a reason it should be prohibited.

util/util.go
```diff
@@ -204,3 +205,13 @@ func WaitFor(pred func() bool, interval time.Duration, timeout time.Duration) er
 		Sleep(interval)
 	}
 }
+
+// PrintUsageString formats and prints usage strings for quiltctl commands.
+func PrintUsageString(commands string, explanation string, flags *flag.FlagSet) {
+	fmt.Println("Usage: " + commands + "\n\n" + explanation + "\n")
+	if flags == nil {
```
ejj: This is minor, but I think this if statement would be easier to read if you said `if flags != nil`  then it's clear you're saying in that case, we'll use the provided variable.

quiltctl/command/stop.go
luise: Ok, I'll just leave them for now

quiltctl/command/ssh.go
luise: Nice catch - I just copied what was there.

PR #944, opened 2017-07-10
ip: Don't allocate IPs in subnets governed by non-Quilt routes

kklin:
Before, cloud providers would sometimes assign machines a private IP in
    a /16 within the 10.0.0.0/8 subnet. When this happened, if a container
    was allocated an IP within that subnet, traffic destined to that
    container from the DNS server, or from the public internet, would not
    get properly routed to OVN via quilt-int. Instead, it would follow the
    route specific by the cloud provider.

    This patch fixes this by having minions report the subnets on the host
    network stack. The master then refuses to assign IPs within those
    subnets. If an entity already has an IP within a blacklisted
    subnet, the entity gets reassigned a new IP.


minion/network/ip.go
```diff
@@ -14,13 +15,27 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
+/*
```
ejj: This is a total nit, but I don't really like these being on their own lines. The rest of the code combines it with the first line of text.

minion/network/ip.go
```diff
@@ -68,19 +91,70 @@ func makeIPContext(view db.Database) ipContext {
...
+	subnetBlacklist, err := makeSubnetBlacklist(view)
+	if err != nil {
+		return fmt.Errorf("make subnet blacklist: %s", err)
 	}
 
-	err := allocateContainerIPs(view, ctx)
-	if err == nil {
-		syncLabelContainerIPs(view)
-		err = allocateLabelIPs(view, ctx)
+	// Attempt to allocate IPs up to three times in case we happen to choose
```
ejj: This feels a bit weird.  Why not push this down into the underlying IP allocation code?

ejj: I.E. have allocateIP take a blacklist?

ejj: I.E. have the `ipSet` take subnet instead of an IP, most of which will be /32s, but some of which will be larger ranges.

minion/network/ip.go
```diff
@@ -68,19 +91,70 @@ func makeIPContext(view db.Database) ipContext {
...
+			err = allocateLabelIPs(view, ctx)
+		}
 	}
 	return err
 }
 
+// makeSubnetBlacklist returns all subnets that are governed by routes in a
+// worker machine's network stack, and intersect with the Quilt container
+// subnet.
+func makeSubnetBlacklist(view db.Database) ([]net.IPNet, error) {
```
ejj: Sigh, this is disappointing.  Nothing we can do about it though.

minion/network/ip.go
kklin: My main reason was that I wanted to make sure that if a minion appeared with a subnet that an IP had already been allocated in, we'd unassign that IP. We could still pass the blacklist down to allocateIP though.

Also, `allocateIP` would be `O(n)` in number of containers to allocate an IP, does that seem OK?

minion/network/ip.go
```diff
@@ -14,58 +15,150 @@ import (
...
+	for i := 0; i < 3; i++ {
+		ctx := makeIPContext(view, subnetBlacklist)
+		if len(ctx.unassignedContainers) == 0 && len(ctx.unassignedLabels) == 0 {
+			return nil
+		}
+
+		err = allocateContainerIPs(view, ctx)
+		if err == nil {
+			syncLabelContainerIPs(view)
+			err = allocateLabelIPs(view, ctx)
```
Baisang: If this succeeds, do we need to keep iterating?

kklin: If the assigned IP happens to fall in a blacklisted subnet, we need to reallocate IPs again.. it's kind of weird; @ejj mentioned in his review too..

PR #945, opened 2017-07-11
docs: Move Make target to docs folder

kayousterhout:
This commit moves the docs make target from the main
Makefile (so a user was expected to run "make docs" from the
main directory) to the docs directory (so now the user just runs
"make" from the docs directory).

FYI @kklin -- this definitely makes more sense (it also removes the
need to have go / govendor installed in order to build the docs)


kklin:
Will merge once travis passes

PR #946, opened 2017-07-11
foreman: Use minion config to determine role

kklin:
Before, the code that generated the etcd IPs used the database to
determine the role of a machine. However, there is a period of time when
we have not yet connected to the minion yet that the database role is
simply a guess. This resulted in a bug where booting machines with
Quilt then restarting the daemon caused the etcd IPs to be incorrect.
This is because the cluster assigned a master the IP of a worker.

This patch works around this problem by referencing the role in the
minion config. Thus, when the minion hasn't connected yet, the role will
be `None`, and the foreman will ignore it.


kklin:
@ejj The build was failing when you approved, but I've fixed it up now. Can you take another look?

PR #947, opened 2017-07-11
Various pre-TLS fixes

kklin:
My TLS branch was getting big again, so I'm pulling out these commits to make it easier to review.


cluster/cluster_test.go
```diff
@@ -79,6 +58,8 @@ func (p *fakeProvider) List() ([]machine.Machine, error) {
 
 func (p *fakeProvider) Boot(bootSet []machine.Machine) error {
 	for _, toBoot := range bootSet {
+		p.bootRequests = append(p.bootRequests, toBoot)
```
kayousterhout: Why do this at the beginning? It's slightly more intuitive to have at the end, after you've updated the properties of toBoot

cluster/cloudcfg/cloudcfg.go
```diff
@@ -67,9 +68,17 @@ func (opts MinionOptions) String() string {
 		"outbound-pub-intf": opts.OutboundPubIntf,
 	}
 
+	// Sort the option keys so that the command line arguments are consistently
```
kayousterhout: Why not do the sorting in the unit test? Seems slightly better so the functionality (sorting to get consistent order) is closer to where it's used.  If you don't want to do that, can you add "(useful for unit testing)" here, so it's obvious why this is happening.

api/client/leader.go
```diff
@@ -3,16 +3,16 @@ package client
...
 	"github.com/quilt/quilt/api"
 	"github.com/quilt/quilt/api/util"
 	"github.com/quilt/quilt/db"
-
-	log "github.com/Sirupsen/logrus"
 )
 
 // Leader obtains a Client connected to the Leader of the cluster.
 func Leader(machines []db.Machine) (Client, error) {
+	getLeaderErrors := map[string]error{}
```
kayousterhout: Can you add a comment about what the key is here ("// Maps IP addresses to the error that resulted from asking that machine for the lead minion's IP")? (Or call it minionIPToError?)

api/client/leader.go
```diff
@@ -23,10 +23,20 @@ func Leader(machines []db.Machine) (Client, error) {
...
+		getLeaderErrors[m.PublicIP] = err
+	}
+
+	var errorStrs []string
+	for m, err := range getLeaderErrors {
+		errorStrs = append(errorStrs, fmt.Sprintf("%s - %s", m, err.Error()))
+	}
+
+	err := "no leader found"
+	if len(errorStrs) != 0 {
```
kayousterhout: would it make sense to add an "else" here that adds something like ": no minions were available to ask"?

api/client/leader.go
```diff
@@ -23,10 +23,20 @@ func Leader(machines []db.Machine) (Client, error) {
...
+		err += ": " + strings.Join(errorStrs, ", ")
```
kayousterhout: super nit but I think it's better to use semi-colons rather than commas for things like this, because a comma is more likely to be in the error string

cluster/cluster_test.go
kklin: I actually didn't want the `bootRequests` to have the fields that the cloud provider populates (e.g. the user doesn't choose the cloud ID that the machine will have). I'll add a comment to make that clear.

cluster/cloudcfg/cloudcfg.go
kklin: By the time it reaches the unit test, the key pairs have been concatenated into one large string, so it seemed hacky to break them back up. I'll ad a comment.

api/client/leader.go
kklin: I added a commit that achieves the same thing by checking if `len(machines) == 0`

PR #949, opened 2017-07-12
stitch: Let blueprints write to stdout

kklin:


quilt-tester/tests/console-log/check_console_log.go
```diff
@@ -0,0 +1,15 @@
+package main
+
+import (
+	"fmt"
+)
+
+func main() {
+	// XXX: Ideally, we would check that the expected output was actually
```
kayousterhout: Can you add a reminder here that explicitly says that console-log.js will be run first?

quilt-tester/tests/console-log/console-log.js
```diff
@@ -0,0 +1,11 @@
+const {createDeployment, Service, Container} = require("@quilt/quilt");
```
kayousterhout: nit: alphabetize

stitch/stitch.go
```diff
@@ -117,24 +117,33 @@ func FromFile(filename string) (Stitch, error) {
 			"failed to locate Node.js. Is it installed and in your PATH?")
 	}
 
-	stderr := bytes.NewBuffer(nil)
-	cmd := exec.Command("node", "-p",
+	outFile, err := ioutil.TempFile("", "quilt-out")
+	if err != nil {
+		return Stitch{}, fmt.Errorf("create deployment file: %s", err)
```
kayousterhout: Can you change this to "failed to create deployment file: %s"

stitch/stitch.go
```diff
@@ -117,24 +117,33 @@ func FromFile(filename string) (Stitch, error) {
...
+		return Stitch{}, err
 	}
-	// If there wasn't an error, still print stderr, in case there were any
-	// warnings or other non-fatal errors.
-	fmt.Fprint(os.Stderr, stderr.String())
 
-	return FromJSON(string(out))
+	depl, err := ioutil.ReadAll(outFile)
+	if err != nil {
+		return Stitch{}, fmt.Errorf("read deployment file: %s", err)
```
kayousterhout: As above can this be "failed to read..."?

stitch/stitch.go
kklin: I'm not sure I agree, it seems that it being an error implies that it failed. I like keeping the pretext as concise as possible because they're often chained, although I don't know if this is a general practice.

kayousterhout: What's the case where there's an error but it didn't fail?

Without "failed to.." I'm worried a user might mis-interpret what the cause of the failure was (e.g., one mis-reading of this is to read it as a directive: "this failed because you need to create a deployment file")

stitch/stitch.go
```diff
@@ -117,24 +117,33 @@ func FromFile(filename string) (Stitch, error) {
...
 	}
 
-	stderr := bytes.NewBuffer(nil)
-	cmd := exec.Command("node", "-p",
+	outFile, err := ioutil.TempFile("", "quilt-out")
+	if err != nil {
+		return Stitch{}, fmt.Errorf("create deployment file: %s", err)
+	}
+	defer outFile.Close()
+	defer os.Remove(outFile.Name())
```
Baisang: Is there a reason we choose to write to a random file and deleting vs. appending to a log?

At least with the latter users can see their `console.log`s somewhere, unless we are trying to discourage this.

stitch/stitch.go
kklin: Fair enough, I've added the "failed to" in both locations.

kklin:
@kayousterhout addressed your comments, and modified the integration test to use the new testing structure.

quilt-tester/tests/console-log/console-log.js
```diff
@@ -0,0 +1,11 @@
+const {Container, createDeployment, Service} = require('@quilt/quilt');
```
kayousterhout: to be correctly alphabetized, this should be Container, Service, createDeployment (caps come first)

PR #950, opened 2017-07-12
quiltctl: Allow querying minion counters

kklin:


quiltctl/command/counter.go
```diff
@@ -1,24 +1,33 @@
...
 	"text/tabwriter"
 
+	"github.com/quilt/quilt/api/client"
 	"github.com/quilt/quilt/api/pb"
 )
 
-var usage = `usage: quilt counters [-H=<daemon_host>]
+const daemonTarget = "daemon"
+
+var usage = fmt.Sprintf(`usage: quilt counters [-H=<daemon_host>] <target>
```
ejj: Why not make `target` optional and have it point at the daemon if omitted?

kklin: The daemon didn't feel "special" enough that it should  be the default argument -- I'd prefer that users have to explicitly say what entity they want the counters for. And if they don't supply any, treat it as a sign that they're not sure how to use the command, and complain loudly with an error.

Don't feel strongly about it though.

PR #952, opened 2017-07-13
lint: Don't lint docs/build

kayousterhout:
The docs/build directory contains outside code that we clone,
so it shouldn't be linted.


kayousterhout:
Wow quilt-bot is really worried about the quality of this PR

PR #953, opened 2017-07-13
Lint JavaScript code

kayousterhout:


stitch/bindings.js
```diff
@@ -137,24 +134,18 @@ Deployment.prototype.toQuiltRepresentation = function() {
...
 
         services.push({
             name: service.name,
             ids: ids,
-            annotations: service.annotations
+            annotations: service.annotations,
         });
     });
 
-    var containers = [];
```
kayousterhout: @kklin I was slightly worried about this change -- this is the only change here that looks potentially like a functional change.  It looks OK to me but would be great if you could double check (one of your StitchID patches introduced the duplicate containers variable).

Makefile
```diff
@@ -116,6 +116,7 @@ lint: format
 	find . \( -path ./vendor -o -path ./node_modules \) -prune -o -name '*' -type f -print | xargs misspell -error || EXIT_CODE=1; \
 	ineffassign . || EXIT_CODE=1; \
 	exit $$EXIT_CODE
+	node_modules/eslint/bin/eslint.js -c stitch/eslint.conf stitch/ quilt-tester/ --fix
```
kayousterhout: One thing that's slightly weird here is using the "--fix" flag -- I'm not sure if that should go in format instead? It's weird since linting and fixing the format (when possible -- there are some things like writing JSDoc that the linter can't do automatically) are coupled.

stitch/bindings.js
```diff
@@ -1,3 +1,4 @@
+/* eslint require-jsdoc: [1] */
```
ejj: What does this mean?

kayousterhout: Related to your other comment -- this means that for this file, we're just warned about lack of JSDoc (otherwise the linter prints an error).  I did that because I didn't want to update all of the JSDoc here.  Do you want to make it a warning everywhere?  I'd be in favor of keeping it as an error (and working to update this file to have JSDoc everywhere) because I think it will substantially help readability (and then we can eventually generate the blueprint API documentation based only on the JSDoc)

kayousterhout:
Also be sure not to merge until Kevin (or you? Did you see it?) checkout the comment I left in bindings.js

ejj:
Yeah I saw the comment.  I don't personally know the answer.  I will let @kklin merge when he's looked at it.

stitch/bindings.js
kklin: Eh, this will break if a container is referenced by multiple services I think. The `containers` variable defined above will contain duplicate containers. Not completely sure if that'll confuse Quilt, but I wouldn't be surprised if it did.

This isn't my best code, sorry for the confusion 😄

kayousterhout:
I updated this to rebase on top of my recent change to fix the containers issue, so this should be a ll set

PR #956, opened 2017-07-13
js: Fix reference to invariant nodes

kklin:
This should fix the build, which is failing because the Spark blueprint uses invariants.

Note that I would normally add unit tests to cover invariants, but I didn't because I'm planning on ripping out that code in the near future anyways (for exactly this reason -- it's not used, and it's overhead to maintain).


kklin:
Note that I'd expect the build to still fail when testing the Spark blueprint before this is merged because that test references the version of `bindings.js` in `master`.

kklin:
No worries @kayousterhout, it's really a failure on our unit tests that it didn't catch it

PR #957, opened 2017-07-13
Support TLS-encrypted communication between Quilt clients and servers

kklin:


kklin:
@kayousterhout I requested your review because your comments last time were super helpful, and you read the previous version of this code

connection/credentials/cli/files.go
```diff
@@ -0,0 +1,22 @@
+package cli
```
kayousterhout: Why is this package called cli / why isn't this all in the tls dir?

connection/credentials/tls/rsa/rsa.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+// KeyPair represents an RSA private key and certificate. The keys and
+// certificates it generates are compatible with our TLS scheme.
+type KeyPair struct {
+	key  *rsa.PrivateKey
+	cert *x509.Certificate
+}
+
+// KeyString returns the PEM-encoded string representing the private key.
```
kayousterhout: can you say more here about the significance of the PEM-encoding?

connection/credentials/tls/rsa/rsa.go
```diff
@@ -0,0 +1,153 @@
...
+	cert, err := x509.ParseCertificate(certDER)
+	if err != nil {
+		return KeyPair{}, fmt.Errorf("parse cert: %s", err)
+	}
+
+	return KeyPair{key, cert}, nil
+}
+
+// getDER decodes a PEM-encoded file from disk.
+func getDER(path string) ([]byte, error) {
```
kayousterhout: What does DER stand for? (can you use the longer name here?)

quiltctl/command/daemon.go
```diff
@@ -76,10 +79,37 @@ func (dCmd *Daemon) Run() int {
...
+	if _, isTLS := creds.(tls.TLS); isTLS {
+		minionTLSDir = "/home/quilt/.quilt/tls"
+		if sshKey == nil {
+			log.Error("A private key must be supplied to " +
+				"distribute TLS certificates")
+			return 1
+		}
+
+		ca, err := rsa.FromFile(
+			filepath.Join(dCmd.tlsDir, cli.CA.Cert()),
```
kayousterhout: can cli.CA.Cert() / cli.CA.Key() be hidden from this code, so this code just calls rsa.FromFile(dCmd.tlsDir)?

connection/credentials/tls/rsa/rsa.go
```diff
@@ -0,0 +1,153 @@
...
+	"fmt"
+	"math/big"
+	"net"
+	"time"
+
+	"github.com/quilt/quilt/util"
+)
+
+// KeyPair represents an RSA private key and certificate. The keys and
+// certificates it generates are compatible with our TLS scheme.
```
kayousterhout: Can you write a longer comment describing somewhere what our TLS scheme is? I'm still confused about what everything is (e.g., what's the significance of the RSA stuff / what's the division of functionality between the RSA package and the rest of the TLS code? Is the stuff in rsa all of the logic around keys whereas the TLS is more generally everything about transport layer security -- and if so, could rsa be renamed keys? or keypair or something like that?)

connection/credentials/tls/tls.go
```diff
@@ -0,0 +1,69 @@
...
+	"crypto/x509"
+	"errors"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+
+	"github.com/quilt/quilt/util"
+)
+
+// TLS configures authentication settings for TLS.
```
kayousterhout: This comment is pretty useless :) and this is a good place for the longer comment I requested above for the overall scheme for how TLS works

minion/run.go
```diff
@@ -58,13 +58,28 @@ func Run(role db.Role, inboundPubIntf, outboundPubIntf string,
...
 	supervisor.Run(conn, dk, role)
 
-	go minionServerRun(conn, creds)
 	go scheduler.Run(conn, dk)
 	go network.Run(conn, inboundPubIntf, outboundPubIntf)
 	go registry.Run(conn, dk)
 	go etcd.Run(conn)
 	go syncAuthorizedKeys(conn)
 
+	var creds connection.Credentials
```
kayousterhout: Can you add a comment here to say that this is waiting for the TLS keys to be installed (which is done by XXXX code) because the minion won't be able to connect to anything until that's done

kayousterhout: Also this code makes me nervous -- let's discuss in person with Ethan tomorrow.

quiltctl/command/daemon.go
```diff
@@ -76,10 +79,37 @@ func (dCmd *Daemon) Run() int {
...
 	go engine.Run(conn, getPublicKey(sshKey))
-	go server.Run(conn, dCmd.host, true, credentials.Insecure{})
-	cluster.Run(conn, credentials.Insecure{})
+	go server.Run(conn, dCmd.host, true, creds)
+
+	var minionTLSDir string
+	if _, isTLS := creds.(tls.TLS); isTLS {
+		minionTLSDir = "/home/quilt/.quilt/tls"
+		if sshKey == nil {
+			log.Error("A private key must be supplied to " +
```
kayousterhout: can you change this to "A ssh key" (so it's obvious to the user what they need to fix)

quiltctl/command/daemon.go
```diff
@@ -76,10 +79,37 @@ func (dCmd *Daemon) Run() int {
...
+	}
+
 	conn := db.New()
 	go engine.Run(conn, getPublicKey(sshKey))
-	go server.Run(conn, dCmd.host, true, credentials.Insecure{})
-	cluster.Run(conn, credentials.Insecure{})
+	go server.Run(conn, dCmd.host, true, creds)
+
+	var minionTLSDir string
+	if _, isTLS := creds.(tls.TLS); isTLS {
```
kayousterhout: What if creds had a `installCreds()` method, and for the insecure one, it did nothing, and for the TLS creds, it did everything here?

cluster/credentials.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/spf13/afero"
+	"github.com/spf13/afero/sftpfs"
+	"golang.org/x/crypto/ssh"
+
+	"github.com/quilt/quilt/connection/credentials/cli"
+	"github.com/quilt/quilt/connection/credentials/tls/rsa"
+	"github.com/quilt/quilt/db"
+)
+
+// SyncCredentials installs TLS certificates on all machines. It generates
```
kayousterhout: Also, my instinct would be to put this whole file in the TLS package rather than the cluster package.  Thoughts?

api/server/server.go
```diff
@@ -44,15 +45,12 @@ type server struct {
...
-		return err
-	}
-
 	var sock net.Listener
+	var s *grpc.Server
 	apiServer := server{conn, runningOnDaemon}
 	for {
-		sock, err = net.Listen(proto, addr)
+		var err error
+		sock, s, err = connection.Server(listenAddr)
```
kklin: I'm not sure.. My gut feeling is that infinite loops are weird, and it's best to do them as high up as possible. E.g. when I see that there's an infinite loop here, I immediately think about whether it's in a separate goroutine, whereas if we hid it in `connection.Server` someone might put it somewhere they didn't want to block.

Don't really know though, that was just my gut reaction. I'm fine changing it if you'd like!

connection/connection.go
```diff
@@ -13,8 +13,19 @@ import (
 // The timeout for connecting to the daemon.
 const connectTimeout = 5 * time.Second
 
+// Credentials defines the credentials to use when creating a client or server.
+type Credentials interface {
+	// Client returns the `DialOption`s necessary to setup the credentials when
+	// obtaining a grpc client connection.
+	Client() []grpc.DialOption
```
kklin: Good idea, I renamed it ClientOpts (also renamed Server as well)

connection/connection.go
```diff
@@ -51,5 +52,9 @@ func Server(lAddr string, creds Credentials) (net.Listener, *grpc.Server, error)
 		return nil, nil, fmt.Errorf("listen: %s", err)
 	}
 
+	// Don't enforce TLS on local connections.
+	if proto == "unix" {
```
kklin: This was part of the design we made a few weeks ago -- by not enforcing TLS on local connections the user doesn't have to supply the `tls-dir` whenever they run other commands like `quilt ps` or `quilt run`. This is why we now proxy connections to the cluster through the daemon, so that the daemon can inject the proper TLS credentials.

This will make features like always autogenerating the TLS credentials easier because we don't have to deal with letting the other subcommands know about the generated credentials.

I could imagine that some users would want to encrypt local connections as well though, so I'll add an `XXX` comment here explaining the above.

api/server/server.go
kayousterhout: I hadn't thought about it that way re: loops being high up but I think I do generally agree with that; e.g., if there's an infinite loop that's running the whole system, I agree it's nice for that to be high up.  But this seems slightly different than that, because it's really saying "try to start the server until it succeeds", which seems fine to happen at a low level (and I think as the system grows, it will be impossible to avoid having things like this at a low level).

cluster/cloudcfg/cloudcfg.go
```diff
@@ -27,19 +27,27 @@ func Ubuntu(opts Options) string {
 
 	img := fmt.Sprintf("%s:%s", quiltImage, ver)
 
+	dockerOpts := ""
+	if opts.MinionOpts.TLSDir != "" {
+		// Mount the TLSDir as a read-only host volume.
+		dockerOpts = fmt.Sprintf("-v %[1]s:%[1]s:ro", opts.MinionOpts.TLSDir)
```
kklin: This is telling Docker to create a volume that links a directory within the minion to a directory on the host. This way, the daemon can SCP certs into `~/.quilt/tls` on the host.

kayousterhout: Ahhh I see and now the comment makes sense.  Can you expand the comment to say this?

cluster/cloudcfg/cloudcfg.go
```diff
@@ -59,13 +67,15 @@ type MinionOptions struct {
...
 	OutboundPubIntf string
+	TLSDir          string
 }
 
 func (opts MinionOptions) String() string {
 	optsMap := map[string]string{
 		"role":              string(opts.Role),
 		"inbound-pub-intf":  opts.InboundPubIntf,
 		"outbound-pub-intf": opts.OutboundPubIntf,
+		"tls-dir":           opts.TLSDir,
```
kklin: I agree that we should do this, but it'll require restructuring a lot of code because of cyclic dependency (the daemon uses the cloudcfg, so the cloudcfg can't require the minion because they're currently in the same `command` package).

I'll submit a PR that fixes this after this is merged -- I don't think it should block this PR because the same problem is there for the other options already.

cluster/credentials.go
kklin: I put this in the cluster package because as we discussed offline, it's pretty related to the foreman. Do you still think it should be moved?

cluster/credentials.go
```diff
@@ -0,0 +1,148 @@
...
+// the certificates using the given certificate authority, and copies them
+// over using the given ssh key.
+func SyncCredentials(conn db.Conn, dstDir string, sshKey ssh.Signer, ca rsa.KeyPair) {
+	initMachines := map[string]struct{}{}
+	for range conn.TriggerTick(30, db.MachineTable).C {
+		machines := conn.SelectFromMachine(nil)
+		syncCredentialsOnce(dstDir, sshKey, ca, machines, initMachines)
```
kklin: Renamed `initMachines`, pulled out the logic for generating and installing certificates into a separate function, and made a (backlog) trello card on possible creating a mock database.

cluster/credentials.go
```diff
@@ -0,0 +1,148 @@
...
+				Error("Failed to generate certs")
+			continue
+		}
+
+		fs, err := getSftpFs(m.PublicIP, sshKey)
+		if err != nil {
+			// Log at debug because failures to SSH are expected
+			// while the machine is still booting.
+			log.WithError(err).WithField("host", m.PublicIP).
+				Debug("Failed to get SFTP client")
```
kklin: Do you think that'll be misleading because it won't try again until the next iteration of the loop? (probably 30 seconds later)

cluster/credentials.go
```diff
@@ -0,0 +1,148 @@
...
+			continue
+		}
+		defer fs.Close()
+
+		err = installCerts(fs, dstDir,
+			keyPair.KeyString(), keyPair.CertString(), ca.CertString())
+		if err != nil {
+			log.WithError(err).WithField("host", m.PublicIP).
+				Error("Failed to install certs")
```
kklin: Discussed offline

cluster/cluster.go
```diff
@@ -55,6 +55,9 @@ type cluster struct {
 	namespace string
 	conn      db.Conn
 	providers map[launchLoc]provider
+
+	// The TLS directory with which minions should run.
```
kklin: I improved the comment, and changed the type to be the new `TLSDir` type. This way, the future maintainer will go to the definition of `TLSDir`, and see the description of the structure there.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ to complete. This decreases the number of cloud provider API calls done by Quilt
 - `quilt ps` now displays the image building status of custom Dockerfiles.
 - Let blueprints write to stdout. Before, if blueprints used `console.log`, the
 text printed to stdout would break the deployment object.
+- Allow an admin key access to all machines deployed by the daemon. The key is
```
kklin: named it `admin-ssh-private-key` -- not too worried about it being overly verbose because we'll autogenerate keys if this isn't supplied, so most users won't need to type it

kklin:
@kayousterhout ready for another round of review!

connection/credentials/tls/rsa/rsa.go
```diff
@@ -0,0 +1,157 @@
...
+	der, _ := pem.Decode([]byte(pemStr))
+	if der == nil {
+		return nil, errors.New("no key PEM data found")
+	}
+
+	return der.Bytes, nil
+}
+
+// NewCertificateAuthority generates a KeyPair that can be used as a
+// certificate authority. Specifically, it has a self-signed certificate that
```
kayousterhout: this comment is also awesome thanks for updating these!

connection/credentials/tls/tls.go
```diff
@@ -0,0 +1,76 @@
...
+	return []grpc.DialOption{grpc.WithTransportCredentials(
+		credentials.NewTLS(&tls.Config{
+			Certificates: []tls.Certificate{tlsAuth.keyPair},
+			RootCAs:      tlsAuth.caPool,
+		}),
+	)}
+}
+
+// FromFile creates a TLS instance from the filesystem.
+func FromFile(caPath, certPath, keyPath string) (TLS, error) {
```
kayousterhout: Similar to my rsa comment, I think the io package should read these files and then pass in the strings -- so that the io package is handling all the io and this just deals with security semantics

connection/credentials/tls/tls.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return TLS{}, err
+	}
+
+	key, err := util.ReadFile(keyPath)
+	if err != nil {
+		return TLS{}, err
+	}
+
+	keyPair, err := tls.X509KeyPair([]byte(cert), []byte(key))
```
kayousterhout: This doesn't feel like the best place for this functionality -- since it's highly related to the logic in the RSA package about how these keys get generated and written to file (my first thought here was "where's the code to write these keys and is it consistent -- which was hard to figure out).  I know this isn't rsa-specific so doesn't make sense there... would it make sense to move the code here and the associated code that's currently in rsa to a new file certificates.go in the TLS package? Or you could put it in the RSA package, give up on that being ONLY rsa (maybe rename it keys? And it just uses an RSA implementation), and move this whole FromFile call to io.go, so then this file purely handles the translation from cert+key to grpc options

quiltctl/command/connection.go
```diff
@@ -5,24 +5,33 @@ import (
...
 type connectionHelper struct {
+	creds  connection.Credentials
 	client client.Client
 
 	connectionFlags
 }
 
-func (ch *connectionHelper) BeforeRun() error {
+func (ch *connectionHelper) BeforeRun() (err error) {
+	ch.creds, err = credentials.Read(ch.tlsDir)
```
kayousterhout: Can you add a comment here -- like "Read the credentials to use to connect to the Quilt cluster" (as opposed to e.g. Credentials to use a cloud providers API)

minion/run.go
kayousterhout: Do you think it would be better if we wrote a file last that signals that all of the other files are there -- basically to use as a signal (I think there's a specific name for this kind of thing that @ejj might know). Then this code could first call something like tlsReady (that checks for that file) and then it knows that other errors are "real"? I think that would help avoid a bunch of possible issues. If you think this is worth doing but not now, feel free to make a Trello for later

quiltctl/command/setupTLS.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Run creates the TLS configuration.
+func (sCmd *SetupTLS) Run() int {
+	if err := util.AppFs.Mkdir(sCmd.out, 0700); err != nil {
+		log.WithError(err).Error("Failed to create output directory")
+		return 1
+	}
+
+	ca, err := rsa.NewCertificateAuthority()
```
kayousterhout: Can this go to the rsa package (or keys if you do that) too?

kklin: Can what go to the rsa package?

quiltctl/command/setupTLS.go
```diff
@@ -0,0 +1,83 @@
...
+		return 1
+	}
+
+	signed, err := rsa.NewSigned(ca, net.IP{0, 0, 0, 0})
+	if err != nil {
+		log.WithError(err).Error("Unable to create signed key pair")
+		return 1
+	}
+
+	for _, f := range io.TLSDir(sCmd.out).DaemonFiles(ca, signed) {
```
kayousterhout: This is part of why the TLSDir struct feels off to me: what IS a TLSDir? I had been thinking of it as a description of an area where TLS files are stored. But here, it isn't that (yet) -- when it's constructed it's just a string name, and it only becomes a place where files are stored after DaemonFiles is called. That's why it fees more clear to me to pass around the string directory name -- because the string directory name is all that this object consistently is. And I don't think there's much risk of someone misusing it or not knowing how to figure out how it's used since all of the functions used to interact with it and generate other files are still in the one place (in io)

connection/credentials/tls/tls.go
kklin: I don't really follow -- what would move to the RSA package?

quiltctl/command/setupTLS.go
kklin: I think it's fine that the folder doesn't actually contain the files yet. TLSDir is a description of what files are expected to be in a "tls directory":  if you're a writer, it's a description of where you should write files, if you're a reader, it's a description of where you should read files.

I think when we talked in person my example of specifying a non-default name for the CA was a good one:
```
io.TLSDir{
  DirPath: "my_tls",
  CAName: "my_company",
}
```
defines the following directory structure:
```
my_tls
|- my_company.crt
|- my_company.key
|- signed.crt
|- signed.key
```

Basically I think of a TLS Directory as an entity that contains more information than just a directory. If you're given a `TLSDir`, you're also given guarantees about what's contained within it.

connection/credentials/tls/tls.go
kayousterhout: Lines 60-63 and 70-73 -- so the logic about the format of the keys.  That code is making some assumptions about the format of the string  that's stored in each file, so I think should be near the code that decides the format of those keys

quiltctl/command/setupTLS.go
kayousterhout: Oops sorry this doesn't make any sense (I wrote this comment on my phone and thought the rsa was coming from the golang rsa package and not the Quilt rsa package)

quiltctl/command/setupTLS.go
kayousterhout: Ok this my last plea for no TLSDir class, and if you don't buy this argument, leave it as-is:

If I'm given a TLSDir, I don't know if it describes a directory with files that exist (so I can use it for reading), or if it describes files that should be written but haven't been written yet. And the constructor doesn't check that files already in that directory comply with any format, so I could in theory do `d := new TLSDir("directory name that contains files in a totally different format")`, giving the illusion that, since `d` is a `TLSDir`, it's safe to read from it.   But then later the read would fail because that directory doesn't contain correctly formatted files.  This issue is particularly pronounced because of the way we use the class.  If the way we used the class was: (1) Create a TLS dir and write things to it and then (2) pass that TLSDir object around and use it for reading, the class would definitely make sense -- because then it's providing exactly the guarantees you described.  But instead, we use the class by constructing it from a string each time we need to read or write, so we're not getting any of the benefit of the guarantees of a class.

To handle the case you described (and that we discussed in person), I'd do it with another package say `newDirFormat`, that had all of the same functions.  So, instead of `io.ReadCredentials("my dirname")`, you'd do `newDirFormat.ReadCredentials("my dirname")`.  You could still encapsulate all of the logic about the expected filenames in the `newDirFormat` package.

If you pass in a string to functions like `ReadCredentials`, you're uniting the assumption that you're making (this directory name contains a set of files that are in a particular format) with checking that assumption (if that assumption is wrong, `ReadCredentials` will fail).

Here's one last way I was thinking about this.  Suppose I want to verify that a `dir.ReadCredentials()` call is going to succeed -- so I want to verify that `dir` was written in the correct format.  I was wondering: does the fact that `dir` is an instance of `TLSDir` help me do that?  We don't pass around the TLSDir object (since the read and write happen in different processes) so I can't just look for a `WriteCredentials` call on `dir`.  Still, my first thought was "yes the TLSDir class helps" (this is what I was thinking when we talked this afternoon) -- because I could search for TLSDirs, and use that to help figure out where the directory got written.  But this evening it occurred to me that I still have to find the `WriteCredentials` call: the existence of a TLSDir in various places doesn't help me verify, because a TLSDir can be created for a directory that's in the wrong format.  I'm stuck with looking for a `WriteCredentials` call associated with the same string -- so exactly the same work I'd do if `WriteCredentials` took a string directory name as input.

quiltctl/command/setupTLS.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/quilt/quilt/connection/credentials/tls/io"
+	"github.com/quilt/quilt/connection/credentials/tls/rsa"
+	"github.com/quilt/quilt/util"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// SetupTLS contains the options for setting up Quilt TLS.
+type SetupTLS struct {
+	out string
```
kayousterhout: Can you give this a better name? (`outputDir`?) and then use it below as the flag name?  (`out` made me think it could be other things, like output from running some internal commands)

Also I'm confused by the usage string below -- what are the two dirs?

quiltctl/command/setupTLS.go
```diff
@@ -0,0 +1,83 @@
...
+		return 1
+	}
+
+	ca, err := rsa.NewCertificateAuthority()
+	if err != nil {
+		log.WithError(err).Error("Unable to create CA")
+		return 1
+	}
+
+	signed, err := rsa.NewSigned(ca, net.IP{0, 0, 0, 0})
```
kayousterhout: Can you a comment here that this is expected to be used to communicate with a daemon on the same machine, which is why the address is 0.0.0.0? (is that right?)

quiltctl/command/setupTLS.go
kklin: Ahh now that argument I find convincing. That the error about existence of files within a TLS Directory should be separate from the error about the validity of the files within the TLS Directory. An `io.Dir` is only useful if it _guarantees_ the existence of the files that its methods specify. I'll add evaluating modularity based on the types of errors it generates to my toolbox!

Just as an exercise though, how would you augment it to specify a non-default CA name? Would you pass it as a second argument to `newDirFormat.ReadCredentials("my_tls", "my_company")`?

minion/run.go
kklin: That sounds like a good idea, I think you should make a trello card on it and we can talk about it with @ejj later

kklin: Sorry, misread your message. I'll make the card.

connection/connection.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"net"
+	"time"
+
+	"google.golang.org/grpc"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// The timeout for connecting to the daemon.
```
kklin: It's general client connections -- updated.

connection/connection.go
```diff
@@ -35,6 +37,12 @@ func Client(proto, addr string, opts []grpc.DialOption) (*grpc.ClientConn, error
 // Server creates a socket listening on `lAddr` and a grpc server. If it fails
 // to open the socket, it tries again until it succeeds.
 func Server(proto, addr string, opts []grpc.ServerOption) (net.Listener, *grpc.Server) {
+	// Don't enforce TLS on local connections. This way, local connections to the
+	// daemon (e.g. when running a spec), don't need to provide credentials.
+	if proto == "unix" {
+		opts = credentials.Insecure{}.ServerOpts()
```
kklin: I didn't think of that concern -- it actually feels like the correct thing is to pass in the credentials explicitly as they were before?

cluster/credentials.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/quilt/quilt/counter"
+	"github.com/quilt/quilt/db"
+)
+
+var credentialsCounter = counter.New("Cluster Credentials")
+
+// SyncCredentials installs TLS certificates on all machines. It generates
+// the certificates using the given certificate authority, and copies them
+// over using the given ssh key. It only installs certificates once -- once
+// certificates are in place on a machine, they are left alone.
```
kklin: What about the last sentence isn't true? Is it about when the daemon restarts?

connection/credentials/tls/io/file.go
```diff
@@ -0,0 +1,22 @@
...
+
+	// signed is the file prefix for the signed certificate.
+	signed = filename("quilt")
+)
+
+// filename represents a filename prefix for RSA keys and certificates.
+type filename string
+
+// key returns the filename for the RSA key.
+func (prefix filename) key() string {
```
kklin: Remove the `filename` type entirely

connection/credentials/tls/tls.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+
+	"github.com/quilt/quilt/util"
+)
+
+// TLS satisfies the connection.Credentials interface for configuring grpc
+// connections encrypted by TLS.
+// It requires the public certificate of a certificate authority, and a
```
kklin: One of the paragraphs happened to be line-wrapped at the end of a sentence. I made one of the lines slightly longer to avoid that as it makes it easier to read.

connection/credentials/tls/tls.go
kklin: I don't really see why it's bad to have this here. This is a constructor that can create a TLS object from any number of key types.

If you'd like some guarantee that the keys generated by the RSA package are compatible with this TLS object perhaps we should use an interface.

```
type KeyPair interface {
  Get() (tls.Certificate, error)
}

type PublicCertificate interface {
  Get() (*x509.Certificate, error)
}

func New(myCert KeyPair, ca PublicCertificate) {
}
```

The RSA class would then satisfy those two interfaces. (There may even be interfaces for this defined in the go stdlib). And throughout the code base you'd read and write RSA keys. After you read in the key, you might pass it into this tls package for use for connections.

kklin: Eh, the only problem is that the CA can't be loaded as a RSA key _pair_ on the minions because we don't give them the private key.

kklin: So we'd have to split the `rsa.KeyPair` class into `rsa.Certificate` and `rsa.KeyPair`

connection/credentials/tls/io/dir.go
```diff
@@ -0,0 +1,65 @@
+package io
```
kklin: I didn't move it into the `tls` package because this reading and writing thing is awfully specific to Quilt.. but I did change the imports to call it `tlsIO`. E.g. `ca, err := tlsIO.ReadCA(dCmd.tlsDir)`

quiltctl/command/setupTLS.go
kklin: It let you pass in the output directory as a flag, or as a positional argument.. Definitely needlessly confusing, just going to make it so you can't pass it in as a flag.

kklin: Ah, the benefit of using Go's flag package is that it generates documentation for the command.

quiltctl/command/setupTLS.go
kayousterhout: Yeah that seems like a reasonable approach, or at some point, if lots of other filenames were common, it might make sense to make a more general ReadCredentials() function that takes a full filepath as an argument, and then have the old code and the new code both call that shared function.

connection/credentials/tls/io/io_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// Generate a CA in memory that we will write to disk, and read back in.
+	ca, err := rsa.NewCertificateAuthority()
+	assert.NoError(t, err)
+
+	testDir := "/tls"
+
+	// Write the CA.
+	util.Mkdir(testDir, 0755)
+	util.WriteFile(caCertPath(testDir), []byte(ca.CertString()), 0644)
```
kayousterhout: Would it make sense to use DaemonFiles here instead (after the commit that adds that), and then check that ReadCredentials works, in addition to ReadCA()?

quiltctl/command/daemon.go
```diff
@@ -55,9 +69,79 @@ func (dCmd *Daemon) AfterRun() error {
...
+		if err != nil {
+			log.WithError(err).Errorf(
+				"Failed to parse private key %s", dCmd.adminSSHPrivateKey)
+			return 1
+		}
+	}
+
+	if sshKey == nil && dCmd.tlsDir != "" {
+		log.Info("No admin key supplied, but TLS is enabled. " +
+			"Auto-generating an in-memory key for use with TLS")
```
kayousterhout: how about "No admin SSH key was supposed, but TLS is enabled, which requires an admin SSH key to copy the TLS credentials to the cluster.  Auto-generating an in-memory SSH key."

(otherwise it sounds like the admin key is used for the actual TLS, rather than just for setting it up)

quiltctl/command/daemon.go
```diff
@@ -55,9 +69,79 @@ func (dCmd *Daemon) AfterRun() error {
...
+		sshKey, err = newSSHPrivateKey()
+		if err != nil {
+			log.WithError(err).Error("Failed to generate SSH key")
+			return 1
+		}
+	}
+
+	creds, err := credentials.Read(dCmd.tlsDir)
+	if err != nil {
+		log.WithError(err).Error("Failed to parse credentials")
```
kayousterhout: in the later commit that adds the setup-tls command, can you also augment this error message to add a suggestion to call that? e.g., "Failed to parse TLS credentials; make sure you've run the setup-tls command to generate credentials".

cluster/credentials.go
```diff
@@ -0,0 +1,163 @@
...
+// XXX: The logic to avoid overwriting existing certificates does not work
+// between restarts to the daemon. If a minion has been initialized with
+// certificates but the daemon restarts, it will overwrite the existing
+// certificate. However, because this code does not cause the minion to reload
+// the certificate from disk, it will continue to run with the old certificates.
+// This will not cause any interruption to connections as long as the same
+// certificate authority is used.
+func SyncCredentials(conn db.Conn, dst string, sshKey ssh.Signer, ca rsa.KeyPair) {
```
kklin: I renamed it `dstDir`

connection/credentials/tls/rsa/rsa.go
```diff
@@ -0,0 +1,153 @@
...
+	"errors"
+	"fmt"
+	"math/big"
+	"net"
+	"time"
+)
+
+// KeyPair represents an RSA private key and certificate. The private key is
+// kept secret and signs outgoing traffic and decrypts incoming traffic. The
+// certificate can be shared publically and is used to prove the holder's
```
kklin: Heh 😅

quiltctl/command/setupTLS.go
```diff
@@ -0,0 +1,88 @@
...
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// SetupTLS contains the options for setting up Quilt TLS.
+type SetupTLS struct {
+	outDir string
+}
+
+const setupTLSCommands = `usage: quilt setup-tls [-out=<dir>] <dir>`
```
kklin: @luise has a PR that replaces the usage strings that show positional arguments with `[OPTIONS]`, which I think makes sense.

```
kevin@Kevins-MacBook-Pro ~quilt % quilt setup-tls -h
Usage: quilt setup-tls [OPTIONS] <outDir>

Create the files necessary for TLS-encrypted communication
with Quilt.  It generates private keys and certs for the signing CA, and peers.

  -outDir string
        the directory to write the certificates
```

connection/credentials/tls/io/io_test.go
kklin: That's actually what `TestReadDaemonCerts` does. I think it's fine to leave as is.

PR #965, opened 2017-07-14
ps: Show image status

kklin:
Before, when a custom Dockerfile was being built by the master,
the `quilt ps` output looked like the system was stalled because the
containers would remain unscheduled while the image was being built.

`quilt ps` now shows "building" when the image is being built, and
"built" when the image is built, but not yet scheduled.


quiltctl/command/ps.go
```diff
@@ -177,9 +193,16 @@ func writeContainers(fd io.Writer, containers []db.Container, machines []db.Mach
...
 			status := dbc.Status
-			if dbc.Status == "" && dbc.Minion != "" {
-				status = "scheduled"
+			if status == "" {
+				if dbc.Minion != "" {
+					status = "scheduled"
+				}
+
+				if imgStatus, ok := imageStatusMap[dbc.Image]; ok {
+					status = imgStatus
```
Baisang: I'm confused, doesn't this mean that a scheduled and built container will always show up as built?

kklin: Ah you're right, good catch! I added a unit test that checks this, and a fix.

PR #966, opened 2017-07-17
ci: Retry the stop test on failure

kklin:
Every once in awhile, the stop test has been failing because the leader
never gets the stop spec from the daemon. This might be because the
test isn't sleeping long enough for the daemon to send the spec.

Rather than try to tweak the timeout, this patch modifies the test so
that we check up to 5 times whether the containers were stopped. Note
that we still fail if the test doesn't succeed on the first try, but the
additional data from extra runs will be useful for debugging.


kklin:
@kayousterhout yeah, if it works on the 5th try (i.e. after ~2 minutes) I'd like to know about it because that's taking way too long.

PR #967, opened 2017-07-18
openflow: Refactor and simplify the OpenFlow tables

ejj:
The OpenFlow tables were needlessly complicated due to a slightly too
clever usage of registers.  This simplifies it without changing the
functionality in any notable way.  This patch also lays the groundwork
for a future patch in which the OpenFlow tables are augmented to
enforce public ACLs.


minion/network/openflow/openflow.go
```diff
@@ -53,80 +47,45 @@ Tables
...
-	if reg0=1 {
-		output:reg2
+		output:LOCAL
 	}
 
-	// Send packets from the patch port to the veth.
-	if reg0=2 {
-		output:reg1
-	}
+	output:reg0
```
kklin: A comment here would be helpful IMO.

ejj:
@kklin outbound-public test passed.  Merging

PR #969, opened 2017-07-18
Distinguish more machine statuses

kklin:
Before, we only tracked whether we were connected to the minion. This
patch adds a `Status` column to the machine row that can be:
- Booting
- Connecting
- Connected
- Reconnecting


cluster/cluster.go
```diff
@@ -149,29 +140,24 @@ func (clst cluster) runOnce() {
 			return
 		}
 
-		clst.updateCloud(jr.boot, boot)
-		clst.updateCloud(jr.terminate, stop)
-		clst.updateCloud(jr.updateIPs, updateIPs)
+		clst.updateCloud(jr.boot, provider.Boot, "boot")
```
ejj: Jeez, this is so much better.  Wish I had thought of it.

PR #970, opened 2017-07-19
openflow: Refactor and simplify the OpenFlow tables (Attempt 2)

ejj:
The OpenFlow tables were needlessly complicated due to a slightly too
clever usage of registers.  This simplifies it without changing the
functionality in any notable way.  This patch also lays the groundwork
for a future patch in which the OpenFlow tables are augmented to
enforce public ACLs.

The first version of this patch (commit e10f2df48) had an error that
caused it to break all network communication and was reverted.  This
version fixed the error.

### DIFF
--- a/minion/network/openflow/openflow.go
+++ b/minion/network/openflow/openflow.go
@@ -37,14 +37,8 @@ Registers
 
 The psuedocode currently uses three registers:
 
-Reg0 -- Indicates what type of port the packet came from.  1 for a Veth.  2 for a patch
-port. 0 if neither.
-
-Reg1 -- Contains the OpenFlow port number of the veth, or zero if the packet came from
-the gateway.
-
-Reg2 -- Contains the OpenFlow port number of the patch port, or zero if the packet came
-from the gateway.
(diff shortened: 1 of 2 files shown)


minion/network/openflow/openflow.go
```diff
@@ -212,16 +163,20 @@ func AddFlows(containers []Container) error {
...
-			"load:0x%x->NXM_NX_REG2[],resubmit(,1)",
-			"%d", "%s", "%x", c.veth, c.patch)
-		flows = append(flows,
-			fmt.Sprintf(template, c.veth, ",dl_src="+c.mac, 1),
-			fmt.Sprintf(template, c.patch, "", 2),
-			fmt.Sprintf("table=2,priority=1000,dl_dst=%s,actions=output:%d",
-				c.mac, c.veth))
+		// Table 0
+		flow1 := fmt.Sprintf("table=0,in_port=%d,dl_src=%s,"+
+			"actions=load:0x%x->NXM_NX_REG0[],resubmit(,1)",
```
ejj: Note the *only* difference between this and the original version, is a change from `%d` to `%x` in this line.

PR #971, opened 2017-07-19
build: Fix format checking

kayousterhout:
The Makefile used to have 3 format-related targets: format, lint,
and format-check.  These had become intertwined over time; e.g.,
the lint command would fix errors in the code (and depended on
format), meaning that if Jenkins called lint, it wouldn't fail
for some format errors (instead, it would fix the errors in the
version of the code running on Jenkins). This commit eliminates
the format command, and updates lint to only include functionality
to check the format of code.  Now, the format make target is the
only target that will update code to comply with the correct format.

@ejj it would be great if you could look at this -- Kevin and I discussed
but your Make knowledge would be a great sanity-check.


kklin:
Assigning @ejj because I don't think my review will be very helpful here.

ejj:
Looks reasonable to me.  You tested everything?  If so I'll merge.

kayousterhout:
Yup I tested this by introducing some lint errors and making sure everything errored / fixed as expected.

PR #973, opened 2017-07-19
js: Check type of JavaScript arguments

kayousterhout:
This gives the user a more helpful error than when we wait until
the JSON is read by the go code.


stitch/bindings.js
```diff
@@ -464,16 +464,58 @@ function boxRange(x) {
...
+    this.provider = getString('provider', optionalArgs.provider);
+    this.role = getString('role', optionalArgs.role);
+    this.region = getString('region', optionalArgs.region);
+    this.size = getString('size', optionalArgs.size);
+    this.floatingIp = getString('floatingIp', optionalArgs.floatingIp);
+    this.diskSize = getNumber('diskSize', optionalArgs.diskSize);
+    this.sshKeys = getArray('sshKeys', optionalArgs.sshKeys);
     this.cpu = boxRange(optionalArgs.cpu);
     this.ram = boxRange(optionalArgs.ram);
     this.preemptible = optionalArgs.preemptible !== undefined ?
```
kklin: Should there be a type check here?

stitch/bindings.js
```diff
@@ -542,7 +584,7 @@ function Container(image, command) {
         throw new Error('bad image type');
     }
 
-    this.command = command || [];
+    this.command = getArray('Container command', command);
```
kklin: Why not name this just 'command'? I think it's more clear to have the field name match exactly in the error message.

kayousterhout: I was worried this one would be less clear than the other args above, because for the ones above, they're in a map where each param is clearly associated with a name, whereas here they're just passed in as normal args. But I agree that the inconsistency is ugly so I'll change it!

kklin: Ah, I didn't notice that. I don't feel as strongly then, up to you!

kayousterhout:
I updated this -- thanks for the review! I made the commit message more specific about what's checked.

stitch/bindings.js
```diff
@@ -464,21 +478,83 @@ function boxRange(x) {
...
+        for (let i = 0; i < arg.length; i++) {
+            if (typeof arg[i] !== 'string') {
+                throw new Error(`${argName} must be an array of strings. ` +
+                    `Item at index ${i} (${stringify(arg[i])}) is not a ` +
+                    `string.`);
+            }
+        }
+        return arg;
+    }
+    throw new Error(`${argName} must be an array of strings ` +
```
kklin: I think it'd be simpler to throw this error if `!Array.isArray`, and move the for loop you currently have to be after that check.

stitch/bindings.js
```diff
@@ -253,6 +253,20 @@ Deployment.prototype.assert = function(rule, desired) {
 };
 
 function Service(name, containers) {
+    if (typeof name !== 'string') {
```
kklin: Eh.. you can't use `getString` here because it's not valid for `name` to be undefined huh. That makes me wonder if we should assert types using something like `assertString(name)`, and have the caller assign default values.

That said, I think what you have now is 100x better than nothing, so I'm fine merging it either way. Just curious what you think.

kayousterhout: Yeah I had the same thought...ideally I think there should be an assertString that getString calls.  But I was trying not to perfect this now since we'll need to do some bigger effort to check all the types later.  Let me know if you think this is worth fixing now.

PR #976, opened 2017-07-21
openflow: Enforce ACLs for traffic to and from public

ejj:
Before this patch, if a container attempted to send packets to the
public internet without permission to.  That packet would leave the
container, traverse our OVS bridge, traverse the NAT unchanged, and be
dumped out onto the public internet with the containers internal
private source IP address.  This is obviously very bad.

This patch solves the issue by enforcing public internet ACLs in the
OpenFlow tables.  Thus, if a container attempts to send a packet to
public that's not allowed, it will be dropped before it leaves the
Quilt OVS bridge.

### DIFF
--- a/minion/network/openflow/openflow.go
+++ b/minion/network/openflow/openflow.go
@@ -152,7 +152,7 @@ func AddFlows(containers []Container) error {
 		return err
 	}
 
-	flows := containerFlows(resolveContainers(ofports, containers))
+	flows := allContainerFlows(resolveContainers(ofports, containers))
 	if err := ofctl("add-flows", flows); err != nil {
 		return fmt.Errorf("ovs-ofctl: %s", err)
 	}
@@ -160,21 +160,25 @@ func AddFlows(containers []Container) error {
 	return nil
 }
 
-func containerFlows(containers []container) []string {
+func allContainerFlows(containers []container) []string {
 	var flows []string
 	for _, c := range containers {
+		flows = append(flows, containerFlows(c)...)
+	}
+	return flows
+}
+
+func containerFlows(c container) []string {
+	flows := []string{
 		// Table 0
-		flow1 := fmt.Sprintf("table=0,in_port=%d,dl_src=%s,"+
+		fmt.Sprintf("table=0,in_port=%d,dl_src=%s,"+
 			"actions=load:0x%x->NXM_NX_REG0[],resubmit(,1)",
-			c.veth, c.mac, c.patch)
-		flow2 := fmt.Sprintf("table=0,in_port=%d,actions=output:%d",
-			c.patch, c.veth)
+			c.veth, c.mac, c.patch),
+		fmt.Sprintf("table=0,in_port=%d,actions=output:%d", c.patch, c.veth),
 
 		// Table 2
-		flow3 := fmt.Sprintf("table=2,priority=900,dl_dst=%s,action=output:%d",
-			c.mac, c.veth)
-
-		flows = append(flows, flow1, flow2, flow3)
+		fmt.Sprintf("table=2,priority=900,dl_dst=%s,action=output:%d",
+			c.mac, c.veth),
 	}
 
 	return flows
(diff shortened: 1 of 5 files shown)


minion/network/openflow/openflow.go
```diff
@@ -84,20 +84,67 @@ Table_2 {
 		output:veth{1..n}
 	}
 
-	// Otherwise output to the veth based on the dest mac.
-	for each db.Container {
-		if dl_dst=dbc.Mac {
+       for each db.Container {
+                if arp && dl_dst=dbc.mac {
```
kklin: A comment here about why we need this rule would be helpful. E.g. is this for the NAT forwarding rules to ARP the containers?

minion/network/openflow/openflow.go
```diff
@@ -84,20 +84,67 @@ Table_2 {
...
+                        output:veth
+                }
+
+		// Packets originated by the gateway (i.e. DNS) are allowed).
```
kklin: Extra `)`.

minion/network/openflow/openflow.go
```diff
@@ -178,8 +232,38 @@ func containerFlows(c container) []string {
 		fmt.Sprintf("table=0,in_port=%d,actions=output:%d", c.patch, c.veth),
 
 		// Table 2
-		fmt.Sprintf("table=2,priority=900,dl_dst=%s,action=output:%d",
+		fmt.Sprintf("table=2,priority=900,arp,dl_dst=%s,action=output:%d",
 			c.Container.Mac, c.veth),
+		fmt.Sprintf("table=2,priority=800,ip,dl_dst=%s,nw_src=%s,"+
+			"action=output:%d", c.Container.Mac, ipdef.GatewayIP, c.veth),
+		fmt.Sprintf("table=2,priority=800,ip,dl_dst=%s,nw_src=%s,"+
```
kklin: This looks like the same rule as the line above?

minion/scheduler/worker.go
```diff
@@ -197,24 +198,70 @@ func filesHash(filepathToContent map[string]string) string {
...
-func openflowContainers(dbcs []db.Container) []openflow.Container {
+func openflowContainers(dbcs []db.Container,
+	conns []db.Connection) []openflow.Container {
+
+	fromPubPorts := map[string][]int{}
+	toPubPorts := map[string][]int{}
+	for _, conn := range conns {
+		if conn.MinPort != conn.MaxPort {
+			c.Inc("Unsupported Public Port Range")
+			log.Debug("Unsupported Public Port Range")
```
kklin: `WithField("connection", conn)` would be helpful for debugging IMO

minion/scheduler/worker.go
```diff
@@ -197,24 +198,70 @@ func filesHash(filepathToContent map[string]string) string {
...
+
+		ofc := openflow.Container{
 			Veth:  ipdef.IFName(dbc.EndpointID),
 			Patch: peerQuilt,
-			Mac:   ipdef.IPStrToMac(dbc.IP)})
+			Mac:   ipdef.IPStrToMac(dbc.IP),
+			IP:    dbc.IP}
+
+		for _, l := range dbc.Labels {
+			ofc.ToPub = map[int]struct{}{}
```
kklin: Won't this break if there are multiple labels? Why not just init `ToPub` and `FromPub` when you define `ofc`?

minion/network/openflow/openflow.go
ejj: Eep good catch

minion/scheduler/worker.go
ejj: Oops good catch,  my brain just broke.

minion/network/openflow/openflow.go
```diff
@@ -84,20 +84,72 @@ Table_2 {
...
+	}
+
+	// Containers are allowed to ARP the gateway.
+	if arp {
+		output:LOCAL
+	}
+
+	for each db.Container {
+		for each toPub {
+			Outbound packets have fromPub as the destination port.
```
kklin: You're missing the `//` to mark this as a comment

minion/network/openflow/openflow.go
```diff
@@ -84,20 +84,72 @@ Table_2 {
...
+			[tcp|udp],dl_src=dbc.mac,ip_src=dbc.ip,tp_dst=toPub,
+				actions=output:LOCAL
+		}
+
+		for each fromPub {
+			Response packets have fromPub as the source port.
```
kklin: Same here re `//`

PR #979, opened 2017-07-21
ps: Show the machine's floating IP if available

kklin:
Before, we would use the machine's ephemeral public IP.


quiltctl/command/ps.go
```diff
@@ -126,9 +126,15 @@ func writeMachines(fd io.Writer, machines []db.Machine) {
 	fmt.Fprintln(w, "MACHINE\tROLE\tPROVIDER\tREGION\tSIZE\tPUBLIC IP\tSTATUS")
 
 	for _, m := range db.SortMachines(machines) {
+		// Prefer the floating IP over the public IP if it's defined.
+		pubIP := m.PublicIP
```
ejj: I wonder if it makes sense to use publicIPStr here?

ejj: (To be clear, not sure either way)

kklin: As in, you'd like the connected ports to show up in the machine output?

PR #981, opened 2017-07-21
ctl: Rename `ps` to `show` while keeping `ps` as an alias

ejj:
Most people don't know what `ps` is, making the `quilt ps` command
seem jargony and inaccessible.  `show` is much more descriptive and
should be the default.  That said, `ps` is used in other tools, and
some users may expect it to work.  So this patch allows either to be
used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,12 @@ fixes a bug where containers would sometimes fail to resolve DNS on DigitalOcean
 that had already been booted. This most visibly resulted in containers restarting.
 - Use an exponential backoff algorithm when waiting for cloud provider actions
 to complete. This decreases the number of cloud provider API calls done by Quilt.
-- `quilt ps` now displays the image building status of custom Dockerfiles.
+- `quilt ps` is now renamed to `quilt show`, though the original `quilt ps`
+  still works as an alias to `quilt show`.
+- `quilt show` now displays the image building status of custom Dockerfiles.
 - Let blueprints write to stdout. Before, if blueprints used `console.log`, the
 text printed to stdout would break the deployment object.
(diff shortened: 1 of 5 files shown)


quiltctl/command/ps.go
```diff
@@ -19,23 +19,23 @@ import (
...
-func NewPsCommand() *Ps {
-	return &Ps{}
+// NewShowCommand creates a new Show command instance.
+func NewShowCommand() *Show {
+	return &Show{}
 }
 
-var psCommands = "quilt ps [-H=<daemon_host>]"
-var psExplanation = "`ps` displays the status of quilt-managed machines and containers."
+var psCommands = "quilt show [-H=<daemon_host>]"
```
kklin: I supposed these vars should be renamed `show*`

quiltctl/command/ps.go
```diff
@@ -45,20 +45,20 @@ func (pCmd *Ps) InstallFlags(flags *flag.FlagSet) {
 }
 
 // Parse parses the command line arguments for the ps command.
```
kklin: "for the _show_ command"

quiltctl/command/ps.go
```diff
@@ -19,23 +19,23 @@ import (
 // An arbitrary length to truncate container commands to.
```
kklin: The file should be renamed `show.go`

quiltctl/command/ps_test.go
```diff
@@ -15,25 +15,25 @@ import (
 	"github.com/quilt/quilt/db"
```
kklin: This file should be renamed show_test.go

kklin:
Also, should this change be part of the Changelog? I don't feel strongly either way.

PR #983, opened 2017-07-22
Clean up Google List() code

kayousterhout:


cluster/google/google.go
```diff
@@ -84,6 +84,29 @@ func New(namespace, zone string) (*Cluster, error) {
...
 }
 
+// getNetwork extracts the NetworkInterface and AccessConfig from a
+// Google instance, and handles checking that these are properly
+// defined in the instance.
+func getNetwork(inst *compute.Instance) (
+	*compute.NetworkInterface, *compute.AccessConfig, error) {
+	if len(inst.NetworkInterfaces) != 1 {
+		return nil, nil, fmt.Errorf("Google instances are expected to "+
+			"have exactly 1 interface; for instance %v, "+
```
kklin: The name format could use `%s` rather than `%v`

cluster/google/google.go
```diff
@@ -84,6 +84,29 @@ func New(namespace, zone string) (*Cluster, error) {
...
+			"found %d", inst.Name, len(inst.NetworkInterfaces))
+	}
+	iface := inst.NetworkInterfaces[0]
+	if len(iface.AccessConfigs) != 1 {
+		return nil, nil, fmt.Errorf("Google instances expected to "+
+			"have exactly 1 access config (instances "+
+			"without a config will not be accessible "+
+			"via the public internet, and Google "+
+			"does not support more than one config); "+
+			"for instance %v, found %d access configs",
```
kklin: Same here re %v format string.

cluster/google/google.go
```diff
@@ -84,6 +84,29 @@ func New(namespace, zone string) (*Cluster, error) {
 	return &clst, nil
 }
 
+// getNetwork extracts the NetworkInterface and AccessConfig from a
+// Google instance, and handles checking that these are properly
+// defined in the instance.
+func getNetwork(inst *compute.Instance) (
```
kklin: I think the name `getNetwork` is mildly confusing -- my first thought is that it's referring to the private network that the instances are a part of. E.g. we have a `clst.createNetwork` defined already.

kayousterhout: What about getNetworkConfig? getNetworkInfo? getNetworkInterfaceAndAccess?

cluster/google/google.go
kayousterhout: @ejj led me astray...

kklin: What'd @ejj say?

cluster/google/google.go
kklin: `getNetworkConfig` sounds fine

cluster/google/google.go
kayousterhout: OH I was asking him about the go-ish way to print structs, and he said to use %v, but I realized I actually ended up printing the name here instead of the whole struct -- which is more useful and also %s appropriate.  So this is all my bad and now fixed!

kklin: I see. Btw, `%+v` can be useful for printing structs sometimes -- it adds the field names.

PR #989, opened 2017-07-24
Usage strings

luise:


quiltctl/command/counter.go
```diff
@@ -17,11 +17,11 @@ import (
...
 var counterCommands = "quilt counters [-H=<daemon_host>] <target>"
-var counterExplanation = fmt.Sprintf(`quilt counters displays internal counters tracked
-for debugging purposes.  It's intended for Quilt experts.
+var counterExplanation = fmt.Sprintf(`Display internal counters tracked for
+debugging purposes.  Most users will not need this command.
 
-<target> should be %q to retrieve the counters for the
-daemon, or a machine's ID.`, daemonTarget)
+<target> should be %q to retrieve the counters for the daemon. To retrieve the
+counters for a machine, let <target> be the machine's ID.`, daemonTarget)
```
kklin: I think `let <target> be` sounds a little weird

quiltctl/command/stop.go
```diff
@@ -25,10 +25,12 @@ func NewStopCommand() *Stop {
 var stopCommands = `quilt stop [-H=<daemon_host>] [-containers] [-namespace=<namespace> |
 <namespace>]`
 
-var stopExplanation = "`stop` " + `creates an empty Stitch for the given namespace, and
-sends it to the Quilt daemon to be executed. If no namespace is specified,
-` + "`stop`" + ` attempts to use the namespace currently tracked by the daemon. The
-result is that resources associated with the namespace, such as VMs, are freed.`
+var stopExplanation = `Stop a deployment.
+
+This will free all resources (e.g. VMs) associtaed with the deployment.
```
kklin: "associtaed" -> "associated"

util/util.go
```diff
@@ -211,7 +211,8 @@ func BackoffWaitFor(pred func() bool, cap time.Duration, timeout time.Duration)
 
 // PrintUsageString formats and prints usage strings for quiltctl commands.
 func PrintUsageString(commands string, explanation string, flags *flag.FlagSet) {
-	fmt.Println("Usage: " + commands + "\n\n" + explanation + "\n")
+	fmt.Println("\nUsage: " + commands + "\n\n" + explanation + "\n")
+	fmt.Println("Options:")
 	if flags != nil {
```
kklin: Not related to this PR, but this feels slightly off.. I think the `flags` shouldn't be allowed to be nil (i.e. don't make it a pointer). The only place we use the global flag set is in the top-level parsing, which should create a local `FlagSet` object anyways.

If you agree with this can you make a card on it?

quilt.go
```diff
@@ -17,15 +17,15 @@ import (
...
+var quiltCommands = "quilt [OPTIONS] COMMAND"
 
 var quiltExplanation = `An approachable way to deploy to the cloud using Node.js.
 
 To see the help text for a given command:
-quilt <command> --help`
+quilt COMMAND --help
+
+Commands:
+  counters, daemon, debug-logs, inspect, logs, minion, ps, run, ssh, stop, version`
```
kklin: What do you think of autogenerating this list using the map in `quiltctl.go`? We could also make add `Description() string` to the command interface so that these commands could have a short description as well.

If you agree, can you make a card?

quiltctl/command/minion.go
```diff
@@ -46,6 +48,9 @@ func (mCmd *Minion) InstallFlags(flags *flag.FlagSet) {
 
 // Parse parses the command line arguments for the minion command.
 func (mCmd *Minion) Parse(args []string) error {
+	if len(args) > 0 {
```
kklin: Could you add a test for this? There's already flag tests in minion_test.go

quilt.go
luise: Oh, yeah, that's a great idea! I'll make a card!

util/util.go
luise: Yeah, that sounds good. I'll make a card.

luise:
@kklin , addressed your comments!

PR #990, opened 2017-07-24
api/server: Fix proxying credentials

kklin:
I didn't thoroughly test after making a change to the TLS PR..


api/server/server.go
```diff
@@ -54,13 +52,14 @@ func Run(conn db.Conn, listenAddr string, runningOnDaemon bool,
 		return err
 	}
 
+	serverCreds := creds
 	// Don't enforce TLS on local connections. This way, local connections to the
 	// daemon (e.g. when running a spec), don't need to provide credentials.
```
kayousterhout: Can you expand this comment a bit -- something like 

If the protocol is Unix, it means that this is the server running on the daemon, so we don't enforce TLS on inbound connections (e.g., from quiltctl connections).  The is helpful because it means that quiltctl connections (e.g., quilt run) don't need to provide credentials.  We still use the full credentials for clients, which are used to connect to machines in the cluster.

(the goal here is to make it obvious why it's different for clients / servers)

PR #991, opened 2017-07-24
ci: Test that the cluster enforces credentials

kklin:
This commit adds a test that tries to connect to the cluster in various
ways, and ensures that only allowable connections succeed.


kklin:
This should also be reviewed: https://github.com/quilt/tester/pull/46

PR #993, opened 2017-07-25
ovs: Upgrade to version 2.7.2

kklin:


kklin:
Realized that I didn't actually properly test this -- it didn't work. Closing for now -- I'll try to get it working in the next 30 minutes or so, but if not, I don't think it's a big deal if it misses the release.

PR #997, opened 2017-07-25
docs: Remove build dependency on rsync

kklin:
We now use `cp` rather than `rsync` to copy the built files.

@kayousterhout was there a particular reason to use `rsync`? Seems like a bit overkill to me


ejj:
Yeah @kayousterhout, @luise,  I'd want to know why rsync was used in the first place before merging.  Otherwise I"m happy with it.

kayousterhout:
This LGTM -- in a previous iteration the docs were compiled in a different directory and moved over, but because of the directory structure, it was easier to use rsync than mv.  I agree that we should change this to cp!

PR #1003, opened 2017-07-27
Docs formatting

luise:


docs/_GettingStarted.md
```diff
@@ -198,7 +198,8 @@ ID in the `quilt ps` output as the argument to `quilt ssh` to login to that
 container:
 
 ```console
-$ quilt ssh bd68 # Replace the argument to ssh with the ID of your container
+#Replace the argument to ssh with the ID of your container
```
kklin: Sigh, that looks so ugly. Did you see this? https://github.com/jneen/rouge/issues/636

docs/_BlueprintAPI.md
```diff
@@ -20,10 +20,10 @@ object must be passed to the `Container` constructor.
 For example,
 
 ```javascript
-new Container(new Image("my-image-name",
+var container = new Container(new Image("my-image-name",
```
kayousterhout: Can you change this to `const` (the new style guide recommends always `let` or `const` to avoid bugs with `var` -- so I've been trying to consistently use that in the docs too)

docs/_GettingStarted.md
kayousterhout: Should this be 
`$ # Replace the argument ....`? (it seems weird to me to have the comment without a prompt)

Also I wonder if we should just say "Replace bd68 with the ID of your container" so it's super obvious and a little shorter

kklin: It's more confusing to have the prompt and comment IMO.

kayousterhout: Since all of these approaches are viewed as ugly by at least one of us...why don't we eliminate the comment and just put it in the text above or below?  We could also then change the command to `quilt ssh <your container ID here>` and then in the text say something like "e.g., for the PS output above, we'd use container ID bd68".

docs/_BlueprintAPI.md
luise: I looked around the docs, and it seems like they all use `var`, so I decided to do that for consistency. I think I might stick with that for now, and then we can change all of them at the same time in a different commit. Sounds ok?

luise:
@kklin, updated the # comments. Going to leave the `var` for now, as explained in my comment to Kay.

docs/_BlueprintAPI.md
kayousterhout: sg!

docs/_GettingStarted.md
```diff
@@ -194,11 +194,12 @@ into a browser window.  A site with "Hello, world!" text should appear.
```
kklin: "adsfs"? (in the commit message)

docs/_BlueprintAPI.md
```diff
@@ -32,7 +32,7 @@ with the `github.com/my/web_repo` repository cloned into `/web_root`.
 If the Dockerfile is saved as a file, it can simply be read in:
 
 ```javascript
-new Container(new Image("my-image-name", fs.readFileSync("./Dockerfile")))
+var container = new Container(new Image("my-image-name", fs.readFileSync("./Dockerfile")));
```
kklin: Is there a particular reason you want to assign the variable? I think it makes it ever so slightly harder to read, and doesn't add much.

docs/_GettingStarted.md
luise: Sigh. That's what happens after a long conference day :)

docs/_BlueprintAPI.md
luise: Yeah, I did that because all our other examples are full lines with semicolons and so, so just looked weird to me. The missing semicolon looks like a typo at first, and it's just not consistent with the rest of our docs. And finally, it somewhat clarifies that `new Container` doesn't deploy a new container, but just creates a reference.

PR #1005, opened 2017-07-27
Use the latest stable release of Docker Engine

kklin:
We were running an "Edge" release before, which is not as thoroughly
vetted as "Stable".


cluster/cloudcfg/template.go
```diff
@@ -76,9 +76,10 @@ initialize_minion() {
 }
 
 install_docker() {
+	curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
```
luise: Is it safe to not verify the key fingerprint as Docker otherwise recommends in [the installation guide](https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/#install-using-the-repository)?

kklin: Good point, that was an oversight on my part. I've gone ahead and added the checking here https://github.com/quilt/quilt/pull/1005/commits/804342b2629282480e7bafffa8921e951be47404#diff-aed3b75ed526cb6b9faeb2197551d600R79

PR #1006, opened 2017-07-27
docs: Document installing Quilt as root

kklin:
I didn't add the note in the README (https://github.com/quilt/quilt/blob/master/README.md#install) because it links to the GettingStarted page.

I also updated documentation here, so they should be reviewed by the same person:
- https://github.com/quilt/install/pull/6
- https://github.com/quilt/demos/pull/6


docs/_GettingStarted.md
```diff
@@ -51,6 +51,12 @@ Next, use Node.js's package manager, `npm`, to install Quilt:
 $ npm install -g @quilt/install
 ```
 
+Note that if installing as root, the `--unsafe-perm` flag is required:
+
+```console
+$ sudo npm install -g @quilt/install --unsafe-perm
+```
+
```
luise: Hmm, I wonder if this should just be in the docs and not in the README. How many people do you think would need this, Kevin? I feel like not a lot of people run as root.

kklin: Both @kayousterhout and @dimant ran into it yesterday. I think it might be more common on Linux because installing globally requires running as root in the default setup.

kayousterhout: Is it easy to characterize here why you'd need to install as root?

kayousterhout: Hah ok cool then fine to leave as-is

kayousterhout: Hah ok cool then fine to leave as-is

kklin:
@luise can you review the other PRs as well? (linked in the description)

PR #1007, opened 2017-07-27
registry: Don't use image cache when building images

kklin:
Before, if an image was rebuilt with the same Dockerfile, we would use
the cached version of the image and not actually rebuild it. This was
confusing for images that interacted with the network such as the
website because it wouldn't pull in the latest website files.

For now, this commit changes the building logic to never use the cache.
Note that we do not rely on the caching logic when there are multiple
containers with the same Dockerfile, so this change will not slow down
the system in the case where multiple copies of a container with a
custom Dockerfile are deployed.


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Quilt Change Log
 
 Up Next
 -------------
+- Don't use the image cache when building custom Dockerfiles. This is necessary
```
kayousterhout: can you change this to add "on the Quilt master" after "cache"? (to make obvious this is not changing anything about the local Docker setup)

minion/docker/docker.go
```diff
@@ -204,7 +204,7 @@ func (dk Client) RemoveID(id string) error {
 
 // Build builds an image with the given name and Dockerfile, and returns the
 // ID of the resulting image.
-func (dk Client) Build(name, dockerfile string) (id string, err error) {
+func (dk Client) Build(name, dockerfile string, noCache bool) (id string, err error) {
```
kayousterhout: Ok actually can you also change this to useCache? In general I think it's considered bad style to use negatively-named booleans because it ends up having to think about the double negative (false noCache = use the cache), which is confusing

kklin: I considered it, but the docker API calls it `noCache` so I thought it'd be confusing if it didn't match. You could also argue that the default value should be to use the cache, so it's fitting that the empty value of the argument matches the default.

Anyways, don't feel strongly either way, just wanted to make sure you took those points into consideration. Let me know if you'd still like me to change it.

kayousterhout: I didn't know about `noCache`.  I still would be _slightly_ in favor of just "useCache", because it seems possible we'll expose this to a user eventually, and I think we favor intuitive semantics to our users over consistency with lower level tools.  But we could also change this if/when that happens, so up to youl

kklin: I went ahead and changed it

PR #1008, opened 2017-07-27
make: Don't run `govendor` when creating coverage targets

ejj:
Because the `go-coverage` target depends the PACKAGES variable, make
was calling govendor whether or not that target was eventually
evaluated.  The govendor tool may not be installed when a user first
runs `make go-get`, so this would cause spurious warnings.  This patch
pulls the calculation of targets into the body of `go-coverage`,
causing it only to be evaluated when needed.

Reported-By: Diman Todorov <diman.todorov@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,16 +74,15 @@ COV_SKIP= /api/client/mocks \
 	  /scripts/format \
 	  /version
 
-COV_PKG = $(subst github.com/quilt/quilt,,$(PACKAGES))
-go-coverage: $(addsuffix .cov, $(filter-out $(COV_SKIP), $(COV_PKG)))
+go-coverage:
+	for package in $(filter-out $(COV_SKIP), $(subst github.com/quilt/quilt,,$(PACKAGES))) ; do \
+	    go test -coverprofile=.$$package.cov .$$package && \
+	    go tool cover -html=.$$package.cov -o .$$package.html ; \
+	done
 	echo "" > coverage.txt
 	for f in $^ ; do \
 	    cat .$$f >> coverage.txt ; \
 	done
-
-%.cov:
-	go test -coverprofile=.$@ .$*
-	go tool cover -html=.$@ -o .$@.html
 
 js-coverage:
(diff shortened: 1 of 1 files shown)


kayousterhout:
I don't think this change quite fixes the problem because the PACKAGES variable is still defined (and used elsewhere).

Can we instead change PACKAGES to `PACKAGES=$(shell go list ./... | grep -v /vendor/)`?  I found that recommended here: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2366#issuecomment-190120274

ejj:
So it definitely does fix the problem (I reproduced ]and tested it).  The PACKAGES variable is used elsewhere, but only in the body of the make targets.  Since variables in make are by default evaluated lazily, that means that the variable is never evaluated and `govendor` is never called.  Before this patch, PACKAGES was used in the dependencies of the `go-coverage` target and thus had to be evaluated every time make ran.

I think doing a `go list` and filtering out with grep is a reasonable idea, but I think that would be something to do after this patch if it makes sense.  If we were to make that change without doing this patch, we would still be calling `go list` every time we ran make, whether or not we needed it, which would feel a bit slower.  At any rate, I don't feel strongly either way about whether we should use `govendor list` or `go list` ...

ejj:
(btw, on an unrelated side note, this is the problem with make ... it's total magic)

kayousterhout:
D'oh I tested your change by running a `make lint`, which uses PACKAGES

kayousterhout:
One piece of evidence for my theory re:vulnerability to this issue resurfacing is https://github.com/quilt/quilt/pull/589

ejj:
Yeah the forward compatibility issue is a pretty reasonable concern.  I think the best thing to do would be to merge this, and then in a separate PR move away from govendor for the package variable as a future proofing mechanism.  I'm generally in favor of reducing our reliance on that tool to the best we can. 

Feel free to make a PR if you'd like to change it =)

PR #1010, opened 2017-07-29
docs: Add warning about the quilt dir location

kayousterhout:
It's tempting to use git to get the Quilt code; this commit
adds a clear warning to the documentation to make sure that
the Quilt directory is in the location go expects.

Here's what the docs look like with this warning (we may need to re-color this after @luise's Slate formatting patch goes in):

image


luise:
@kayousterhout , I changed the colors in https://github.com/quilt/slate/pull/4. Let me know if you'd prefer something different

dimant:
Possibly link to the go documentation explaining about GOROOT and GOCODE?

kayousterhout:
@kklin fine to go ahead and merge this -- since the colors will get changed in a separate PR anyway (and Luise has an outstanding PR for that!).  

@dimant thanks for the suggestion -- there's actually already a link to that documentation at the beginning of the developer section in the docs: http://docs.quilt.io/#developer-setup

PR #1012, opened 2017-07-31
Lint and tweak docs

luise:
Eventually, we should automatically check that docs are linted, just like we test that the README compiles.


docs/_BlueprintWritersGuide.md
```diff
@@ -174,15 +174,15 @@ service.  With Quilt, by default, all network connections are blocked.  To allow
 lobsters to talk to mysql, we need to explicitly open the mysql port (3306):
 
 ```javascript
-lobstersService.connect(3306, sqlService);
+sqlService.allowFrom(3306, lobstersService);
```
kklin: `3306` and `lobsterService` should be swapped. same for the other change.

luise: Ah, good catch. Fixed it now!

PR #1026, opened 2017-08-05
Increase Database Coverage

ejj:


db/connection.go
```diff
@@ -68,7 +68,7 @@ func (c Connection) less(r row) bool {
 		return c.To < o.To
 	case c.MaxPort != o.MaxPort:
 		return c.MaxPort < o.MaxPort
-	case c.MinPort != o.MaxPort:
+	case c.MinPort != o.MinPort:
```
kayousterhout: super nit but can you change the commit message to "Unit test" from "Cover" (I *guessed* this meant test but it wasn't as obvious as it could have been)

db/db_test.go
```diff
@@ -9,21 +9,27 @@ import (
...
 	err := conn.Txn(AllTables...).Run(func(db Database) error {
-		m = db.InsertMachine()
+		db.InsertMachine()
 		return nil
 	})
 	if err != nil {
 		t.FailNow()
 	}
 
+	dbms := MachineSlice(conn.SelectFromMachine(func(m Machine) bool {
```
kayousterhout: Can you add a comment here about what this is testing? (I think "Make sure exactly one machine has been inserted into the database"?)

db/db_test.go
```diff
@@ -153,12 +170,26 @@ func TestContainerString(t *testing.T) {
...
-		"Labels: [label1], Env: map[test:tester], Status: testing, " +
-		"Created: " + fakeTimeString + "}"
+	exp = "Container-1{run test/test run /bin/sh, ImageID: imageid, " +
+		"DockerID: DockerID, Minion: Test, StitchID: 1, IP: 1.2.3.4, " +
+		"Hostname: hostname, Labels: [label1], Env: map[test:tester], " +
+		"Status: testing, Created: " + fakeTimeString + "}"
 
 	assert.Equal(t, exp, c.String())
+
+	conn := New()
```
kayousterhout: can you leave the old testcontainerString test name here, and put these lines in a new test (e.g., testContainerInsertion)? I've found the current tests often hard to read because a bunch of stuff is all in one testing function, when it could be easily separated.

db/db_test.go
```diff
@@ -611,22 +642,256 @@ func TestGetBlueprintNamespace(t *testing.T) {
...
+	t.Parallel()
+
+	p, err := ParseProvider("Amazon")
+	assert.NoError(t, err)
+	assert.Equal(t, p, Amazon)
+
+	_, err = ParseProvider("error")
+	assert.Error(t, err)
+
+	assert.Equal(t, Amazon, ProviderSlice([]Provider{Amazon}).Get(0))
```
kayousterhout: What are these two lines testing? (It looks like you're testing that making a slice works correctly? Or is this testing specifically the ProviderSlice class? (in that case can these lines go into their own test for ProviderSlices?

db/db_test.go
```diff
@@ -611,22 +642,256 @@ func TestGetBlueprintNamespace(t *testing.T) {
...
+	minions := MinionSlice(conn.SelectFromMinion(func(i Minion) bool { return true }))
+	assert.Equal(t, 1, minions.Len())
+
+	minion := minions[0]
+	assert.Equal(t, "foo", minion.Blueprint)
+	assert.Equal(t, id, minion.getID())
+
+	assert.Equal(t, "Minion-1{Self=true, HostSubnets=[]}", minion.String())
+
+	assert.Equal(t, minion, minions.Get(0))
```
kayousterhout: What is this line testing? That MinionSlice properly aliases []Minion? (it seems kind of useless)

db/db_test.go
```diff
@@ -611,22 +642,256 @@ func TestGetBlueprintNamespace(t *testing.T) {
...
+		func(i Placement) bool { return true }))
+	assert.Equal(t, 1, placements.Len())
+
+	placement := placements[0]
+	assert.Equal(t, "foo", placement.Size)
+	assert.Equal(t, id, placement.getID())
+
+	assert.Equal(t, "Placement-1{Exclusive=false, Size=foo}", placement.String())
+
+	assert.Equal(t, placement, placements.Get(0))
```
kayousterhout: Same thing here as above re: what is this testing

db/db_test.go
```diff
@@ -611,22 +642,256 @@ func TestGetBlueprintNamespace(t *testing.T) {
...
+
+	labels := LabelSlice(conn.SelectFromLabel(
+		func(i Label) bool { return true }))
+	assert.Equal(t, 1, labels.Len())
+
+	label := labels[0]
+	assert.Equal(t, "foo", label.Label)
+	assert.Equal(t, id, label.getID())
+
+	assert.Equal(t, "Label-1{Label=foo, ContainerIPs=[]}", label.String())
```
kayousterhout: It would be nice if these tests separated the tests for the Label class functions (e.g., this line) from the tests for Inserting and then using Select.  Not worth fixing -- just food for thought.  As-is, with these tests as a big blob, it's harder to get a clear sense of how the unit test maps to the code getting tested.  In a similar vein, given the way the code is broken down, it would also be nice if this were in it's own label_test.go file - so the unit test files map clearly to the files that they're testing.

db/machine.go
```diff
@@ -126,22 +126,13 @@ func (m Machine) String() string {
...
 	case l.Role != r.Role:
 		return l.Role == Master || r.Role == ""
-	case upl != upr:
-		return upl
-	case downl != downr:
-		return !downl
-	case l.ID != r.ID:
-		return l.ID < r.ID
+	case l.CloudID != r.CloudID:
+		return l.CloudID > r.CloudID // Prefer non-zero IDs.
```
ejj: It doesn't really matter much ... I guess I did it because the CloudID has some semantic meaning, while the ID is a completely meaningless database identifier.

db/connection.go
ejj: 👍

db/db_test.go
ejj: So really the only thing we're trying to verify is that conn.SelectFromMachine works.  There's more thorough test of db.SelectFromMachine below, so we really only need to check that it works.  Added a short comment.

ejj:
@kayousterhout updated.  Really the only major change was to add a patch at the end that breaks up  db_test.go

kayousterhout:
Can you squash the first and last commits?  Git doesn't track the diff well when you split it into files, which means having it in a separate commit doesn't add any value (and it's harder for me to look at the two commits separately because of all of the redundancy)

ejj:
I actually disagree, I don't like the idea of squashing it.  The nice thing about having separate commits, is it's very clear what the code changes are in the first commit.  While in the second commit, it's *just* moving things around.  If you squash them, it becomes harder to tell what was a pure move, and what was a move+change.

kayousterhout:
Then can you do the move in the first commit (in which case you're moving very little code since there wasn't much old code) and then add the new stuff in the next commit?  Seems weird to add a bunch of code in one place and then immediately have another commit that moves it (and move commits are super annoying for git log / blame, so nice to avoid as much as possible)

kayousterhout:
Or did you not too it that way b/c it was too much work?

ejj:
Nah it's fine.  Just did it.

PR #1028, opened 2017-08-08
js: Remove deprecated language features

kklin:
This PR removes language features that block the transition from deploying services to deploying containers.

These two PRs that use floating IP placements should be merged at the same time:
- https://github.com/quilt/infrastructure/pull/3
- https://github.com/quilt-corp/infra/pull/3


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ non-deterministic and rely on pulling data from the network.
 - Use the latest release of OVS (2.7.2).
 - Remove support for invariants.
 - Remove support for placement based on service groups.
```
luise: In the commit message, shouldn't it be `Service.placeOn({floatingIp: floatingIp})`?

kklin: `{floatingIp}` actually evaluates to `{'floatingIp': valueOfFloatingIp}`. In other words, it uses the variable name as the key, and the variable value as the value. It's a little funky, but it's recommended by the AirBnB styleguide (https://github.com/airbnb/javascript#es6-object-concise).

luise: Ah, that's interesting

PR #1030, opened 2017-08-08
js: Ensure container hostnames are unique by appending IDs

kklin:
Before, libraries that deployed containers with hostnames had to
internally ensure that they never deployed two containers with the same
hostname. This situation is common when users expect calling the library
twice will return two separate deployments, rather than two identical
deployments with the same hostnames.

As a side-effect, this change also ensures that a container and label
will not be assigned the same hostname. Before, it was possible to
create a Service and Container with the same hostname.


stitch/bindings.js
```diff
@@ -397,16 +397,16 @@ Service.prototype.getQuiltConnections = function() {
...
 };
 
-let labelNameCount = {};
-function uniqueLabelName(name) {
-    if (!(name in labelNameCount)) {
-        labelNameCount[name] = 0;
+let hostnameCount = {};
+function uniqueHostname(name) {
+    if (!(name in hostnameCount)) {
+        hostnameCount[name] = 0;
```
Baisang: Is there a reason why we don't just initialize this to 1?

``` js
if (!(name in hostnameCount)) {
    hostnameCount[name] = 1;
    return name;
}
hostnameCount[name]++;
return name + hostnameCount[name];

kklin: I agree that that's simpler -- I've added another commit that makes the change you suggested. Thanks!

PR #1035, opened 2017-08-10
ci: Don't destructure Quilt imports

kklin:
Destructuring the imports makes it hard to tell what functions were
imported.


ejj:
I might note in the commit message that it doesn't work with old versions of quilt as well

PR #1039, opened 2017-08-14
Small changes before deploying containers

kklin:
This PR is an assortment of changes I made for deploying containers that can be merged now.


minion/engine.go
```diff
@@ -39,8 +39,8 @@ func portPlacements(connections []db.Connection) (placements []db.Placement) {
 	}
 
 	for _, labels := range ports {
-		for _, tgt := range labels {
-			for _, other := range labels {
+		for i, tgt := range labels {
```
ejj: This is very confusing.  Needs a comment.

minion/engine_test.go
```diff
@@ -363,25 +364,41 @@ func fired(c chan struct{}) bool {
...
-
+			actual = db.PlacementSlice(view.SelectFromPlacement(nil))
 			return nil
 		})
 
-		assert.Equal(t, len(exp), len(placements))
-		for _, p := range exp {
-			_, ok := placements[p]
-			assert.True(t, ok)
+		key := func(left, right interface{}) int {
```
ejj: Did you consider doing a HashJoin where you always set "TargetLabel" to the lexicographically smaller of the two in the key?

kklin: No, but that sounds cleaner!

PR #1041, opened 2017-08-16
Implement ACLs on Digital Ocean

mmussomele:
This shouldn't be considered ready for merge, but it's at the point where it could be reviewed. Merge-ablility pending on running intergration tests.

### DIFF
--- a/cloud/digitalocean/client/mocks/Client.go
+++ b/cloud/digitalocean/client/mocks/Client.go
@@ -10,6 +10,29 @@ type Client struct {
 	mock.Mock
 }
 
+// AddRules provides a mock function with given fields: _a0, _a1
+func (_m *Client) AddRules(_a0 string, _a1 []godo.InboundRule) (*godo.Response, error) {
+	ret := _m.Called(_a0, _a1)
+
+	var r0 *godo.Response
+	if rf, ok := ret.Get(0).(func(string, []godo.InboundRule) *godo.Response); ok {
+		r0 = rf(_a0, _a1)
+	} else {
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).(*godo.Response)
+		}
+	}
+
+	var r1 error
+	if rf, ok := ret.Get(1).(func(string, []godo.InboundRule) error); ok {
+		r1 = rf(_a0, _a1)
+	} else {
+		r1 = ret.Error(1)
+	}
+
+	return r0, r1
+}
+
 // AssignFloatingIP provides a mock function with given fields: _a0, _a1
 func (_m *Client) AssignFloatingIP(_a0 string, _a1 int) (*godo.Action, *godo.Response, error) {
 	ret := _m.Called(_a0, _a1)
@@ -67,6 +90,38 @@ func (_m *Client) CreateDroplet(_a0 *godo.DropletCreateRequest) (*godo.Droplet, 
 	var r2 error
 	if rf, ok := ret.Get(2).(func(*godo.DropletCreateRequest) error); ok {
 		r2 = rf(_a0)
+	} else {
+		r2 = ret.Error(2)
+	}
+
+	return r0, r1, r2
+}
+
+// CreateFirewall provides a mock function with given fields: _a0, _a1
+func (_m *Client) CreateFirewall(_a0 string, _a1 []godo.OutboundRule) (*godo.Firewall, *godo.Response, error) {
+	ret := _m.Called(_a0, _a1)
+
+	var r0 *godo.Firewall
+	if rf, ok := ret.Get(0).(func(string, []godo.OutboundRule) *godo.Firewall); ok {
+		r0 = rf(_a0, _a1)
+	} else {
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).(*godo.Firewall)
+		}
+	}
+
+	var r1 *godo.Response
+	if rf, ok := ret.Get(1).(func(string, []godo.OutboundRule) *godo.Response); ok {
+		r1 = rf(_a0, _a1)
+	} else {
+		if ret.Get(1) != nil {
+			r1 = ret.Get(1).(*godo.Response)
+		}
+	}
+
+	var r2 error
+	if rf, ok := ret.Get(2).(func(string, []godo.OutboundRule) error); ok {
+		r2 = rf(_a0, _a1)
 	} else {
 		r2 = ret.Error(2)
 	}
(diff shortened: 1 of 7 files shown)


cloud/digitalocean/digitalocean.go
```diff
@@ -284,8 +316,102 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+		}
+		curACLSet[key] = struct{}{}
+	}
+
+	var curACLs aclList
+	for key := range curACLSet {
+		curACLs = append(curACLs, key)
+	}
+
+	_, toAdd, toRemove := join.HashJoin(aclList(desired), curACLs, nil, nil)
```
kklin: There's an `acl.Slice` already

cloud/digitalocean/digitalocean.go
```diff
@@ -284,8 +316,102 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+			panic(err)
+		}
+
+		to, err := strconv.ParseInt(ports[1], 10, 0)
+		if err != nil {
+			panic(err)
+		}
+
+		key := acl.ACL{
+			CidrIP:  cur.Sources.Addresses[0],
```
kklin: I think we should just loop over all the addresses and generate ACLs for them. Then it's obvious that we aren't going to get an out of bounds error, or miss some sources.

cloud/digitalocean/digitalocean.go
```diff
@@ -284,8 +316,102 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+
+func syncACLs(desired []acl.ACL, current []godo.InboundRule) ([]godo.InboundRule,
+	[]godo.InboundRule) {
+
+	curACLSet := map[acl.ACL]struct{}{}
+	for _, cur := range current {
+		ports := strings.Split(cur.PortRange, "-")
+		from, err := strconv.ParseInt(ports[0], 10, 0)
+		if err != nil {
+			panic(err)
```
kklin: What if someone fiddled with the DigitalOcean UI and added some ACLs in a different format? I think we should just log here, and continue. We might even consider removing unparsable rules, not sure.

cloud/digitalocean/digitalocean.go
```diff
@@ -284,8 +316,102 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+		}
+
+		to, err := strconv.ParseInt(ports[1], 10, 0)
```
kklin: These `ParseInts` could just be `strconv.Atoi`

cloud/digitalocean/digitalocean.go
kklin: A check on the length of `ports` would also be good I think

cloud/digitalocean/digitalocean.go
```diff
@@ -284,8 +316,102 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+	if _, err := prvdr.AddRules(firewall.ID, add); err != nil {
+		return err
+	}
+	if _, err := prvdr.RemoveRules(firewall.ID, remove); err != nil {
+		return err
+	}
 	return nil
 }
+
+func (prvdr Provider) getCreateFirewall() (*godo.Firewall, error) {
```
kklin: Any particular reason this should return a pointer?

cloud/digitalocean/digitalocean.go
```diff
@@ -284,8 +316,102 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+		if firewall.Name == prvdr.namespace {
+			return &firewall, nil
+		}
+	}
+
+	firewall, _, err := prvdr.CreateFirewall(prvdr.namespace, allowAll)
+	return firewall, err
+}
+
+func syncACLs(desired []acl.ACL, current []godo.InboundRule) ([]godo.InboundRule,
```
kklin: This function signature would be a little easier to read IMO if you named the return arguments

cloud/digitalocean/digitalocean.go
```diff
@@ -176,24 +177,19 @@ func (prvdr Provider) List() (machines []db.Machine, err error) {
...
+	var eg errgroup.Group
 	for _, m := range bootSet {
 		if m.Preemptible {
 			return errors.New("preemptible instances are not yet implemented")
 		}
 
-		go func(m db.Machine) {
-			errChan <- prvdr.createAndAttach(m)
-		}(m)
+		m := m
```
kklin: This looks weird.. I could totally see someone removing that line in the future. I wonder if we should curry `m`:
```
func machineAction(m db.Machine, act func(db.Machine) error) func() error {
  return func() error {
    return act(m);
  }
}
```

Might be overkill though. At the least I'd name the variable something other than the iterator variable

cloud/digitalocean/client/mocks/Client.go
```diff
@@ -1,5 +1,3 @@
-// Code generated by mockery v1.0.1 DO NOT EDIT.
```
kklin: I don't think these headers should be removed

cloud/digitalocean/digitalocean.go
```diff
@@ -34,6 +32,39 @@ var c = counter.New("Digital Ocean")
...
+// associated with it. It accepts the ones listed below, and we want to allow traffic
+// only based on IP and port, so allow them all.
+//
+// See https://developers.digitalocean.com/documentation/v2/#add-rules-to-a-firewall
+var protocols = []string{"tcp", "udp", "icmp"}
+
+// DigitalOcean firewalls block all traffic that is not explicitly allowed. We want to
+// allow all outgoing traffic.
+var allowAll = []godo.OutboundRule{
+	{
```
kklin: These are sort of repetitive, I wonder if we should generate them in a function. Don't feel strongly about it though.

cloud/digitalocean/digitalocean_test.go
```diff
@@ -317,21 +317,119 @@ func TestStop(t *testing.T) {
...
 		{
-			CidrIP:  "ocean",
+			CidrIP:  "11.0.0.0/27",
 			MinPort: 22,
 			MaxPort: 22,
 		},
-	})
+	}
+
+	mc.On("CreateFirewall", testNamespace, allowAll).Return(
```
kklin: Could you add a comment above each test "group" saying the behavior you're testing? It'd make the test a lot easier to follow (I'll look at the tests after you've done this)

cloud/digitalocean/client/mocks/Client.go
mmussomele: hmmm odd I just ran `make generate`. Anyhow, I put them back

cloud/digitalocean/digitalocean.go
mmussomele: I think leaving them is fine, but that reminds me I forgot IPv6, so I'll factor the destinations out and add that

cloud/digitalocean/digitalocean.go
mmussomele: not really, changed

cloud/digitalocean/digitalocean.go
mmussomele: I think wrapping it is worth it, especially if someone doesn't look carefully at why the assignment is there.

mmussomele:
Ready for another look.

kklin:
Sorry about the delay testing this!

I tried booting this branch and got errors of the form
```
WARNING [Sep  6 15:47:39.920] Could not update ACLs on DigitalOcean in sfo1. error=POST https://api.digitalocean.com/v2/firewalls: 422 (request "84cd4a2d-eaea-4162-b208-7db3e280877b") You must specify a positive value for ports.
```

I think it's failing to create the firewalls.

I know that you're busy with other things now @mmussomele, so unless you want to fix it up (which I know is tough since you don't have a DO token), one of us can take a look.

kklin:
Closing because @aegamesi opened a new PR with some bug fixes

PR #1043, opened 2017-08-17
Deploy container

kklin:
The PRs to make the other blueprints work with the new API must be merged at the same time:
- [ ] https://github.com/quilt/demos/pull/10
- [ ] https://github.com/quilt/django/pull/7
- [ ] https://github.com/quilt/elasticsearch/pull/5
- [ ] https://github.com/quilt/etcd/pull/4
- [ ] https://github.com/quilt/haproxy/pull/10
- [ ] https://github.com/quilt-corp/infra/pull/4
- [ ] https://github.com/quilt/kibana/pull/5
- [ ] https://github.com/quilt/lobsters/pull/8
- [ ] https://github.com/quilt/mean/pull/9
- [ ] https://github.com/quilt/mongo/pull/8
- [ ] https://github.com/quilt/nginx/pull/8
- [ ] https://github.com/quilt/nodejs/pull/5
- [ ] https://github.com/quilt/quilt-bot/pull/14
- [ ] https://github.com/quilt/redis/pull/6
- [ ] https://github.com/quilt/slackin/pull/3
- [ ] https://github.com/quilt/spark/pull/7
- [ ] https://github.com/quilt/tester/pull/48
- [ ] https://github.com/quilt/zookeeper/pull/4


stitch/bindings.js
```diff
@@ -238,80 +229,75 @@ Service.prototype.deploy = function(deployment) {
...
 
-    this.outgoingPublic.forEach(function(rng) {
-        connections.push({
-            from: that.name,
-            to: publicInternetLabel,
-            minPort: rng.min,
-            maxPort: rng.max,
-        });
-    });
+function boxContainers(x) {
```
kayousterhout: It's confusing that this is called boxContainerS (plural), but extra boxing only happens when there's just a single container.  I don't have a better idea though, so fine to leave it unless you can think if something better.

stitch/bindings.js
```diff
@@ -601,6 +591,113 @@ Container.prototype.getQuiltPlacements = function() {
...
+Container.prototype.allowFrom = function(srcArg, portRange) {
+    if (srcArg === publicInternet) {
+        return this.allowFromPublic(portRange);
+    }
+
+    let src;
+    try {
+        src = boxContainers(srcArg);
+    } catch (err) {
+        throw new Error(`Containers can only connect to other containers. ` +
```
kayousterhout: should this be "other containers or container groups"?

stitch/bindings.js
```diff
@@ -601,6 +591,113 @@ Container.prototype.getQuiltPlacements = function() {
...
+            `not from a Service or other object.`);
+    }
+
+    src.forEach((c) => {
+      this.allowedInboundConnections.push(
+          new Connection(c, boxRange(portRange)));
+    });
+};
+
+Container.prototype.allowOutboundPublic = function(range) {
```
kayousterhout: I'm not sure we should have this method -- it seems too repetitive with public.AllowFrom?

stitch/bindings.js
```diff
@@ -674,6 +674,13 @@ class ContainerGroup extends Array {
             dst.allowFrom(src, port);
         });
     }
+
+    /**
+     * Get all the hostnames of the containers within the ContainerGroup.
+     */
+    hosts() {
```
kayousterhout: I'm not sure this is worth adding a function for.  I'd be in favor of holding off on adding this until we've seen a huge number of cases where it's really useful (based on the philosophy you mentioned to me earlier this week -- of keeping our language as simple as possible)

docs/_BlueprintWritersGuide.md
```diff
@@ -127,67 +127,57 @@ First, let's write the Quilt blueprint to get the MySQL container up and running
 need to create a container based on the mysql image:
 
 ```javascript
-let sqlContainer = new Container('mysql:5.6.32');
+let sql = new Container('sql', 'mysql:5.6.32');
```
kayousterhout: why this change? sqlContainer -> sql makes it harder to read -- since it's not obvious what sql is

kayousterhout:
One more thought about having both Service and ContainerGroup: IIUC, it seems like a user could just use a Service instead of a ContainerGroup and get the same functionality.  They'd get this extra load balanced IP, but they could just ignore it.  Just trying to imagine ways people may potentially confuse / misuse these...

stitch/bindings.js
```diff
@@ -574,6 +545,144 @@ Container.prototype.hash = function() {
...
+Container.prototype.getQuiltPlacements = function() {
+    return this.placements.map((plcm) => {
+        plcm.targetContainer = this.id;
+        return plcm;
+    });
+};
+
+Container.prototype.allowFrom = function(srcArg, portRange) {
+    if (srcArg === publicInternet) {
+        return this.allowFromPublic(portRange);
```
ejj: Shouldn't return here

CHANGELOG.md
```diff
@@ -17,6 +17,15 @@ floating IP is now expressed as `myService.placeOn({floatingIp: '8.8.8.8'})`.
 - Restart containers if their hostname changes.
 - Fix a bug where containers might get assigned duplicate hostnames.
 - Remove `Service.children`. Container hostnames should be used from now on.
+- Changed the container constructor syntax to take optional settings as the
```
kklin: The commit to make hostnames required is the one after this.

That sounds like a good idea, but I'm not sure we have to do that for this PR. There's a couple different ways we could lay out the CHANGELOG. I'll bring it up in our meeting today.

stitch/bindings.js
```diff
@@ -489,7 +515,7 @@ Image.prototype.clone = function() {
     return new Image(this.name, this.dockerfile);
 };
 
-function Container(image, command) {
+function Container(image, {command=[], env={}, filepathToContent={}} = {}) {
```
kklin: Unfortunately, object spreads don't seem to be supported yet (at least in the 7.10.0 version I'm devving on).

Do you think I should write a function that checks for unexpected args in this commit? I was thinking we might do it for Machines and Containers at the same time in a later PR.

kayousterhout: Ah ok then I agree with you that it's fine to do this for everything in a later commit

stitch/bindings_test.js
```diff
@@ -213,17 +215,32 @@ describe('Bindings', function() {
...
-                new Container('image', ['arg1', 'arg2'])
-                    .withEnv({foo: 'bar'})
-                    .withFiles({qux: 'quuz'}),
+                new Container('image', {
+                  command: ['arg1', 'arg2'],
+                }).withEnv({foo: 'bar'})
+                  .withFiles({qux: 'quuz'}),
+            ]));
+
+            // Test passing the env and files in the constructor.
```
kklin: I did something similar by pulling the test out into a new `describe`. It's messy to leave the two tests in the same `it` block because the IDs are unique even when there are two containers with the exact same attributes.

stitch/bindings_test.js
```diff
@@ -166,155 +166,171 @@ describe('Bindings', function() {
...
-            containerC.setHostname('host2');
+            const containerA = new Container('host', 'ignoreme');
+            const containerB = new Container('host', 'ignoreme');
+            const containerC = new Container('host2', 'ignoreme');
             const hostnames = [containerA, containerB, containerC]
               .map((c) => c.getHostname());
             const hostnameSet = new Set(hostnames);
             expect(hostnames.length).to.equal(hostnameSet.size);
         });
+        it('clone doesn\'t clone the hostname index', function() {
```
kklin: I'm not sure -- this was testing a specific bug in the original code that I wrote. Before, I didn't have a `hostnamePrefix`, and just cloned the `hostname`. This would result in funky hostnames like:

```
const contA = new Container('host', ...) // => 'host'
const contB = contA.clone() // => 'host2'
const contC = contB.clone() // => 'host22'
```

Do you have a better name idea?

kayousterhout: Ah I see how about "clone increments existing index if one exists"?

minion/engine_test.go
```diff
@@ -463,33 +457,14 @@ func TestPlacementTxn(t *testing.T) {
...
-			TargetLabel: "foo",
-			Exclusive:   true,
-			OtherLabel:  "foo",
-		},
-
-		db.Placement{
-			TargetLabel: "bar",
-			Exclusive:   true,
-			OtherLabel:  "bar",
+			TargetContainer: "1a84f87eebbf7dc7edda83a38f34a49f2116240b",
```
kklin: I pulled the IDs out into a variable here -- note that I didn't pull out the label names in this commit because they'll go away once containers are connected directly. I'll fix up this test more in that commit.

minion/scheduler/master.go
```diff
@@ -89,43 +89,8 @@ Outer:
...
-	_, opOK := pLabels[target]
-	oValid := !ocOK || !opOK
-
-	return tValid && oValid
-}
-
-func checkExclusionConstraint(constraint db.Placement, cLabels,
-	pLabels map[string]struct{}) bool {
-
+func checkExclusionConstraint(constraint db.Placement, toPlace db.Container,
```
kklin: I went with `canBeColocated`

minion/scheduler/master.go
```diff
@@ -89,43 +89,8 @@ Outer:
...
+	peers []*db.Container) bool {
 	if !constraint.Exclusive {
 		// XXX: Inclusive OtherLabel is hard because we can't
```
kklin: Sigh, placement needs to be redesigned. `!constraint.Exclusive` is used by machine constraints to specify that a container _should_ be placed on a certain type of machine.

minion/scheduler/master.go
```diff
@@ -136,29 +101,29 @@ func checkExclusionConstraint(constraint db.Placement, cLabels,
...
-		cLabels[label] = struct{}{}
-	}
-
-	var peerLabels map[string]struct{}
 	for _, constraint := range constraints {
-		if constraint.OtherLabel != "" {
-			peerLabels = computePeerLabels(peerLabels, peers, dbc.ID)
-			ok := checkExclusionConstraint(constraint, cLabels, peerLabels)
-			if !ok {
+		if constraint.OtherContainer != "" {
```
kklin: That won't work because `constraint.TargetContainer` might be `dbc.StitchID`, but `constraint.OtherContainer` might be empty (i.e. it's a machine placement rule).

This code is confusing and I'm hesitant of saying to not fix bad code, but a lot of this stuff will be fixed with a more general placement rewrite. Although I don't think that will happen soon.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@quilt/quilt",
-  "version": "0.0.9",
```
kklin: I'm not against it, but I also think it's worth discussing. I think it'll depend on how our release process will work. E.g. we might only bump the version number when releasing.

stitch/bindings.js
kklin: This is the method that `public.AllowFrom` calls. Users shouldn't call this.

minion/scheduler/master.go
kayousterhout: Ah ok fine to leave as-is then, since it's not directly related to your PR and sounds like there isn't a quick fix.

docs/_BlueprintWritersGuide.md
kklin: I dunno, I think it's a bit simpler. Do you feel strongly about this?

kklin:
@kayousterhout ready for another look

docs/_BlueprintWritersGuide.md
kayousterhout: No fine to leave

stitch/bindings.js
```diff
@@ -138,58 +139,32 @@ function hash(str) {
...
 
     let services = [];
     let connections = [];
     let placements = [];
+    let containers = [];
 
-    // For each service, convert the associated connections, and placements.
-    // Also, convert the service into the form required by the Quilt
-    // deployment engine.
+    // Convert the services.
```
kayousterhout: IIUC, now, you have to explicitly deploy the containers that make up a service (it no longer happens automatically as part of deploying a service).  First, why?

Second, can you update the changelog to say that more explicitly?

quilt-tester/tests/haproxy/haproxy.js
```diff
@@ -10,23 +10,25 @@ const indexPath = '/usr/share/nginx/html/index.html';
...
-  return new quilt.Container('web', 'nginx').withFiles(files);
+  return new quilt.Container('web', 'nginx', {
+    filepathToContent: {
+      [indexPath]: content,
+    },
+  });
 };
 
-let serviceA = new quilt.Service('serviceA', [
+let serviceA = [
```
kayousterhout: Can you rename these to something other than service, now that they're not actually services? Or do you think we should ignore this because we're about to rename Service to LoadBalancer? (if the latter, that's fine with me)

kklin: I actually think the name service makes sense here since it's a group of containers that traffic is being sent to. Would you prefer `loadBalancedA`?

README.md
```diff
@@ -60,8 +61,8 @@ Set up a secure network.
 
 [//]: # (b1)
 ```javascript
-webApp.allowFrom(publicInternet, 80); // Open the webservers' port 80 to the public internet.
-redis.allowFrom(webApp, redis.port); // Let the web app communicate with Redis.
+allow(publicInternet, webGroup, 80); // Open the webservers' port 80 to the public internet.
```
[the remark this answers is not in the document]
kklin: we don't support public connections to load balancers (and never have FWIW).

stitch/bindings_test.js
```diff
@@ -396,7 +396,7 @@ describe('Bindings', function() {
                 provider: 'Amazon',
             });
             checkPlacements([{
-                targetLabel: 'target',
+                targetContainer: '293fc7ad8a799d3cf2619a3db7124b0459f395cb',
```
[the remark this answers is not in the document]
kklin: To do that, I'd have to modify the `id` of the `target` Container after it's been deployed. It's pretty hacky, and I'm not sure it's worth it.

minion/engine.go
```diff
@@ -26,33 +26,44 @@ func updatePolicy(view db.Database, blueprint string) {
...
+	// the same port. We do not need to create a rule for every permutation
+	// because order does not matter for the `TargetContainer` and
+	// `OtherContainer` fields -- the placement is equivalent if the two fields
+	// are swapped.  We do so by creating a placement rule between each
+	// container, and the containers after it. There is no need to create rules
+	// for the preceding containers because the previous rules will have
+	// covered it.
+	for _, cids := range ports {
+		for i, tgt := range cids {
+			for _, other := range cids[i+1:] {
```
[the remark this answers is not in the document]
kklin: It's subtle, but it wasn't actually broken before. Before, when the placement rules were in terms of labels, you wanted to make sure none of the containers within the same label got colocated.

Would you have preferred that I leave the index unchanged in this commit, and then had a commit _after_ reducing the number of rules?

quilt-tester/tests/haproxy/haproxy.js
kayousterhout: ahh ok fine to just leave it then

README.md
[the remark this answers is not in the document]
kayousterhout: I'm confused - what's webLoadBalancer doing then?

stitch/bindings_test.js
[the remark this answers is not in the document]
kayousterhout: Ah ok cool agree not worth it

README.md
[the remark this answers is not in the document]
kklin: It's not being used directly in this code snippet -- I thought the idea was just to show that we support load balancers.

minion/engine.go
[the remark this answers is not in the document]
kayousterhout: I see that makes sense.  The extra commit would still have been useful to explicitly explain this but it's borderline.

README.md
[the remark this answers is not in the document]
kayousterhout: I see. This is confusing and probably should be removed but is not related to your PR so let's leave it for now.

kayousterhout:
I'm copying this comment so I can edit the checklist:

- [ ]  quilt/demos#10
- [ ]  quilt/django#7
- [ ]  quilt/elasticsearch#5
- [ ]  quilt/etcd#4
- [ ]  quilt/haproxy#10
- [ ]  quilt-corp/infra#4
- [ ]  quilt/kibana#5
- [ ]  quilt/lobsters#8
- [ ]  quilt/mean#9
- [ ]  quilt/mongo#8
- [ ]  quilt/nginx#8
- [ ]  quilt/nodejs#5
- [ ]  quilt/quilt-bot#14
- [ ]  quilt/redis#6
- [ ]  quilt/slackin#3
- [ ]  quilt/spark#7
- [ ]  quilt/tester#48
- [ ]  quilt/zookeeper#4

PR #1045, opened 2017-08-23
`quilt init` and `baseInfrastructure`

luise:
This is the initial version of `quilt init`. Below is a (slow) sample of what the flow could look like (it depends on which options are chosen on the way).

init

There are a few big things to do (in a later PR, or in this):
- [ ] Add docs for what the prompts in `quilt init` mean.
- [ ] Add docs for how to use `baseInfrastructure`.
- [ ] Update getting started guide.
- [ ] Add more extensive tests. Especially for `baseInfrastructure()`. I can't think of a good way to test the `prompter` as it depends asynchronously on user input. 
- [ ] Decide if `quilt init` is the right name. I would like something more related to infrastructure -- maybe `infra`, but it might not be intuitive.
- [ ] Update the changelog? 

I will also need to lint based on the Airbnb config if that patch gets merged first.


ejj:
```init: Add init functionality
This commit adds the init functionality to create infrastructure blueprints
based on user prompts.```

I don't know what "adds the init functionality" means.  I think the commit message could be made a bit clearer.

File: quiltctl/command/init/initializer.js [hunk 1]
```diff
@@ -0,0 +1,213 @@
+/* eslint no-console: 0 */
+
+const path = require('path');
+const os = require('os');
+let fs = require('fs');
```
ejj: Why not const?

quiltctl/command/init/initializer.js
```diff
@@ -0,0 +1,213 @@
...
+let fsExtra = require('fs-extra');
+const execSync = require('child_process').execSync;
+const exit = require('process').exit;
+
+const handlebars = require('handlebars');
+
+const prompter = require('./prompter');
+const consts = require('./constants');
+const util = require('../../../util/util');
```
ejj: Are these relative path things going to work when we `npm install` this?

quiltctl/command/init/initializer.js
```diff
@@ -0,0 +1,213 @@
...
+const infraTemplateFile = path.join(templateDir, 'inf_template');
+
+/**
+  * Creates a file from the given template and context, and writes the result to
+  * a file.
+  *
+  * @param {string} templateLocation The path to the template file.
+  * @param {Object.<string, string>} context The context to apply to the
+  *   template.
+  * @param {string} destination The path to write the result to.
```
ejj: All of these doc comments need an `@returns`.  If they return nothing, it can be an `@returns {void}`

util/util.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+  * Log the given string, unless this function is called during testing, in
+  * which case we do not want log messages. All test files that use this
+  * function must set `process.env.NODE_ENV = 'test'` at the top.
+  *
+  * @param {*} message The value to log.
+  * @return {void}
+  */
+function logExceptOnTest(message) {
```
ejj: Do we really need this? What's so bad about the tests logging?  Adds unnecessary complexity in my view.

ejj:
I wonder if it would make more sense to put all of this init code into it's own directory rather than burying it in the quiltclt/command  and util packages.  I'm imagining a top-level directory called `js` with a subdirectory called `init` or something.  What do you think?

quiltctl/command/init/initializer.js
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/**
+  * Given a file path, recursively create all directories in the path.
+  * This function assumes that the passed in path is to a file, not a directory.
+  *
+  * @param {string} filePath The complete file path for which to create
+  *  directories.
+  */
+function makeDirPath(filePath) {
```
ejj: I'm surprised there isn't an npm utility that does something like this.  It's a pretty standard thing `mkdir -p`.

ejj: Btw, the reason I ask, is this sort of thing is very easy to get subtly wrong for unforseen reasons.

ejj:
It looks to me like, for some reason, the tests for initializer.js aren't showing up in coverage.  Do they need to be added to some makefile variable or something?

quiltctl/command/init/initializer.js
```diff
@@ -0,0 +1,213 @@
...
+  */
+function sshGenerateKeyPair() {
+  try {
+    execSync(
+      `ssh-keygen -t rsa -b 2048 -f ${consts.quiltSshKeyLocationPrivate}` +
+      ` -N '' -C ''`,
+      {stdio: 'ignore'});
+    return true;
+  } catch (err) {
+    return false;
```
ejj: I'm slightly worried that this swallows the error.  I wonder if it makes sense to pull the caller's log message here, so you can include the error returned by ssh-keygen in it's error message.  Or alternatively I suppose you could return the err.  Not quite sure what's idiomatic.

quiltctl/command/init/initializer.js
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/**
+  * Create the infrastructure file to be used with baseInfrastructure().
+  *
+  * @param {string} infraName The user chosen name of this infrastructure.
+  * @param {Object} config The context to fill into the infrastructure template.
+  * @return {void} Returns nothing.
+  */
+function writeInfraFile(infraName, config) {
```
ejj: This function is only used in one place and is kind of trivial.  I don't see why it shouldn't just be embedded in the caller.

quiltctl/command/init/initializer.js
kklin: Btw, we could generate the keys natively using a package. I used sshpk for the tester. Not sure if it would help this problem, but it'd at least remove the dependency on the existence of ssh-keygen

quiltctl/command/init/initializer.js
luise: It has to be `let` for mocking it out in the tests

quiltctl/command/init/initializer.js
luise: I'm not quite sure what you mean. When we `npm install` quilt? The dependencies are in the Quilt root, so they will be installed, and the directory structure is preserved in `node_modules`, so the relative path should work. I can't seem to find any reason it shouldn't work - maybe @kklin  knows of one?
That being said, the relative path is super ugly and not scalable. There are various workarounds, but given how little JS code we have, I'm not sure it's worth doing anything about.

quiltctl/command/init/initializer.js
luise: Yeah, I know. This is all linted with Google. I will update when the Airbnb linting is merged.

util/util.js
luise: It just looks ugly in the test output, but I can get rid of it. I don't feel strongly. I agree that it might be unnecessary complexity.

quiltctl/command/init/initializer.js
ejj: Ah ok.  Perhaps a comment?

quiltctl/command/init/initializer.js
ejj: Fair enough.  I'd just try it out and make sure ...

quiltctl/command/init/initializer.js
ejj: I personally trust ssh-keygen.  I'm a bit skeptical of some random npm package for this type of security relate thing.

quiltctl/command/init/initializer.js
```diff
@@ -0,0 +1,213 @@
...
+  let keyPath;
+
+  switch (answers[consts.sshKeyOption]) {
+    case consts.skip:
+      return '';
+
+    case consts.sshGenerateKey:
+      let success = sshGenerateKeyPair();
+      if (!success) {
+        throw new Error('failed to create SSH key. Is ssh-keygen installed?');
```
ejj: Yeah looking at this now.  I think it'd be cleaner if `sshGenerateKeyPair` returned nothing, and just threw this exception on error.  You wouldn't have to catch it, could just let it bubble up.

quiltctl/command/init/initializer.js
```diff
@@ -0,0 +1,213 @@
...
+
+    case consts.sshUseExistingKey:
+      keyPath = answers[consts.sshKeyPath];
+      break;
+
+    default:
+      throw new Error('Unrecognized ssh key option: ' +
+        `${answers[consts.sshKeyOption]}`);
+  }
+  // return fs.readFileSync(keyPath, 'utf8').trim();
```
ejj: should be deleted

quiltctl/command/init/prompter.js
```diff
@@ -0,0 +1,428 @@
...
+function quiltSshKeyExists() {
+  return (fs.existsSync(consts.quiltSshKeyLocationPublic) &&
+      fs.existsSync(consts.quiltSshKeyLocationPrivate));
+}
+
+/**
+  * Returns true if the input string is a number. Otherwise return a string.
+  *
+  * @param {string} input The string to check.
+  * @return {boolean|string} True if the input string is a number. Otherwise
```
ejj: This is a little weird.  I wonder if it would be cleaner to throw an exception rather than rely on the return value?

quiltctl/command/init/prompter.js
```diff
@@ -0,0 +1,428 @@
...
+  * Prompt the user for basic details about the infrastructure.
+  *
+  * @return {Promise} A promise that contains the user's answers.
+  */
+function basicPrompts() {
+  const questions = [
+    {
+      type: 'input',
+      name: consts.namespace,
+      message: 'Choose a deployment identifier (namespace):',
```
ejj: I would just say "Choose a namespace".  "deployment identifier" sounds scary, and most users aren't going to know what either it, or namespace means.

I might say something like "Choose a namespace.  (The default is fine for most people.)" Just so they know it doesn't really matter.

Though more I think about it, should we really be asking this?  Why not just make everyone use the default?  If they want a new namespace, they can edit the js.

quiltctl/command/init/prompter.js
```diff
@@ -0,0 +1,428 @@
...
+  * @return {Promise} A promise that contains the user's answers.
+  */
+function machineConfigPrompts(provider) {
+  const sizeChoices = getSizeDescriptions(provider);
+  sizeChoices[consts.other] = consts.other;
+  const questions = [
+    {
+      type: 'confirm',
+      name: consts.preemptible,
+      message: 'Do you want to run spot instances?',
```
ejj: I would say "preemptible" instances just so the terminology is consistent.   Btw, do these questions need defaults?

quiltctl/command/init/prompter.js
```diff
@@ -0,0 +1,428 @@
...
+        return isNumber(input);
+      },
+      when(answers) {
+        return answers[consts.sizeType] !== consts.instanceTypeSize;
+      },
+    },
+    {
+      type: 'input',
+      name: 'ram',
+      message: 'How many GiB of RAM do you want?',
```
ejj: I think GB is more common that GiB

quiltctl/command/init/provider.js
```diff
@@ -0,0 +1,182 @@
...
+    */
+  hasPreemptible() {
+    return this._hasPreemptible;
+  }
+
+  /**
+    * @return {boolean} True if this provider has size descriptions,
+    *   otherwise false.
+    */
+  hasSizes() {
```
ejj: I wonder how much value this sort of function adds.  Why can't the caller just check if the values are undefined directly?  It feels like unnecessary boilerplate, but I'm not sure.

util/util.js
ejj: What about this, why not just say "util.log == console.log" and then in unit tests mock out util.log?  That way you don't have to deal with this environment variable stuff.

stitch/bindings.js
```diff
@@ -24,6 +24,30 @@ function githubKeys(user) {
     return keys;
 }
 
+/**
+  * Returns a base infrastructure. The infrastructure can be deployed to simply
+  * by calling .deploy() on the returned object.
+  * The base infrastructure should be created with `quilt init`.
```
ejj: s/should/could

stitch/bindings.js
```diff
@@ -24,6 +24,30 @@ function githubKeys(user) {
...
+  *
+  * @param {Object} quilt A copy of the Quilt module as returned by `require`.
```
ejj: This seems a bit weird to me.  Why can't we just `require` quilt ourselves?

ejj: Or rather, sense we are the quilt modules, call the functions directly?  Basically, it seems very weird to be passing imports around to functions ... it's a sign there's something wrong with the design

docs/_QuiltCLI.md
```diff
@@ -29,6 +29,7 @@ $ quilt COMMAND --help
 | `counters`   | Display internal counters tracked for debugging purposes. Most users will not need this command. |
 | `daemon`     | Start the quilt daemon, which listens for quilt API requests.                                    |
 | `debug-logs` | Fetch logs for a set of machines or containers.                                                  |
+| `init`       | Create an infrastructure that can be accessed in blueprints using baseInfrastructure().          |
```
ejj: I don't think this should be in its own commit.  I'd fold it in with the commit that adds the `quilt init` command.

ejj:
Overall LGTM.  This is much cleaner than I expected.  I'm done with my review, happy to discuss any of the comments.

ejj:
Oh also, please create a card in asana for each of the TODO items for this.  We should probably get them done next sprint.

util/util.js
luise: Ah, yes, I like that

quiltctl/command/init/initializer.js
luise: Cool, I'll go with `ssh-keygen` for now then, and propagate the error up. 
Then we can always switch to `sshpk` if the dependency gets annoying -- `sshpk` a Joyent package and seems to be super popular, so I am inclined to trust it. Should be easy to change later.

quiltctl/command/init/initializer.js
luise: I'll move it to the caller body!

quiltctl/command/init/initializer.js
luise: I could let `sshGenerateKeyPair` throw the error. The error that is thrown is not very helpful by itself -- it just says the command failed -- so I will probably still catch the error and then just throw a new error with the helpful information. Let me know if that doesn't sound right!

quiltctl/command/init/prompter.js
luise: Hmm, yeah, maybe we just shouldn't ask. People who care will be comfortable editing the JS. I'll get rid of it.

quiltctl/command/init/prompter.js
luise: Cool, I'll say preemptible. The `list` questions just "default" to the first choice in the list, and the preemptible question defaults to true automatically -- I'll add a `default` option to that question, just to make it clearer.

quiltctl/command/init/prompter.js
luise: GiB and GB are actually slightly different (1 GiB = 2^30 B, 1 GB = 10^9 B). They are close to equal, but GiB is actually the correct value to use in this case (according to AWS). Do you prefer GB even though it's less exact?

quiltctl/command/init/provider.js
luise: It was to make the other code independent of the internal provider logic -- e.g. should you check agains `undefined` or `{}` or `false`?

But since the functions aren't used often, it's not a lot of hassle to make the comparisons. I'm fine with switching over to `X !== Y`.

stitch/bindings.js
luise: Yeah, so I talked to Kevin about this before I left. The problem is that the infrastructure file calls quilt functions (`new Machine` etc.), so it needs a quilt module. And if we make the file require `quilt` there has to be a package.json in `~/.quilt/infra` that we then need to `npm install` and `update` with new quilt versions, which gets messy. This was the best solution we could come up with, but if you have other suggestions, I'm happy to hear them. I agree this is ugly!

I don't think we can call the functions directly from here since it is stuff that should happen when the infrastructure blueprint is run. I might be misunderstanding what you suggested.

docs/_QuiltCLI.md
luise: OK!

quiltctl/command/init/prompter.js
luise: Oh, yeah, I'll fix that. Turns out the `validate` prompter option can handle `Error`s as well - I didn't realize that which is why I returned the string. The error is much better.

stitch/bindings.js
kklin: Maybe you could simplify this a bit by not requiring the caller to pass a reference to the import, and passing `module` to `infraGetter`? At least then the ugliness will be hidden from the user.

quiltctl/command/init/initializer.js
luise: We should keep this in mind when you fix `npm install` to get both the JS and the Go binary.

luise:
@ejj  and @kklin , I addressed your comments. 
Re: non-code comments:
* I left the code in quiltctl/command/init for now, but it might make sense to make a js repository at some point.
* The initializer.js test doesn't show up in codecov, because codecov only considers go code for now.
* I will add cards to trello for the things I mentioned in the PR comment

quiltctl/command/init/prompter.js
ejj: No, we'll do whatever is more correct.

PR #1049, opened 2017-08-25
util: Remove the JavaScript util library

ejj:
The JavaScript util library was extremely thin.  The functions it
exports could easily be embedded in their callers.  Furthermore, it
was so far from the code that needed it, that packaging up the init
script in `@quilt/install` was proving more complex than necessary.
To simplify things, this patch ditches it.


luise:
`bindings.js` actually also used `util.js`, but mistakenly didn't require the module. This PR fixes that: https://github.com/quilt/quilt/pull/1050.

PR #1057, opened 2017-08-26
Fix minor issues in JSDoc

kayousterhout:
This series of commits fixes a few minor issues in the JSDoc, and one more major one, which is to add @private annotations to all of the functions that are not exported publicly.


stitch/bindings.js
```diff
@@ -468,6 +479,7 @@ function getString(argName, arg) {
 }
 
 /**
+ * @private
  * @param {string} argName - The name of `arg` (for logging).
  * @param {Object.<string, string>} arg - The map of strings.
  * @return {Deployment} An empty object if `arg` is not defined, and
```
kklin: Not part of this PR, but this should be `Object.<string, string>` right?

stitch/bindings.js
```diff
@@ -306,7 +306,7 @@ Service.prototype.deploy = function serviceDeploy(deployment) {
 };
 
 /**
- * Allow inbound connections to the load balancer. Note that this does not
+ * Allows inbound connections to the load balancer. Note that this does not
```
kklin: JOOC how did we decide to use this tense? I have a slight preference for the old tense in these cases but that's just going off of what sounds nice to my ear.

ejj: @kayousterhout wanted it.  I don't care either way.  Feels like bikeshedding.  If you and kay can agree on going the other way, feel free to change it back.

kayousterhout: I had a seemingly incorrect memory of you asking me to fix the tense at some point in Go Kevin -- I don't care which we use but this is consistent with Go so seemed easier to remember.

stitch/bindings.js
kayousterhout: Yup

kayousterhout:
@kklin just in case you didn't see -- I added a new commit to fix the issue you mentioned and this is good to merge!

kklin:
@kayousterhout Ethan actually fixed up the param thing already (https://github.com/quilt/quilt/commit/2547bd0b924cb379fb7494518b864494f95d7c5a)

Could you rebase? I'll merge after

kayousterhout:
@kklin updated!

PR #1059, opened 2017-08-27
npm: Organize all of the npm packages into one subdirectory

ejj:
This patch moves bindings.js, the initializer, and `@quilt/install`
(which was formerly in a different repo), all under one subdiretory.
It also changes `@quilt/install` to use locally built binaries,
instead of pulling them from github.

Note `make jscheck` fails for me, but it's also failing on master.  This shouldn't be merged until that is sorted.


Makefile
```diff
@@ -162,3 +161,13 @@ docker-build-ovs:
...
 # Include all .mk files so you can have your own local configurations
 include $(wildcard *.mk)
+
+# Prepare the js/install directory for either `npm install` or `npm publish`.
+prep-install: linux darwin js/install/package.json
+	cp quilt_linux js/install
+	cp quilt_darwin js/install
+	cp -r js/initializer js/install
+
+
```
kklin: Extra newline?

js/bindings/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@quilt/quilt",
+  "version": "0.0.6",
+  "main": "bindings.js",
+  "license": "MIT",
+  "dependencies": {
```
kklin: The dev dependencies should be moved here as well I think

package.json
```diff
@@ -1,37 +1,19 @@
...
-    "fs-extra": "^4.0.1",
-    "handlebars": "^4.0.10",
-    "inquirer": "^3.2.1",
-    "json-stable-stringify": "^1.0.1",
-    "mkdirp": "^0.5.1",
-    "sync-request": "^4.0.1",
-    "underscore": "^1.8.3"
-  },
-  "nyc": {
-    "exclude": "**/*_test.js"
```
kklin: Should we be removing this?

package.json
```diff
@@ -1,37 +1,19 @@
 {
-  "name": "@quilt/quilt",
```
kklin: It feels weird to have this `package.json` in the root even though it's not really describing a package anymore..

js/install/package_template.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "0.2.0",
+  "homepage": "http://quilt.io",
+  "license": "MIT",
+  "preferGlobal": true,
+  "scripts": {
+    "preinstall": "node install.js"
+  },
+  "bin": {
+    "quilt": "quilt",
+    "quilt-initializer.js": "./initializer/initializer.js"
```
kklin: I thought you were going to make the initializer package a dependency of `install`, rather than merging the `package.json`s?

kklin:
Also, can you add removing the `quilt/install` repo to your checklist of things to do when merging this?

package.json
ejj: 👍

js/install/package_template.json
ejj: Yeah I'm a bit conflicted on this.  I don't think there's a good way to do it without pushing an actually @quilt/initializer package to npm when we release (so we'd have three packages).  That may be fine though, do you think I should change it?

package.json
ejj: Yeah it is weird.  As discussed offline, it's the best of several bad options.

js/bindings/package.json
ejj: Why?  We only run the tests from the main directory, so do we need them here?

Makefile
ejj: 👍

js/install/package_template.json
kklin: I don't know what's best, so I think we should just do what's easiest (i.e. merge it as is)

PR #1061, opened 2017-08-27
javascript: Fix up the js-coverage make target

ejj:


package.json
```diff
@@ -19,8 +19,7 @@
     "sinon": "^2.1.0"
   },
   "scripts": {
-    "test": "mocha ./stitch ./quiltctl/command/init/ ./util/",
-    "cov": "nyc --reporter=text-lcov npm test"
+    "test": "mocha ./stitch ./quiltctl/command/init/ ./util/"
```
kklin: Could you remove the `./util` while you're in here?

Makefile
```diff
@@ -83,7 +83,8 @@ go-coverage:
 	done
 
 js-coverage:
-	npm run-script cov > bindings.lcov
+	./node_modules/.bin/nyc npm test
+	./node_modules/.bin/nyc report --reporter=text-lcov >> coverage.lcov
```
kklin: Also, would this be more correct if it were just `>`? Not sure either way, just wondering if you had a specific reason for switching it

ejj: No reason to switch it.  Had a previous that needed it and forgot to change it back.

PR #1064, opened 2017-08-28
Add integration tests for Lobsters and Django

kklin:


quilt-tester/tests/django/django_test.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"testing"
+
+	"github.com/quilt/quilt/api"
+	"github.com/quilt/quilt/api/client"
+	"github.com/quilt/quilt/connection/credentials"
+	"github.com/quilt/quilt/quilt-tester/util"
+)
+
+func TestLobsters(t *testing.T) {
```
Baisang: this should be `TestDjango`?

kklin: That was sloppy -- thanks!

PR #1069, opened 2017-08-31
Add examples directory

kayousterhout:
These commits add a new examples directory for how to write blueprints, and copies the floating IP spec into the examples directory.  Once this PR is merged, we should delete the quilt/infrastructure repositories.  These commits copy the examples exactly as they were in the old directories, and doesn't fix any formatting or update the documentation (which we should do at some point).


kayousterhout:
BTW I wonder if this should be in stitch/examples?

kayousterhout:
@ejj this is fine to review now!

kklin:
I'd like to see a test for the example specs similar to https://github.com/quilt/quilt/blob/master/scripts/blueprints-tester/tests/test_blueprints.go

kayousterhout:
Thanks for the idea Kevin -- done!

scripts/blueprints-tester/tests/test_blueprints.go
```diff
@@ -27,6 +27,30 @@ func run(name string, args ...string) error {
...
+	if err != nil {
+		return err
+	}
+
+	for _, blueprint := range blueprints {
+		log.Infof("Testing %s", blueprint)
+
+		// Change into the directory of the blueprint in order to install
+		// dependencies.
+		os.Chdir(filepath.Dir(blueprint))
```
kklin: I think we should restore the directory after, or else if there are multiple example blueprints it won't successfully `chdir` into the second one?

kayousterhout: It's done here: https://github.com/quilt/quilt/blob/master/scripts/blueprints-tester/blueprints-tester.go#L25 (which is why the following test doesn't reset the directory either, I think)

kklin: This code is changing directories within the loop though, so that reset won't happen until after all these blueprints have been tested.

kayousterhout: Oh good point I thought Glob() returned absolute directories but now see that it does not.  Fixed!

Makefile
```diff
@@ -17,7 +17,11 @@ LINE_LENGTH_EXCLUDE=./api/pb/pb.pb.go \
 		    ./stitch/bindings.js.go
 
 JS_LINT_COMMAND = node_modules/eslint/bin/eslint.js \
-                  stitch/ quilt-tester/ cli/command/init/ util/
+                  examples/**/*.js \
```
ejj: Do you have to do could this just be `examples/`?  Based on the other elements in this variable, the linter seems smart enough to walk the tree recursively and just check the js files.

kayousterhout: The linter does walk the tree recursively but it lints examples/*/node_modules/*, which leads to lint errors in the required packages (I spent a little bit of time researching this and couldn't find a good way to ignore the node_modules directory that was better than just using this pattern here -- any ideas?)

ejj:
I think if you just add a comment explaining it, that's sufficient.

________________________________

Makefile
kklin: What if you added `**/node_modules` to `.eslintignore`?

ejj: Oh I like that idea if it works

kayousterhout:
Good idea @kklin -- I did that so this is ready to merge. @ejj

PR #1073, opened 2017-09-01
Init tests and tweaks

luise:


stitch/bindings.js
```diff
@@ -6,7 +6,10 @@ const stringify = require('json-stable-stringify');
 const _ = require('underscore');
 const path = require('path');
 const os = require('os');
-const fs = require('fs');
+
+// Use `let` to enable mocking of `fs` in tests.
+/* eslint-disable prefer-const */
```
ejj: I think this syntax is disabling it for the entire file.  You want something like `eslint-disable-line` -- I don't remember precisely, but there are other examples in the code base.

luise:
@ejj , updated.

PR #1074, opened 2017-09-01
Ignore etcd pull image errors in quilt-tester

ejj:


quilt-tester/tests/100-logs/logs_test.go
```diff
@@ -38,16 +38,31 @@ func TestMinionLogs(t *testing.T) {
...
+		// should still be logged as warnings to the user, but they shouldn't
+		// cause a test failure.
+		if strings.Contains(line, "pull image error") {
+			fmt.Printf("Ignoring pull image error: %s\n", line)
+			continue
+		}
+
+		// "goroutine 0" is the main goroutine and is printed in stacktraces.
+		if strings.Contains(line, "goroutine 0") {
+			t.Errorf("Minion logs has a stack trace")
```
kklin: This should just be `t.Error`

quilt-tester/tests/100-logs/logs_test.go
```diff
@@ -38,16 +38,31 @@ func TestMinionLogs(t *testing.T) {
...
+		checkString(t, outputStr)
+	}
+}
+
+func checkString(t *testing.T, str string) {
+	for _, line := range strings.Split(str, "\n") {
+		// Errors pulling an image are completely out of our control.  They
+		// should still be logged as warnings to the user, but they shouldn't
+		// cause a test failure.
+		if strings.Contains(line, "pull image error") {
```
kklin: This is a bit concerning to me because it'll ignore failures where an invalid image is specified. Is there a way we could be more specific about the error?

ejj: Actually this isn't an issue.  I tested with an invalid image and it isn't an error.

quilt-tester/tests/100-logs/logs_test.go
ejj: 👍

PR #1075, opened 2017-09-01
Cleanup db.Provider code

kayousterhout:
This PR does a few minor simplifications of the db.Provider code, and adds a more helpful error message when the user gives an unsupported provider.


db/constants.go
```diff
@@ -3,6 +3,7 @@ package db
 //The Role within the cluster each machine assumes.
```
kklin: Nit - `unsupp.` took me a bit to process in the commit message, I think it's fine to just expand it

kklin: Also, in the commit message you have `hiAmazon` 😛

db/constants.go
```diff
@@ -66,14 +67,22 @@ const (
...
 
+var allProviders = []Provider{
+	Amazon,
+	Google,
+	DigitalOcean,
+	Vagrant,
+}
+
 // ParseProvider returns the Provider represented by 'name' or an error.
 func ParseProvider(name string) (Provider, error) {
```
kklin: Could you add a test for this?

kklin:
Oh, also the build is failing

db/constants.go
kayousterhout: Done -- and writing that test made me find a bug in the Provider declarations

db/constants.go
kayousterhout: Yeah that was intentional (it was supposed to be a not-real provider) but I updated it to be more obvious than hiAmazon

db/constants_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseProvider(t *testing.T) {
+	t.Parallel()
+
+	_, err := ParseProvider("not_a_provider")
+	assert.Error(t, err)
```
kklin: Should we check the contents of the error string?

db/constants_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	p, err := ParseProvider(string(Amazon))
```
kklin: The table driven testing approach to this would be something like

```
tests := []struct{in string, expPrvdr Provider, expErr error}{
  {string(Amazon), Amazon, nil},
  {string(Google), Google, nil},
  {"not_a_provider", nil, "the expected error message"},
}

for _, test := range tests {
  actualProvider, actualErr := ParseProvider(test.in)
  assert.Equal(t, test.expProvider, actualProvider)
  assert.Equal(t, test.expErr, actualErr)
}
```

It'd be easier to read IMO, but since this test is pretty simple I don't necessarily think it's worth the effort. Just wanted to point out the style.

db/constants_test.go
kayousterhout: I'm slightly opposed to this because I think it leads to brittle tests where the test has to be updated for every minor change in the error message (e.g., if we added an additional provider)

db/constants_test.go
kayousterhout: Thanks for noting this -- I updated the test to do this! (And then realized there was a good simplification that could be done later using AllProviders)

db/constants_test.go
kklin: Yeah, I can see that argument. It's especially strong for when your errors are made by concatenating errors returned by 3rd party libraries, so when they change your errors, your tests have to change as well. (I've seen this problem solved by using `strings.Contains` rather than checking exact equality)

I think this case is a bit different though. I don't see it as that bad if we have to change the tests when the code changes.

kayousterhout: I went ahead and changed this, but as food for thought, here's how I currently think about testing for error messages (which has evolved somewhat after having some experience dealing with error-checking unit-test failures caused by unrelated changes):

(1) What's the risk of not checking for the error -- i.e., is it likely that the error caught by the test is caused by something *other* than the error I'd like to check for?

If yes, it's nice to check for the actual error message (or in other languages, the type of the error) to make sure it's the right error.  In this case, this code is very simple, so there is a very low likelihood of the error being the wrong one.

(2) If you check for the exact error, is it likely that unrelated code changes will force an update to the error checking?

This is the case you were getting at with 3rd party packages.  It's likely here too; e.g., if we added another cloud provider, we'd need to update the test, even though "not_a_provider" should still def throw an error.

I prefer not checking the error message when (1) is no and (2) is yes.

kklin: I think those are good guidelines. It's worth noting that smarter checking than an exact comparison makes the bar for (1) and (2) higher.

There might also be a philosophical difference here in what "very simple" code is, and whether it should be tested.

BTW, this test isn't particularly important, but I'm picking at this as an exercise for our hopefully soon-to-come style guide.

kklin: More food for thought:  sometimes error messages aren't deterministic (e.g. if `allProviders` were stored in a map, and we printed them in a random order). This makes testing them harder.

PR #1076, opened 2017-09-02
cloud: Get rid of the joinMachine struct

ejj:
It's not adding a lot of value over db.Machine.

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -153,18 +153,19 @@ func (cld cloud) runOnce() {
 }
 
 func (cld cloud) boot(machines []db.Machine) {
-	var cloudMachines []joinMachine
+	// As a defensive measure, we only copy over the fields that the underlying
+	// provider should care about instead of passing `machines` to updateCloud
+	// directly.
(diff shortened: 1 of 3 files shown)


cloud/cloud.go
```diff
@@ -153,18 +153,18 @@ func (cld cloud) runOnce() {
 }
 
 func (cld cloud) boot(machines []db.Machine) {
-	var cloudMachines []joinMachine
+	// As a defensive measure, make sure set everything to zero that a provider
```
kklin: I'm not sure what this comment means

ejj: Updated the comment, make sense now?

kklin: Kind of -- I don't get what we're defending against

ejj: We're just defending against provider writers from depending on machine fields that they shouldn't. I.E. breaking layering boundaries which can lead to icky spaghetti code.  This is why machine.Machine existed in the first place -- it's generally not best to give a module information it shouldn't be aware of.

PR #1078, opened 2017-09-05
quiltctl: Rename to cli

kayousterhout:
It's not obvious what quiltctl is (and how it's different from the
other components, like the engine and the daemon), so this commit
renames it to CLI.

This commit also renames the uses of "quiltctl client" in various
log messages to "daemon client", since the client is from the CLI
to the daemon, and daemon client is slightly more intuitive.


api/server/server.go
```diff
@@ -40,7 +40,7 @@ type server struct {
 	clientCreds connection.Credentials
 }
 
-// Run starts a server that responds to `quiltctl` connections. It runs on both
+// Run starts a server that responds to connections from the CLI. It runs on both
```
kayousterhout: @kklin is this comment still correct (does the api still run on both the daemon and the minion?)

kklin: Yes it does (so maybe the name `daemon client` isn't the best)

kklin: Although if it's just being changed in comments/log messages and makes sense in the context, it's fine

kklin:
A little context on the naming, when `quiltctl` was first written there wasn't a single CLI. Users would start the spec running process with `quilt -f config.spec`, and then query the spec with commands like `quiltctl machines`.

The new name makes much more sense now.

api/server/server.go
kayousterhout: Ah I see -- so to clarify the CLI communicates directly with both the daemon and the minion, and the api package describes the api use for those connections?

kayousterhout:
Thanks for the context -- that makes sense.

api/server/server.go
kklin: The CLI communicates directly with the daemon. The daemon sometimes opens connections to the minions, but hides this from the CLI.

And exactly re the API package

kayousterhout:
Ok updated the comments to change `daemon client` to `api client`, as per discussions with Kevin

PR #1080, opened 2017-09-06
Add the old JS documentation (from the markdown) to the jsdoc

kayousterhout:
For reference, here's the old markdown: https://github.com/quilt/quilt/blob/00e85e1d738fc55d4a6b0351e24f8220bcc5c1dd/docs/_BlueprintAPI.md


stitch/bindings.js
```diff
@@ -627,10 +627,31 @@ Machine.prototype.hash = function machineHash() {
...
  * Creates a Docker Image.
+ *
+ * If two images with the same name but different Dockerfiles are referenced, an
+ * error will be thrown.
+ *
  * @constructor
  *
- * @param {string} name - The name of the docker image.
- * @param {string} dockerfile - The string contents of the Docker file that
+ * @example <caption>Create an image that uses the nginx image stored on
```
kklin: Nice that JSDoc supports examples!

stitch/bindings.js
```diff
@@ -627,10 +627,31 @@ Machine.prototype.hash = function machineHash() {
...
+ *
+ * @example <caption>Create an Image named my-image-name that's built on top of
+ * the nginx image, and additionally includes the Git repository at
+ * github.com/my/web/repo cloned into /web_root.</caption>
+ * let image = new Image('my-image-name',
+ *   'FROM nginx\n' +
+ *   'RUN cd /web_root && git clone github.com/my/web_repo');
+ *
+ * @example <caption>Create an image named my-inage-name that's built using a
+ * Docker file saved locally at 'Dockerfile'.</caption>
```
kklin: Docker file => Dockerfile

stitch/bindings.js
```diff
@@ -627,10 +627,31 @@ Machine.prototype.hash = function machineHash() {
...
+ * If two images with the same name but different Dockerfiles are referenced, an
+ * error will be thrown.
+ *
  * @constructor
  *
- * @param {string} name - The name of the docker image.
- * @param {string} dockerfile - The string contents of the Docker file that
+ * @example <caption>Create an image that uses the nginx image stored on
+ * Docker Hub.</caption>
+ * let image = new Image('nginx');
```
kklin: Should we just make all these `let`s `const`s while we're here?

stitch/bindings.js
```diff
@@ -627,10 +627,31 @@ Machine.prototype.hash = function machineHash() {
...
+ * github.com/my/web/repo cloned into /web_root.</caption>
+ * let image = new Image('my-image-name',
+ *   'FROM nginx\n' +
+ *   'RUN cd /web_root && git clone github.com/my/web_repo');
+ *
+ * @example <caption>Create an image named my-inage-name that's built using a
+ * Docker file saved locally at 'Dockerfile'.</caption>
+ * let container = new Image('my-image-name', fs.readFileSync('./Dockerfile'));
+ *
+ * @param {string} name - The name to use for the docker image, or if no
```
kklin: Capitalize "docker"

stitch/bindings.js
```diff
@@ -627,10 +627,31 @@ Machine.prototype.hash = function machineHash() {
...
+ *   dockerfile is specified, the name of an image in Docker Hub to use
```
kklin: Capitalize "dockerfile"

kklin: Also, I think the description is a bit specific to DockerHub. You could specify a different registry if you wanted (e.g. `quay.io/coreos/etcd`)

stitch/bindings.js
```diff
@@ -627,10 +627,31 @@ Machine.prototype.hash = function machineHash() {
...
+ *   (e.g., nginx or nginx:1.13.3).
+ * @param {string} [dockerfile] - The string contents of the Docker file that
```
kklin: Docker file => Dockerfile

stitch/bindings.js
```diff
@@ -643,13 +664,40 @@ Image.prototype.clone = function imageClone() {
...
+ * re-started with the new image.
  *
  * @constructor
  * @implements {Connectable}
  *
+ * @example <caption>Create a Container with hostname myApp that uses the nginx
+ * image on Docker Hub, and that includes a file located at /etc/myconf with
+ * contents foo.</caption>
+ * let container = new Container(
+ *   'myApp', 'nginx', {filepathToContent = {'/etc/myconf': 'foo'}});
```
kklin: s/=/:

stitch/bindings.js
```diff
@@ -716,12 +764,32 @@ Container.prototype.withEnv = function containerWithEnv(env) {
...
+ *   that filepathToContent is set to the given mappng.
+ */
 Container.prototype.withFiles = function containerWithFiles(fileMap) {
   const cloned = this.clone();
   cloned.filepathToContent = fileMap;
   return cloned;
 };
 
+/**
+ * @returns {string} The container's hostname.
```
kklin: I'm a little surprised this didn't throw a lint error and fail the build, I seem to remember a rule about preferring `@return` over `@returns`. Do you know what's up?

stitch/bindings.js
```diff
@@ -553,11 +553,39 @@ function getBoolean(argName, arg) {
...
+ *   it can be deployed.  This argument is not required, so that users can
+ *   create a template to use for all machines in the cluster;
+ *   {@link Machine#asWorker} and {@link Machine#asMaster} can be called on the
+ *   template to create a machine with the appropriate role, as in the example.
+ * @param {string} [optionalArgs.region] - The region the machine will run-in
+ *   (provider-specific; e.g., for Amazon, this could be 'us-west-2').
+ * @param {string} [optionalArgs.size] - The instance type (provider-specific).
+ * @param {Range|int} [optionalArgs.cpu] - The desired number of CPUs.
+ * @param {Range|int} [optionalArgs.ram] - The desired amount of RAM in GiB.
+ * @param {int} [optionalArgs.diskSize] - The desired about of disk space in GB.
```
kklin: s/about/amount

stitch/bindings.js
kayousterhout: Yeah I was pretty excited about this!! Alas the code doesn't get formatted in the same way as the other code in our markdown files (it's fixed-width but doesn't get colored), but we can fix this in the future _I think_

stitch/bindings.js
kayousterhout: I looked into this and the airbnb style guide doesn't enable the rule preferring one over the other (and we use them inconsistently now).  I added a commit here that changes all of the ones in bindings.js to be consistent.  I'll change them to be consistent throughout our codebase (and add the eslint config so they get checked) later, because it conflicts with one of my other outstanding PRs to move files around

kayousterhout:
Thanks for looking over this so thoroughly!

stitch/bindings.js
```diff
@@ -13,7 +13,7 @@ const githubCache = {};
 /**
```
kklin: The commit message should read "Change @return to @returns"

kayousterhout: Derp fixed

PR #1084, opened 2017-09-11
Check for invalid optional arguments

luise:


stitch/bindings.js
```diff
@@ -9,6 +9,7 @@ const os = require('os');
 const fs = require('fs');
```
kklin: This deserves to be in the changelog IMO

stitch/bindings.js
```diff
@@ -439,6 +442,36 @@ function boxRange(x) {
...
 
 /**
+  * Throws an error if the first object contains any keys that do not appear in
+  * the second object. This is useful for checking if the user passed any
+  * invalid options to functions that take optional arguments.
+  * @private
+  *
+  * @param {Object} opts - The Object to check for redundant keys.
+  * @param {Object} obj - The object to compare against.
+  * @return {void} Throws an error if redundant keys are found.
```
kklin: I'm a bit surprised there's no JSDoc tag for errors

stitch/bindings.js
```diff
@@ -439,6 +442,36 @@ function boxRange(x) {
...
+  */
+function checkExtraKeys(opts, obj) {
+  // Sometimes, prototype constructors are called internally by Quilt. In these
+  // cases, an existing object is passed as the optional argument, and the
+  // optional argument thus contains not just the keys  passed by the user, but
```
kklin: Two spaces here

stitch/bindings.js
```diff
@@ -439,6 +442,36 @@ function boxRange(x) {
...
+  // also the keys Quilt set on the object, as well as all the prototype
+  // methods. Since we only want to check the optional arguments passed by the
+  // user, we skip all calls internally from Quilt (indicated by having the
+  // refID set in the options).
+  if (objectHasKey.call(opts, '_refID')) { return; }
+
+  const extras = [];
+  Object.keys(opts).forEach((key) => {
```
kklin: This would be a bit cleaner if you used `const extras = Object.keys(opts).filter(...)`

stitch/bindings_test.js
```diff
@@ -82,6 +82,13 @@ describe('Bindings', () => {
         sshKeys: ['key1', 'key2'],
       }]);
     });
+    it('errors when passed invalid optional arguments', () => {
+      expect(() => new Machine({ badArg: 'foo' })).to
+        .throw('Unrecognized keys passed to Machine constructor: ');
```
kklin: Shouldn't there be the list of extra keys in all these `throw` messages?

stitch/bindings.js
```diff
@@ -439,6 +442,36 @@ function boxRange(x) {
 }
 
 /**
+  * Throws an error if the first object contains any keys that do not appear in
```
kklin: I think it'd be helpful to explain here the pattern of checking the argument object against the object we instantiate.

kklin: The fact that we use that pattern, and why it's a valid approach.

stitch/bindings.js
```diff
@@ -439,6 +442,36 @@ function boxRange(x) {
...
+  */
+function checkExtraKeys(opts, obj) {
+  // Sometimes, prototype constructors are called internally by Quilt. In these
+  // cases, an existing object is passed as the optional argument, and the
+  // optional argument thus contains not just the keys  passed by the user, but
+  // also the keys Quilt set on the object, as well as all the prototype
+  // methods. Since we only want to check the optional arguments passed by the
+  // user, we skip all calls internally from Quilt (indicated by having the
+  // refID set in the options).
+  if (objectHasKey.call(opts, '_refID')) { return; }
```
kklin: Sigh, this is ugly.. we probably shouldn't be passing instantiated objects as the optional argument. I'm not sure what to do here.

stitch/bindings.js
luise: Ah, cool, I'll do that!

stitch/bindings.js
luise: There is a `@throws` tag! I'll fix.

stitch/bindings.js
luise: Yeah, I wasn't sure either how to restructure the `clone()` calls. Do you think I should fix this in this PR? If so, I'll think more about `clone()`.

stitch/bindings_test.js
luise: In the actual error message, yes, the extra keys will be there. But the test passes as long as the string is a substring of the error message -- and this way we don't have to deal with the indeterministic order of the keys.
Would you prefer having the keys in the test?

kklin: I'm fine leaving them out. What if you tested the full error for the errors that are only one argument?

stitch/bindings.js
kklin: No, definitely not a fix for this PR

stitch/bindings_test.js
luise: Yup, that's a great idea.

luise:
@kklin , updated

PR #1086, opened 2017-09-14
Ensure that there are not machines in multiple regions/providers

aegamesi:


stitch/bindings.js
```diff
@@ -257,6 +257,16 @@ function vet(deployment) {
     }
     dockerfiles[name] = c.image.dockerfile;
   });
+
+  // check to make sure all machines have the same region and provider
```
kklin: Nit -- the comment should start with a capital letter, and end with a period.

stitch/bindings.js
```diff
@@ -257,6 +257,16 @@ function vet(deployment) {
     }
```
kklin: Nit - We usually use the `js` prefix for this code rather than `bindings`, and the first letter after the colon should be capitalized.

kklin: I'd also put something in the commit description describing why we made this change:
```
Quilt currently does not support running in multiple regions/providers simultaneously, so this commit makes `quilt run` fail immediately if multiple regions or providers are specified.
```

stitch/bindings.js
```diff
@@ -257,6 +257,16 @@ function vet(deployment) {
...
     dockerfiles[name] = c.image.dockerfile;
   });
+
+  // check to make sure all machines have the same region and provider
+  let lastMachine;
+  deployment.machines.forEach((m) => {
+    if (lastMachine !== undefined &&
+      (lastMachine.region !== m.region || lastMachine.provider !== m.provider)) {
+      throw new Error('All machines must have the same provider and region.');
```
kklin: Could you add a unit test?

kayousterhout: Can you add the names of the two providers here (so something like, at the end, add "(found providers X and Y)")

kklin: @kayousterhout If we do that, do you think we should include _all_ regions if there's more than 2 specified?

kayousterhout: I wouldn't bother since it's harder to do, and if there are more than 2 the user will quickly figure it out once they eliminate one and then get a new error.  More generally - IMO the principle behind this is that if we report an error to the user, we should try to give them all the info they need to debug it (and the usability of a system is often driven by how easy it is to debug issues). Not listing the providers at all is omitting useful info; listing the first two gives them enough to debug iteratively.

stitch/bindings.js
aegamesi: Done

kklin:
Btw @aegamesi usually once a PR is ready for more review, we ping the reviewer. But it looks like you addressed the comments I left, so I'll review it again now!

stitch/bindings.js
```diff
@@ -257,6 +257,18 @@ function vet(deployment) {
     }
     dockerfiles[name] = c.image.dockerfile;
   });
+
+  // Check to make sure all machines have the same region and provider
```
kklin: The comment is still missing a period.

stitch/bindings.js
```diff
@@ -257,6 +257,18 @@ function vet(deployment) {
...
+  let lastMachine;
+  deployment.machines.forEach((m) => {
+    if (lastMachine !== undefined &&
+      (lastMachine.region !== m.region || lastMachine.provider !== m.provider)) {
+      throw new Error('All machines must have the same provider and region.\n'
```
kklin: I find it a bit odd to have a newline in the error message. If it's more readable with the newlines then I'd leave it, but otherwise, I'd just use spaces.

stitch/bindings_test.js
```diff
@@ -660,6 +660,40 @@ describe('Bindings', () => {
...
+    });
+    it('machines with different regions', () => {
+      deployment.deploy([new Machine({
+        provider: 'Amazon',
+        region: 'us-west-2',
+      }), new Machine({
+        provider: 'Amazon',
+        region: 'us-east-2',
+      })]);
+      expect(deploy).to.throw('All machines must have the same provider and region.'
```
kklin: This would be a bit more readable if you used the backticks and linebreaks, rather than string concatenation I think.

stitch/bindings.js
```diff
@@ -257,6 +257,18 @@ function vet(deployment) {
     }
     dockerfiles[name] = c.image.dockerfile;
   });
+
+  // Check to make sure all machines have the same region and provider
+  let lastMachine;
+  deployment.machines.forEach((m) => {
+    if (lastMachine !== undefined &&
+      (lastMachine.region !== m.region || lastMachine.provider !== m.provider)) {
```
kklin: There's actually a case where this code will misleadingly fail:

If no region is specified, we pick a default region in the deployment engine. For example, I think for Amazon this is `us-west-1`. So
```
[
  { provider: 'Amazon', region: 'us-west-1' },
  { provider: 'Amazon'},
]
```
will fail, when it's actually fine. Don't think it's worth accounting for here though, the right fix is probably to move that default region logic into the machine constructor.

stitch/bindings.js
```diff
@@ -257,6 +257,18 @@ function vet(deployment) {
     }
     dockerfiles[name] = c.image.dockerfile;
   });
+
+  // Check to make sure all machines have the same region and provider
+  let lastMachine;
```
kayousterhout: Btw no need to change this but for future reference, since this is a pattern that comes up regularly: you could change this to firstMachine and then you could axe line 270.

aegamesi:
@kklin I've made the changes you requested. I didn't realize you could have multiline comments with backticks. I switched the message to use spaces though -- so I think I *have* to use string concatenation in the tests (unless there's another new JavaScript feature I don't know about)?

stitch/bindings.js
```diff
@@ -257,6 +257,17 @@ function vet(deployment) {
...
     dockerfiles[name] = c.image.dockerfile;
   });
+
+  // Check to make sure all machines have the same region and provider.
+  let firstMachine = deployment.machines[0];
+  deployment.machines.forEach((m) => {
+    if (firstMachine !== undefined &&
+      (firstMachine.region !== m.region || firstMachine.provider !== m.provider)) {
+      throw new Error(`All machines must have the same provider and region. `
+        + `Found providers ${firstMachine.provider} in region '${firstMachine.region}' `
```
kayousterhout: why is the region in quotes but not the provider?

aegamesi: I figured region could potentially be blank or a not-really-human-readable word, but provider would always be a human readable word. Think I should just put them both in quotes?

stitch/bindings.js
kklin: Eh, I don't know if I agree.. then you need to check the length of `machines` before indexing.

aegamesi: That's what I was thinking, but if you just do machines[0] on an empty array, it'll just set it to undefined and not cause a problem. Although it feels a little bit wrong.

kklin: I think your old approach was fine, but I don't feel that strongly about it either way. I do think there should be a length check if you go this way though.

Since @kayousterhout made the suggestion, I'll let her decide.

stitch/bindings.js
kayousterhout: I'd probably just put both in quotes, since the user could put a weird space or something in provider

stitch/bindings.js
kayousterhout: My memory is hazy but IIRC the original approach looked like:

```
let lastMachine;
loop {
  if (lastMachine == undefined) {
    lastMachine = m
  }
  // check some stuff
  lastMachine = m
}
```

I was proposing _just_ changing lastMachine to first Machine, so:
```
let firstMachine;
loop {
  if (firstMachine == undefined) {
    firstMachine = m
  }
  // check some stuff
  // Now this line isn't needed: firstMachine = m
}
```

I agree that the current approach feels wrong (and is error prone later, if someone tries to use that firstMachine reference for something else).  The old approach was also totally fine!!

aegamesi:
Made some changes, rebased master. It _should_ be good to go now? @kklin

kklin:
Oh, looks like linting failed.

aegamesi:
Hm, I was under the impression that `make check` linted too -- guess not. Anyway, fixed

PR #1087, opened 2017-09-14
init: Error when quilt-initializer.js isn't found

luise:
The `quilt init` command assumes that the initializer.js module is installed
in the $PATH as quilt-initializer.js. Before this commit, `quilt init` would
just silently exit if this was not the case. We now print a descriptive
error message instead.


cli/command/init.go
```diff
@@ -46,13 +48,29 @@ func (iCmd *Init) Run() int {
...
-	cmd := exec.Command("quilt-initializer.js")
+	executable := "quilt-initializer.js"
+	cmd := exec.Command(executable)
 
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	cmd.Stdin = os.Stdin
 
 	if err := cmd.Run(); err != nil {
+		// An ExitError indicates that the command ran but didn't complete
```
kklin: You can check for the existence of the executable directly, e.g. https://github.com/quilt/quilt/blob/master/stitch/stitch.go#L109

luise: That's so much cleaner. Fixed.

cli/command/init.go
```diff
@@ -46,13 +48,25 @@ func (iCmd *Init) Run() int {
...
 	if err := cmd.Run(); err != nil {
+		if _, err = exec.LookPath(executable); err != nil {
+			log.Errorf("%s: Make sure that "+
+				"cli/command/init/intializer.js is installed in your "+
+				"$PATH as %s. This is done automatically with "+
+				"`npm install -g @quilt/install`, but if you're "+
+				"running Quilt from source, you must set up the "+
+				"symlink manually. You can do this by executing "+
+				"`ln -s <ABS_PATH_TO_QUILT_SOURCE>/"+
+				"cli/command/init/initializer.js /usr/local/bin/%s'",
```
kklin: The closing quote for the command should be a backtick rather than single quote I think

cli/command/init.go
```diff
@@ -46,13 +48,25 @@ func (iCmd *Init) Run() int {
...
-	cmd := exec.Command("quilt-initializer.js")
+	executable := "quilt-initializer.js"
+	cmd := exec.Command(executable)
 
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	cmd.Stdin = os.Stdin
 
 	if err := cmd.Run(); err != nil {
+		if _, err = exec.LookPath(executable); err != nil {
```
kklin: I think we might as well do this check above, before even trying to run the command.

luise:
@kklin , updated!

PR #1089, opened 2017-09-15
eslint: Fix syntax for disabling the underscore dangle rule

kklin:


stitch/bindings.js
```diff
@@ -1,4 +1,3 @@
-/* eslint require-jsdoc: [1] valid-jsdoc: [1] */
```
kayousterhout: The AirBNB jslinter doesn't check either of these things (at least not in my version -- did you update it?) so the effect of removing this line is to neither warn nor error on JS doc issues.  Was this the intention / am I missing something here?  I think the right change is to change these to errors and add to the eslint config (although last I checked that would require adding some documentation to other JS files that we have)

stitch/bindings.js
```diff
@@ -1,5 +1,5 @@
 /* eslint require-jsdoc: [1] valid-jsdoc: [1] */
-/* eslint-disable no-underscore-dangle */
```
kayousterhout: Can you update the commit to state what the eslint version it's referring to is?

stitch/bindings.js
kklin: Ah, I just misinterpreted. Thanks!

kklin:
@kayousterhout updated

PR #1090, opened 2017-09-15
inspect: Move the inspect command into cli/command

kayousterhout:
The inspect code should live in cli/command, and not be a top-level
package.


cli/command/inspect.go
```diff
@@ -2,7 +2,7 @@ package command
 
 import (
 	"flag"
-	"github.com/quilt/quilt/inspect"
+	"github.com/quilt/quilt/cli/command/inspect"
```
kklin: Could you condense `cli/command/inspect/main.go` into this file? I'd recommend deleting this file, and defining the inspect `SubCommand` implementation in `cli/command/inspect/`

kayousterhout: Good idea -- done in a new commit

PR #1091, opened 2017-09-15
build: Exclude node_modules from spell check

kayousterhout:
This commit adds two more paths to node modules to the paths
excluded from the spell check (currently, the golint command
can fail if the developer has installed dependencies in
quilt-tester or the examples repository, due to spelling errors
in the dependency files).


Makefile
```diff
@@ -122,7 +122,7 @@ golint: scripts/format/format
 			golint -min_confidence .25 -set_exit_status $$package || EXIT_CODE=1; \
 		fi \
 	done ; \
-	find . \( -path ./vendor -or -path ./node_modules -or -path ./docs/build \) -prune -or -name '*' -type f -print | xargs misspell -error || EXIT_CODE=1; \
+	find . \( -path ./vendor -or -path ./node_modules -or -path ./docs/build -or -path ./quilt-tester/node_modules -or -path ./examples/*/node_modules\) -prune -or -name '*' -type f -print | xargs misspell -error || EXIT_CODE=1; \
```
kklin: Can you just ignore all `node_modules` directories?

kayousterhout: Ah I had a suspicion you'd ask that but lazily didn't want to learn how -path worked... my bad.  Fixed.

Makefile
```diff
@@ -122,7 +122,7 @@ golint: scripts/format/format
 			golint -min_confidence .25 -set_exit_status $$package || EXIT_CODE=1; \
 		fi \
 	done ; \
-	find . \( -path ./vendor -or -path ./node_modules -or -path ./docs/build \) -prune -or -name '*' -type f -print | xargs misspell -error || EXIT_CODE=1; \
+	find . \( -path ./vendor -or -path *node_modules/* -or -path ./docs/build \) -prune -or -name '*' -type f -print | xargs misspell -error || EXIT_CODE=1; \
```
kklin: nit but it seems slightly more explicit to match `*/node_modules/*` so that patterns like `not_node_modules` doesn't get matched

PR #1092, opened 2017-09-16
Rename StitchID to BlueprintID

kayousterhout:
We should rename Stitch to Blueprint (we have too many names in the code -- Stitch was the old name for the language, and now everything that's called stitch should be called blueprint).  This series of commits takes one more step in that direction by renaming StitchID BlueprintID.


kayousterhout:
Done. I wrote a short message -- tried to write a longer one explaining the backwards incompatibility but it was clunky and didn't seem helpful.  Let me know if you have a good idea for what to put, beyond the short text I left.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Up Next
 -------------
 - Check for unexpected keys in the optional argument passed to the Machine,
 Container, and Deployment constructors.
+- Rename StitchID to BlueprintID in the database.
```
kklin: I think it's important to emphasize that it's an API breaking change.

kklin: Even if we just put it in the breaking changes section

kayousterhout: Updated.  I made it clear that this is internal -- I'd think of breaking changes as things where the JS API changes.  In general I'm not sure we can expect that users can update one part of Quilt without updating everything (e.g., update a client without updating the server too).

kklin: Yeah, this is actually a more nuanced problem around how we expect users to update Quilt.

Maybe it was overkill in this case -- I just want to encourage a culture where we think hard about how changes will affect existing deployments (not saying that you don't do this!).

PR #1095, opened 2017-09-16
Complete the stitch rename

kayousterhout:


kayousterhout:
BTW one remaining question here is whether bindings.js should be moved to a dedicated JS folder -- I think there's an argument to be made for that, but I'd like to do that separately from this PR.

cli/command/run_test.go
```diff
@@ -222,7 +222,7 @@ func TestRunFlags(t *testing.T) {
 	t.Parallel()
 
 	expStitch := "blueprint"
-	checkRunParsing(t, []string{"-stitch", expStitch}, Run{blueprint: expStitch}, nil)
```
kklin: Shouldn't it be that the previous commit should rename this `-stitch` to `-blueprint`?

Also, the line you added is the same as the next line, but that's moot if you change the test to use `-blueprint`.

kayousterhout: Oops yes sorry misunderstood this! Fixed

kayousterhout:
Thanks for looking at this Kevin!

PR #1096, opened 2017-09-18
js: Force namespaces to be lowercase

HongjianZhang:
GCE doesn't accept capitalized namespaces.
In vet(), throw an error if provided namespace has capitalized letter.


docs/_GettingStarted.md
```diff
@@ -107,7 +107,7 @@ $ git clone https://github.com/quilt/nginx.git
 ```
 
 The blueprint is the `main.js` file in the nginx directory; take a look at this
-file if you'd like an to see an example of what blueprints look like.  This
+file if you'd like to see an example of what blueprints look like.  This
```
kklin: This should be in another commit since it's not really related to the namespace capitalization issue

stitch/bindings.js
```diff
@@ -220,14 +220,18 @@ Deployment.prototype.toQuiltRepresentation = function toQuiltRepresentation() {
 };
 
 /**
- * Checks if all referenced containers in connections and load balancers are
- * really deployed.
+ * Checks if namespace is lower case, and if all referenced containers
```
kklin: I think this would read a little better as "if _the_ namespace"

docs/_GettingStarted.md
```diff
@@ -107,7 +107,7 @@ $ git clone https://github.com/quilt/nginx.git
 ```
```
kklin: A couple comments on the commit message:
- We've sort of standardized on categorizing this part of the code as `js` rather than `bindings`. It doesn't really matter, but it makes `git log` a little easier to follow if it's consistent.
- The first letter after the colon in the commit title should be capitalized.
- The body should be fleshed out more, and be written in prose.

So, I'd write the commit message to be:
```
js: Force namespaces to be lowercase

GCE doesn't accept capitalized namespaces. Before this commit, if a user tried
to use a namespace with a capital letter on GCE, the daemon would log an error while
booting machines. This could be hard to debug, so this commit makes it so `quilt run`
fails when the blueprint specifies a namespace with any capitalized letters.
```

stitch/bindings.js
```diff
@@ -220,14 +220,18 @@ Deployment.prototype.toQuiltRepresentation = function toQuiltRepresentation() {
...
+ * in connections and load balancers are really deployed.
  * @private
  *
  * @param {Deployment} deployment - A deployment object.
  * @returns {void}
  */
 function vet(deployment) {
+  if (deployment.namespace !== deployment.namespace.toLowerCase()) {
+    throw new Error(`namespace "${deployment.namespace}" ` +
```
kklin: I think this error message could be more user-friendly. Something like `namespace "foo" contains uppercase letters. Namespaces must be lowercase.`. (@kayousterhout may have opinions on this)

stitch/bindings_test.js
```diff
@@ -717,14 +717,19 @@ describe('Bindings', () => {
...
+      deployment = createDeployment({ namespace: 'mynamespace' });
       expect(deployment.toQuiltRepresentation().namespace).to.equal(
-        'myNamespace');
+        'mynamespace');
     });
     it('default namespace', () => {
       expect(deployment.toQuiltRepresentation().namespace).to.equal(
         'default-namespace');
     });
+    it('should error when given namespace contains upper case letters', () => {
```
kklin: This test would be a little more appropriate in the `vet` section IMO

kklin:
Oh, looks like you need to rebase as well because @kayousterhout merged some renaming stuff this weekend.

Also, I think the author in your commit should be tweaked a bit. I see
```
HongjianZhang <zhanghongjian.1995@gmail.com>
```
I don't remember if we decided whether to use your Quilt email for git (@kayousterhout?), but your name should have a space between first and last name.

kayousterhout:
Re: email, I don't have strong feelings about which address to use (either seems fine to me)

kklin:
Either is fine with me as well

stitch/bindings.js
kayousterhout: Your suggestion looks great @kklin

blueprint/bindings_test.js
```diff
@@ -658,6 +658,12 @@ describe('Bindings', () => {
       foo = new LoadBalancer('foo', []);
       deployment.deploy([foo]);
     });
+    it('should error when given namespace contains upper case letters', () => {
+      deployment = createDeployment({ namespace: 'BadNamespace' });
+      expect(() => deployment.toQuiltRepresentation())
```
kklin: This could just be `expect(deploy)`

HongjianZhang: Made the change, and amend to the commit.

PR #1099, opened 2017-09-19
lint: Avoid redundant error checks

ejj:
The go linter updated with a new rule that checks for unnecessary
checks of error return values.  This patch updates the code to make
the linter happy.


kklin:
It's cool that the linter catches this!

PR #1101, opened 2017-09-19
`quilt init` and other docs updates

luise:
This PR should be merged together with these two:
* https://github.com/quilt/slate/pull/6
* https://github.com/quilt/nginx/pull/9


cli/command/init/prompter.js
```diff
@@ -63,6 +63,23 @@ function getSizeDescriptions(provider) {
   return sizeDescriptions;
```
kklin: In the commit message, "_as_ for" should be "_ask_ for"

cli/command/init/prompter.js
```diff
@@ -63,6 +63,23 @@ function getSizeDescriptions(provider) {
...
+  newQuestion.validate = function validate(input) {
+    if (input === '?') {
+      return helpstring;
+    }
+    if (oldValidate === undefined) {
+      return true;
+    }
+    return oldValidate(input);
+  };
+  return question;
```
kklin: Shouldn't this be `newQuestion`? Does it not matter because they're pointers to objects?

kklin: If that's the case, then I'd just get rid of `newQuestion` I think

cli/command/init/prompter.js
```diff
@@ -213,43 +233,49 @@ function credentialsPrompts(provider) {
...
+
+          // Ask this question if the credentials should be given as a file path.
+          // E.g. the path to the GCE project ID file.
+          when(answers) {
+            return keyName === consts.inputCredsPath &&
+              (!keyExists || answers[consts.credsConfirmOverwrite]);
+          },
+
+          validate(input) {
+            if (fs.existsSync(input)) return true;
```
kklin: Just a comment on Javascript in general.. I know this is how the library works and not your decision.. but returning multiple types like this makes me feel gross 🤢

cli/command/init/prompter.js
```diff
@@ -63,6 +63,23 @@ function getSizeDescriptions(provider) {
   return sizeDescriptions;
 }
 
+function questionWithHelp(helpstring, question) {
```
kklin: I think taking the `question` first makes a bit more sense. Since that matches the function name.

docs/_GettingStarted.md
```diff
@@ -26,36 +26,44 @@ window to edit and run blueprints).
...
-locally with Vagrant is currently experimental.  Contact us if you're interested
-in a cloud provider that we don't yet support.
+If you prefer to use another provider than Amazon EC2, check out
+[Cloud Provider Configuration](#cloud-provider-configuration). You'll need the
+credentials in the next step.
 
-Below, we describe how to setup Amazon EC2; refer to the
-[Cloud Providers](#cloud-provider-configuration) section if you'd like to setup
-a different cloud provider.
+## Creating an Infrastructure
```
kklin: Are you planning on using the `quilt init` stuff for your user studies? If so, we should make sure that it's properly packaged right now (and do we need to make a release?)

docs/_GettingStarted.md
```diff
@@ -185,8 +185,18 @@ bd68:
...
-subset of it, Quilt will log in to the correct machine.
+To check the logs of the same container or VM, use `quilt logs`:
+
+```console
+$ quilt logs bd68
+```
+
+If you run `logs` on the nginx container for instance, you'll see that nginx
+logs a GET request for each time you access the website. This is not thrilling
+information, but the logs will come in handy if we ever encounter any errors.
```
kklin: s/we/you ?

cli/command/init/prompter.js
luise: Yup, makes sense!

cli/command/init/prompter.js
luise: Yeah, so they're the same object, so the outcome is the same. But the linter doesn't allow modifying the function argument, so that's why we have the `newQuestion`.

cli/command/init/prompter.js
luise: I know -- it is gross!

docs/_GettingStarted.md
luise: Sure!

docs/_GettingStarted.md
luise: Yup, I'd like to use `quilt init`. I thought Ethan fixed that in the latest release, but we should definitely check!

luise:
Updated, @kklin!

PR #1102, opened 2017-09-20
util: Error on non-200 status codes

kklin:


util/util.go
```diff
@@ -24,14 +24,14 @@ var Sleep = time.Sleep
 func httpRequest(url string) (string, error) {
 	resp, err := http.Get(url)
 	if err != nil {
-		return "<error>", err
+		return "", err
```
kayousterhout: should this just be nil? (wondering if that's more consistent with elsewhere)

util/util.go
```diff
@@ -28,6 +28,9 @@ func httpRequest(url string) (string, error) {
 	}
 
 	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
```
kayousterhout: I found this commit message hard to parse.  what about: (if you feel strongly leave the existing one)

Before, we didn't check for server errors in the response code, which meant that we would (incorrectly) still use the response to form the desired cloud ACLs.

util/util.go
kklin: I'm assuming you're implying `return nil, err`, and not `return "", nil`.

You can't do `return nil, err` because `nil` isn't of type `string`. We can return `nil` in some other parts of the code because the return type is an interface or pointer.

kayousterhout: Ah k didn't realize Go doesn't allow nil for strings.

util/util.go
kklin: I had no attachment, updated!

kayousterhout: Cool looks good

PR #1103, opened 2017-09-20
tls: Don't verify server's IP address

kklin:
Before, clients would verify that the remote server's IP address matches
the one defined in the server's certificate. This was
problematic when the server's IP changed due to floating IP assignment.
Rather than generate a certificate for the new IP and force the
minion to use the new certificate, we simply ignore the hostname. This
is safe to do because the client only trusts a single CA, and we have
complete control over what certificates the CA signs.


kklin:
@HongjianZhang I updated the Changelog, if you'd like to take another look

PR #1104, opened 2017-09-21
init: Add real world location to region options

luise:
Before, we would have the user choose from the provider's region names, such
as 'us-west-1'. To make the options more helpful to users, we now also list
the real geographical region of the data center -- e.g. "N. California".


cli/command/init/prompter.js
```diff
@@ -42,25 +42,25 @@ function isNumber(input) {
...
-  * chosen provider in providers.json.
-  * The output map would have entries like {'small (m3.medium)': 'm3.medium'}
+  * Get user friendly descriptions for the key/value pairs in the given object.
+  * Given the object {'small': 'm3.medium'}, the output map would be
+  * {'small (m3.medium)': 'm3.medium'}. This function is helpful for creating
+  * descriptions for e.g. regions and instance types.
   *
-  * @param {Provider} provider The chosen Provider.
+  * @param {Object.<string, string>} info The data to make descriptions for.
   * @return {Object.<string, string>} A map from the size description to the
```
kklin: This comment is out of date

cli/command/init/prompter.js
```diff
@@ -42,25 +42,25 @@ function isNumber(input) {
...
-function getSizeDescriptions(provider) {
-  const sizeDescriptions = {};
-  const sizes = provider.getSizes();
-  const sizeCategories = Object.keys(sizes);
-
-  sizeCategories.forEach((category) => {
-    const size = sizes[category];
-    const description = `${category} (${size})`;
-    sizeDescriptions[description] = size;
+function getDescriptions(info) {
```
kklin: I find this code pretty hard to understand, and I think it's because it's converting objects into a format that's highly dependent on how the prompting code consumes it. And it's hard to clearly document the prompting code format.

What if this function converted the friendly name to value mappings into objects that you can pass directly into `choices`?
```
// getInquirerChoices converts a map of friendly names to formal values into an object compatible
// with the `choices` attribute of inquirer questions. For example, `{small: 't2.micro'}` would convert
// to `{name: 'small (t2.micro)', value: 't2.micro'}.
function getInquirerChoices(friendlyNameToValue) {
  return friendlyNameToValue.map(([friendlyName, value]) => {
    return {name: `${friendlyName} (${value})`, value};
  });
}
```

Then you could get rid of the `filter`, and it'd be obvious at a glance what the purpose of this function is, and what the expected output is.

luise: Ah, that's much better. I didn't realize the `name`/`value` magic. That's super convenient (and basically what my function and `filter` were trying to replicate). Thanks for finding this!

luise:
@kklin , updated!

cli/command/init/prompter.js
```diff
@@ -283,8 +278,10 @@ function credentialsPrompts(provider) {
   * @return {Promise} A promise that contains the user's answers.
   */
 function machineConfigPrompts(provider) {
-  const sizeChoices = getSizeDescriptions(provider);
-  sizeChoices[consts.other] = consts.other;
+  const regionChoices = getInquirerDescriptions(provider.getRegions());
+  const sizeChoices = getInquirerDescriptions(provider.getSizes());
+  sizeChoices.push({ name: [consts.other], value: consts.other });
```
kklin: Why is `consts.other` an array?

luise: My b. Evaluated values vs. non-evaluated keys.

PR #1105, opened 2017-09-21
Remove maxprice

HongjianZhang:
Before this commit, maxPrice is used to specify the upperbound for the cost
of a machine. A machine is chosen only if it fits the description and costs
less than maxPrice. However this feature is rarely used. This commit will
remove it.


blueprint/bindings.js
```diff
@@ -241,14 +241,18 @@ Deployment.prototype.toQuiltRepresentation = function toQuiltRepresentation() {
 };
```
ejj: In the commit message s/latter/letter

blueprint/bindings.js
```diff
@@ -127,14 +127,12 @@ function createDeployment(deploymentOpts) {
  * @constructor
  *
  * @param {Object} deploymentOpts - An object containing options that can tweak
- *   the behavor of the namespace.  Options include: `maxPrice` which defines
- *   the price that should be bid in spot auctions for preemptible machines,
- *   `namespace` which instructs the deployment what namespace it should
```
kklin: This is line-wrapped funny

cloud/machine/machine.go
```diff
@@ -19,24 +19,23 @@ type Description struct {
...
 	case db.Vagrant:
 		return vagrantSize(ram, cpu)
 	default:
 		panic(fmt.Sprintf("Unknown Cloud Provider: %s", provider))
 	}
 }
 
-func chooseBestSize(descriptions []Description, ram, cpu blueprint.Range,
-	maxPrice float64) string {
+func chooseBestSize(descriptions []Description, ram,
```
kklin: That line break looks a little short, could you place this entire function header on the same line? If not, breaking before the `)` would look a little cleaner IMO

cloud/machine/machine_test.go
```diff
@@ -20,23 +20,23 @@ func TestConstraints(t *testing.T) {
...
 	checkConstraint(testDescriptions, blueprint.Range{Min: 2},
-		blueprint.Range{Min: 2}, 2, "size1")
+		blueprint.Range{Min: 2}, "size1")
 
 	// Test no match
 	checkConstraint(testDescriptions, blueprint.Range{Min: 3},
-		blueprint.Range{Min: 2}, 2, "")
+		blueprint.Range{Min: 2}, "")
 
 	// Test price too expensive
```
kklin: This test should just be removed

cloud/machine/machine_test.go
```diff
@@ -45,19 +45,19 @@ func TestConstraints(t *testing.T) {
 		{Size: "size4", Price: 0.5, RAM: 3, CPU: 4},
 	}
 	checkConstraint(testDescriptions, blueprint.Range{Min: 4},
-		blueprint.Range{Min: 3}, 2, "size3")
+		blueprint.Range{Min: 3}, "size3")
 
 	// Test infinite price
```
kklin: This test should be removed

engine/engine.go
```diff
@@ -108,8 +106,7 @@ func toDBMachine(machines []blueprint.Machine, maxPrice float64,
 
 func machineTxn(view db.Database, bp blueprint.Blueprint, adminKey string) {
 	// XXX: How best to deal with machines that don't specify enough information?
-	maxPrice := bp.MaxPrice
```
kklin: The `MaxPrice` field can be removed from `blueprint/blueprint.go` (I found this by running `git grep -i maxprice`)

cloud/machine/machine_test.go
HongjianZhang: The test to be removed should be "// Test price too expensive" not "Test no match" right?

kklin: Yep

HongjianZhang:
Fixed comments and rebased against master.

blueprint/bindings.js
HongjianZhang: I don't think I have latter in this commit. I removed the previous commit that was wrongfully added to this pull request.

PR #1109, opened 2017-09-21
docs: Small tweaks

luise:
This commit just tweaks the language and clarifies a few steps in the Getting
Started guide.


kayousterhout:
@ejj I can look at this

docs/_GettingStarted.md
```diff
@@ -34,13 +33,15 @@ with one of Quilt's supported cloud providers ([Google Cloud](https://cloud.goog
 credentials for this account.
 
 For Amazon EC2, you can create an account with [Amazon Web Services](https://aws.amazon.com/ec2/)
-and then find your access credentials from the
+and then find your "Access Keys" from the
 [Security Credentials](https://console.aws.amazon.com/iam/home?#security_credential)
-page in the AWS Management Console.
+page in the AWS Management Console. If you don't have any key, you should
```
kayousterhout: This is slightly confusing because if there is a key there, the user will only be able to actually use if it they've previously downloaded it to their machine (so what it says in that page is not very helpful), and also it's not obvious that we'll describe (later) how to use the keys.  Fine to merge this as-is but let's think about how to improve in the future.

docs/_GettingStarted.md
```diff
@@ -65,9 +69,9 @@ It is also possible to use the
 time.
 
 ## Getting the Blueprint
-This section will walk you through using Quilt to run Nginx, which is an
-open-source HTTP server.  In the example, we'll use Nginx to serve a
-simple webpage. Start by downloading the blueprint using git:
+This section will walk you through how to run Nginx (an open-source HTTP server)
```
kayousterhout: I'd say "web" instead of "HTTP" for maximum approachability

docs/_GettingStarted.md
```diff
@@ -169,10 +174,11 @@ from the public internet,
 opens port 80 on the Nginx container to the outside world:
 
 ```javascript
-webTier.allowFrom(publicInternet, 80);
+webTier.allowFrom(publicInternet, port);
 ```
 
-Without this line, the website wouldn't be accessible from the browser.
+Without this line, the website wouldn't be accessible from the browser. The
+`port` variable defaults to 80, but can be changed.
```
kayousterhout: You can change the value of the `port` variable to use a port other than 80 ?

luise: Sounds good

PR #1110, opened 2017-09-22
db: Don't print deployment object in database output

kklin:
Printing the deployment object in the daemon output is too hard to read
to be useful.


aegamesi:
Looks good (and useful) to me.

PR #1112, opened 2017-09-24
join: Fix algorithm to strictly prioritize better matches

aegamesi:
Previously, join used an overly greedy algorithm that could, depending
on the order of input slice, create an 'okay' pair for an element
even if that element could also be involved in a 'perfect' pair. The
new algorithm does not necessarily return in the same order as the
previous algorithm; some tests were adjusted to account for this


join/join.go
```diff
@@ -25,60 +34,91 @@ type Pair struct {
...
+				Pair{left[scoredPair.left], right[scoredPair.right]})
+			delete(unpairedLefts, scoredPair.left)
+			delete(unpairedRights, scoredPair.right)
+		}
+	}
+
+	// Collect unpaired elements.
+	lonelyLefts = make([]interface{}, 0, len(unpairedLefts))
+	lonelyRights = make([]interface{}, 0, len(unpairedRights))
+	for i := range left {
```
luise: Given that you delete all the paired elements from the `unpairedX`, why not just loop over `unpairedX`? Then you could avoid the `if` block.

ejj:
I would like to review this as well before it's merged.

ejj:
One thing, you have an uncovered line of code.  For this sort of super algorithmic thing I like 100% coverage.

Ethan

join/join.go
```diff
@@ -25,60 +34,91 @@ type Pair struct {
...
 func Join(lSlice, rSlice interface{}, score func(left, right interface{}) int) (
 	pairs []Pair, lonelyLefts, lonelyRights []interface{}) {
 	c.Inc("Join")
 
-	val := reflect.ValueOf(rSlice)
-	len := val.Len()
-	lonelyRights = make([]interface{}, 0, len)
+	left := copySlice(lSlice)
+	right := copySlice(rSlice)
+	unpairedLefts := make(map[int]bool)
```
ejj: Would be more idiomatic for this to be a `map[int]struct{}`.  Also in general, instead of using make, we just create it directly: `map[int]struct{}{}`.

join/join.go
```diff
@@ -25,60 +34,91 @@ type Pair struct {
...
-			case s < 0:
-				continue
-			case s == 0:
-				pairs = append(pairs, Pair{l, r})
-				lonelyRights = sliceDel(lonelyRights, i)
-				continue Outer
-			case s < bestScore || bestScore < 0:
-				bestIndex = i
-				bestScore = s
+	scoredPairs := make([]ScoredPair, 0)
```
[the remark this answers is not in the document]
ejj: Yep.  Same for Pairs too `var pairs []Pair`

join/join.go
```diff
@@ -25,60 +34,91 @@ type Pair struct {
...
+// matches first, but negative scores are never matched. The algorithm does not minimize
+// the total score of all matches.
 func Join(lSlice, rSlice interface{}, score func(left, right interface{}) int) (
 	pairs []Pair, lonelyLefts, lonelyRights []interface{}) {
 	c.Inc("Join")
 
-	val := reflect.ValueOf(rSlice)
-	len := val.Len()
-	lonelyRights = make([]interface{}, 0, len)
+	left := copySlice(lSlice)
```
ejj: Why are you copying the slice?  I don't see what that gets us.  You don't edit `left` in the function ...

join/join.go
```diff
@@ -25,60 +34,91 @@ type Pair struct {
...
+	right := copySlice(rSlice)
+	unpairedLefts := make(map[int]bool)
+	unpairedRights := make(map[int]bool)
 
-	for i := 0; i < len; i++ {
-		lonelyRights = append(lonelyRights, val.Index(i).Interface())
+	for i := range left {
+		unpairedLefts[i] = true
```
ejj: Looking this over, I think the code would be generally easier to read and think about, if you inverted these maps to be `pairedLefts` and `pairedRights` for a couple of reasons.  (1) it's just easier to think about variables that are positives instead of negatives (i.e. paired is a positive, not-paired is the negation of it).  Second, I think it's more intuitive if every time you make a pair you add it to the map, instead of pre-poulating and then every time you make a pair having to remove it from the map

join/join.go
```diff
@@ -25,60 +34,91 @@ type Pair struct {
...
-				bestIndex = i
-				bestScore = s
+	scoredPairs := make([]ScoredPair, 0)
+	pairs = make([]Pair, 0, len(left))
+
+	// Generate initial list of pairs.
+OuterPairing:
+	for i := 0; i < len(left); i++ {
+		for j := 0; j < len(right); j++ {
+			if unpairedRights[j] {
```
ejj: In general, instead of having large indended code blocks, better to just check the negation and continue.  I.e.
```
if !unpairedRights[j] {
    continue
}

...
```

join/join.go
```diff
@@ -16,6 +17,14 @@ type Pair struct {
 	L, R interface{}
 }
 
+// A ScoredPair represents a candidate pair along with its associated score,
+// given by the scoring function.
+type ScoredPair struct {
```
ejj: So here, you're actually exporting ScoredPair from the join package because it's capitalized.  Minimally you should rename it `scoredPair`, but I really think that sense this is a minor implementation detail of the internals of join, it would be better to define it within the Join function.

join/join.go
```diff
@@ -25,60 +34,91 @@ type Pair struct {
...
 			}
 		}
+	}
 
-		if bestIndex >= 0 {
-			pairs = append(pairs, Pair{l, lonelyRights[bestIndex]})
-			lonelyRights = sliceDel(lonelyRights, bestIndex)
-			continue Outer
+	// Sort and collect 'best' pairs.
+	sort.Slice(scoredPairs, func(i, j int) bool {
```
ejj: `sort.SliceStable`

luise: Why does it matter whether this is stable? I didn't think the order of the input implied anything

ejj:
That's my first round of feedback.  Let me know when I should look at it again.

Ethan

join/join.go
ejj: Well if you consider that each join is likely running over and over again, having something stable may be somewhat less likely to cause us to unnecessarily oscillated between valid matches.  Also, it futures proofs the tests a bit.  Consistent results mean we're less likely to end up in a situation where a test fails intermittently (or coverage changes slightly between runs).

It's not a very big deal, but it's a trivial change, so why not?

luise: Ah, ok, that makes sense. I was just curious if there were some assumptions about Join I didn't know about.

join/join.go
aegamesi: Oh, I didn't even realize that's possible in Go.

join/join.go
aegamesi: We have to iterate over scoredPairs (which is sorted in ascending score) so that we prioritize matching lower scoring pairs first. Or is there something else that I'm missing?

luise: But these loops iterate over `left` and `right`, which have no particular order, right?

aegamesi:
@luise @ejj I believe I've addressed all of your comments

join/join.go
```diff
@@ -25,60 +26,89 @@ type Pair struct {
...
+			score := score(lVal, rVal)
+			if score == 0 {
+				// Pair immediately.
+				pairs = append(pairs, Pair{lVal, rVal})
+				pairedLefts[i] = struct{}{}
+				pairedRights[j] = struct{}{}
+
+				continue OuterPairing
+			} else if score > 0 {
+				scoredPairs = append(scoredPairs,
```
ejj: Can this fit on one line now?

PR #1113, opened 2017-09-24
cli: Add a message telling user to do 'quilt show' after 'quilt run'

aegamesi:
Previously, after a successful 'quilt run', there would be no user
feedback. This commit adds a hint telling the user to use 'quilt show'
to check the status of their deployment.


cli/command/run.go
```diff
@@ -112,6 +112,8 @@ func (rCmd *Run) Run() int {
 		return 1
 	}
 
+	fmt.Println("Your blueprint is being deployed. " +
+		"You can check its status with `quilt show`.")
```
luise: We're trying to use "you" a little less in docs and so, so how about this?
```
The blueprint is being deployed. Check its status with `quilt show`
```

luise: Or maybe somewhere in between: 
```
Your blueprint is being deployed. Check its status with `quilt show`
```

kayousterhout: Another idea that's a little more specific is `The Quilt daemon is deploying the blueprint. Check its status with ...`

luise: I'm not sure the users need to worry about the Daemon here. It's more of a "how" than "what". But I don't feel strongly about it.

aegamesi:
@luise I've made the changes you requested

PR #1116, opened 2017-09-25
Support (and test) Node.js version 6

kayousterhout:
This commit changes Quilt so that Travis tests node version 6, and so that our docs say we support it too.


kklin:
BTW @kayousterhout, you might already be tracking this somewhere but we should also:
- Update the docs on the new minimum version
- Update the integration tests to use 6.9.0.

PR #1119, opened 2017-09-26
Docs fixes based on UX studies

luise:


cli/command/init/prompter.js
```diff
@@ -282,6 +282,12 @@ function machineConfigPrompts(provider) {
   const sizeChoices = getInquirerDescriptions(provider.getSizes());
   sizeChoices.push({ name: consts.other, value: consts.other });
 
+  const sizeOptions = [{
+    name: `${consts.instanceTypeSize} (recommended)`,
+    value: consts.instanceTypeSize },
+  { name: consts.ramCpuSize },
```
HongjianZhang: Not sure if the this line needs to be indented.

luise: I thought so too! But the linter corrects it to this indentation level... I agree it looks weird though.

kklin: Maybe it'll be prettier if you put the `}` on line 287 on a new line?

kklin: The `{` on line 285 may need to go on a new line as well. dunno

HongjianZhang: Yeah. I think that would be a good idea.

luise: Agreed! Found a way that looks good and makes the linter happy. I'll update

luise: @HongjianZhang , updated! Thanks!

cli/command/init/prompter.js
```diff
@@ -356,6 +364,7 @@ function machineConfigPrompts(provider) {
         return answers[consts.sizeType] !== consts.instanceTypeSize;
       },
     },
+    'For small applications, 2 GiB is a suitable choice.'),
```
kklin: Super nit - above you have `}, 'help'`, while here you have `},
'help'`. It'd be nice to be consistent, I don't really care which one you go with

kklin:
This needs to be rebased now

luise:
@kklin , rebased!

PR #1122, opened 2017-09-26
cli: Make version print the client version even if daemon is unreachable

aegamesi:
Previously, running `quilt version` would immediately try to connect to
daemon, would would prevent any version information from being printed
if the daemon connection failed. This makes version print the client
version before attempting to fetch the daemon's version.

I'm printing out "Finding daemon version..." before attempting to connect to the daemon, then clearing the line with \r when I find the version. This has the effect of immediately showing the client and daemon version if the daemon *is* running. If it *isn't* running (or it takes a while to connect), the CLI informs the user what it is doing so it doesn't seem unresponsive.


kklin:
I'll review this

cli/command/version.go
```diff
@@ -38,14 +38,41 @@ func (vCmd *Version) Parse(args []string) error {
 
 // Run prints the version information.
 func (vCmd *Version) Run() int {
-	fmt.Println("Client:", version.Version)
+	fmt.Println("Client Version:", version.Version)
 
+	// Manually create the connection to daemon.
+	fmt.Print("Finding daemon version... ")
+	if vCmd.connectionHelper.client == nil {
```
kklin: It's kind of confusing how you're overriding `BeforeRun` to delay the connection creation. It'd be more clear to make this command embed `connectionFlags` and copy the `BeforeRun` logic from `connectionHelper` to here I think

aegamesi: It looks like there are actually a lot of things (pretty much the entire connection.go file) that I'd have to duplicate into version.go for that to work. Is there a simpler way that doesn't duplicate so much functionality? I could change the comment before `BeforeRun` to indicate that, too.

aegamesi:
@kklin

cli/command/version.go
```diff
@@ -38,14 +41,52 @@ func (vCmd *Version) Parse(args []string) error {
 
 // Run prints the version information.
 func (vCmd *Version) Run() int {
-	fmt.Println("Client:", version.Version)
+	fmt.Println("Client Version:", version.Version)
 
+	// Manually create the connection to daemon.
+	fmt.Print("Finding daemon version... ")
+
+	err := vCmd.connectToDaemon()
```
kklin: These two lines could just be `if err := vCmd.connectToDaemon() ; err != nil {`. This comment might be moot if you get rid of the `client` instance variable though

cli/command/version.go
```diff
@@ -38,14 +41,52 @@ func (vCmd *Version) Parse(args []string) error {
 
 // Run prints the version information.
 func (vCmd *Version) Run() int {
-	fmt.Println("Client:", version.Version)
+	fmt.Println("Client Version:", version.Version)
 
+	// Manually create the connection to daemon.
+	fmt.Print("Finding daemon version... ")
```
kklin: Eh, I don't like that so much logic below relates to how you might need to clear this line. E.g. it's easy to forget to add the `fmt.Println` if you're logging an error.

What if you handled the logging in the function that gets the client?

```
func (vCmd *Version) getDaemonClient() (client.Client, error) {
  fmt.Print("Connecting to daemon...")
  defer fmt.Print("\r                         \r")

  creds, err := credentials.Read(vCmd.connectionFlags.tlsDir)
  if err != nil {
    return nil, err
  }
  return client.New(vCmd.host, creds)
}
```

And the caller would log the error if necessary.

cli/command/version.go
```diff
@@ -38,14 +41,52 @@ func (vCmd *Version) Parse(args []string) error {
...
+
+	err := vCmd.connectToDaemon()
+	if err != nil {
+		fmt.Println()
+		log.WithError(err).Error("Failed to connect to daemon")
+		return 1
+	}
```
kklin: A newline here would make the code slightly easier to read IMO

cli/command/version.go
kklin: Also, you could generate the string that clears the "Connecting to daemon" message at runtime based on the string's length.. Then if you changed that message, you wouldn't need to change the `defer fmt.Print`. Might be overkill, but I think it's cleaner.

cli/command/version.go
```diff
@@ -7,12 +7,15 @@ import (
...
 
+	"github.com/quilt/quilt/api/client"
+	"github.com/quilt/quilt/cli/command/credentials"
 	log "github.com/sirupsen/logrus"
 )
 
 // Version prints the Quilt version information.
 type Version struct {
-	connectionHelper
+	client client.Client
```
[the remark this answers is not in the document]
aegamesi: It does feel a bit strange to have it as an instance variable since it's only directly used in `Run`, but leaving it as an instance variable allows `version_test.go` to easily inject a mocked version of `client.Client`: https://github.com/aegamesi/quilt/blob/version-without-daemon/cli/command/version_test.go#L31 -- do you think there's a better way to do this?

cli/command/version.go
aegamesi: I think the main issue with doing that is that connecting isn't the only operation that might take a while. I think in the most common situation (the daemon isn't running, and the time it takes to connect is the time it takes to timeout) it is, but actually getting the remote daemon version might take some time so I think it's worth keeping the progress message for fetching the version.
Also, I specifically wanted to keep the 'fetching daemon version' message if there's an error, so there's at least an explanation of *why* the CLI was trying to connect to the daemon if it fails (rather than just saying "Client version: blahblah ... connecting to daemon failed: <error message>". I might just be overthinking it though (and perhaps it's not so unclear anyway). 
--
Yeah, I guess it's reasonable to generate the clearing message at runtime.

aegamesi:
@kklin I've responded to your comments -- haven't made any changes yet though

cli/command/version.go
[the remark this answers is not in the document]
kklin: You could pull the code that uses the `client.Client` into a different method, and unit test that method by passing the mock client in as a parameter.

Or you could the Version struct have a field for `client.Getter`, and use `client.New` in `NewVersionCommand()`, but use a custom function when unit testing. I think I prefer this approach.. although both are ugly.

aegamesi:
@kklin made some more changes and updated to work with Kelda / the new TLS stuff.

The code around printing and clearing that line is a little bit gross looking, but refactoring it into a separate generic function doesn't make it cleaner.

The whole `credsGetter` thing is just to make Unit testing possible without trying to read credentials from the disk. Not sure if the way I did it is the canonical Go / Kelda way though.

cli/command/version.go
```diff
@@ -7,12 +7,17 @@ import (
 	"github.com/kelda/kelda/util"
 	"github.com/kelda/kelda/version"
 
+	"github.com/kelda/kelda/api/client"
+	cliPath "github.com/kelda/kelda/cli/path"
+	"github.com/kelda/kelda/connection/tls"
+	tlsIO "github.com/kelda/kelda/connection/tls/io"
 	log "github.com/sirupsen/logrus"
+	"strings"
```
kklin: Imports should be grouped approximately as follows:
```
stdlib

kelda imports

3rd party imports
```

so, `strings` should be moved to the first section, and the kelda imports should be moved to the second section.

cli/command/version.go
```diff
@@ -38,14 +43,46 @@ func (vCmd *Version) Parse(args []string) error {
 
 // Run prints the version information.
 func (vCmd *Version) Run() int {
-	fmt.Println("Client:", version.Version)
+	fmt.Println("Client Version:", version.Version)
+
+	fetchingVersionStr := "Fetching daemon version..."
+	fmt.Print(fetchingVersionStr)
+	daemonVersion, err := vCmd.fetchDaemonVersion(func() (tls.TLS, error) {
```
kklin: This anonymous function is pretty ugly -- I'd modify the signature so that it could take `tlsIO.ReadCredentials` (i.e. take the path as a parameter) to avoid the anonymous function.

cli/command/version.go
```diff
@@ -38,14 +43,46 @@ func (vCmd *Version) Parse(args []string) error {
...
 }
+
+type credsGetter func() (tls.TLS, error)
+
+func (vCmd *Version) fetchDaemonVersion(credsGetter credsGetter,
+	getter client.Getter) (string, error) {
+	creds, err := credsGetter()
+	if err != nil {
+		return "", err
+	}
```
kklin: I think a newline here would make this easier to read

cli/command/version.go
```diff
@@ -38,14 +43,46 @@ func (vCmd *Version) Parse(args []string) error {
...
-	fmt.Println("Daemon:", daemonVersion)
 
+	fmt.Println("Daemon Version:", daemonVersion)
 	return 0
 }
+
+type credsGetter func() (tls.TLS, error)
+
+func (vCmd *Version) fetchDaemonVersion(credsGetter credsGetter,
+	getter client.Getter) (string, error) {
```
kklin: Naming this `clientGetter` would make the code a little more obvious given that you have a `credsGetter` here now as well now.

aegamesi:
@kklin ready to review again

PR #1124, opened 2017-09-26
cloud: Set "Booting" status in the main join() transaction

ejj:
Before this patch, the machine booting status was set outside of the
main join transaction.  This was not technically safe as the namespace
could have changed before that second transaction which could corrupt
the database.   Furthermore, the code was unnecessarily complicated,
and won't work with some upcoming cloud package changes.

### DIFF
--- a/cloud/status.go
+++ b/cloud/status.go
@@ -1,13 +1,9 @@
 package cloud
 
 import (
-	"errors"
-
 	"github.com/quilt/quilt/cloud/foreman"
 	"github.com/quilt/quilt/db"
 	"github.com/quilt/quilt/util"
-
-	log "github.com/sirupsen/logrus"
 )
 
 func updateMachineStatuses(conn db.Conn) {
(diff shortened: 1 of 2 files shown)


cloud/status.go
```diff
@@ -20,9 +16,8 @@ func updateMachineStatuses(conn db.Conn) {
 func updateMachineStatusesOnce(conn db.Conn) {
 	conn.Txn(db.MachineTable).Run(func(view db.Database) error {
 		for _, dbm := range view.SelectFromMachine(nil) {
-			// Don't touch machines that are booting. `clst.boot` will take
-			// care of unsetting the status when it's no longer booting.
-			if dbm.Status == db.Booting {
+			// Don't touch machines that are still booting.
```
kayousterhout: Can you add to this comment to explain why it's not safe to check the Status?

ejj: Done (though a patch very soon is going to remove all of this code, so it's moot).

kayousterhout: Ah I was wondering about that and squelched a bunch more comment requests as a result :).  Thanks for updating this one!

PR #1126, opened 2017-09-27
ci: Add test for floating IPs

kklin:
This commit adds a test that assigns a floating IP to a random worker,
and deploys a Nginx container to that IP. We then check whether we can
successfully access the page through the floating IP.


integration-tester/tests/floating-ip/floating_ip_test.go
```diff
@@ -0,0 +1,46 @@
...
+		t.Fatalf("couldn't get api client: %s", err)
+	}
+	defer clnt.Close()
+
+	machines, err := clnt.QueryMachines()
+	if err != nil || len(machines) == 0 {
+		t.Fatalf("couldn't query machines: %s", err)
+	}
+
+	provider := machines[0].Provider
```
aegamesi: The JavaScript portion looks for the first worker machine, but this portion just picks the first machine found. Since all that seems to matter is the provider, and Quilt currently doesn't support multiple providers in a single deployment, there isn't any issue. However, if in the future Quilt *does* support multiple providers, could this test break?

kklin: It would break, but that's so far off we don't really know how it would look, so it's not worth accounting for that possibility now IMO.

PR #1128, opened 2017-09-28
Don't allow unencrypted GRPC connections

kklin:
Before, if no `tls-dir` was specified, the traffic between the daemon and cluster was unencrypted. This was obviously insecure, and blocks the implementation of secrets. This series of commits forces users to use TLS, and makes various UX improvements so that the user workflow doesn't change:  users can still just run all the quilt commands such as `quilt daemon` and `quilt run foo` without adding additional flags, or running setup commands.


kklin:
This PR will also need to be reviewed: https://github.com/quilt/tester/pull/54

kayousterhout:
@HongjianZhang @kklin I can review this

CHANGELOG.md
```diff
@@ -9,8 +9,9 @@ parent does not exist.
 QUILT_TLS_DIR.
 - Require TLS credentials to connect to the daemon when it's running with TLS
 enabled. Before, users could connect insecurely even if TLS was enabled.
-- Always encrypt communication between the daemon and cluster. `quilt setup-tls`
-must be run before using `quilt daemon` for the first time.
+- Always encrypt communication between the daemon and cluster.
+- Auto-generate TLS credentials when starting the daemon if the provided
```
kayousterhout: what does it mean "if the provided credentials don't exist"? Should it be "if no credentials are provided"?

cli/command/daemon.go
```diff
@@ -67,6 +68,16 @@ func (dCmd *Daemon) AfterRun() error {
 func (dCmd *Daemon) Run() int {
 	log.WithField("version", version.Version).Info("Starting Quilt daemon")
 
+	// If the specified TLS credential path does not exist, autogenerate
+	// credentials for the given path.
+	if _, err := util.Stat(dCmd.tlsDir); os.IsNotExist(err) {
+		log.WithField("path", dCmd.tlsDir).Info("Auto-generating TLS credentials")
+		if exitCode := (&SetupTLS{dCmd.tlsDir}).Run(); exitCode != 0 {
+			log.Error("TLS credential generation failed")
```
kayousterhout: Can we include the error from SetupTLS here?  Otherwise I'm worried we're not telling the user enough about what the problem was.

cli/command/daemon.go
```diff
@@ -67,6 +68,16 @@ func (dCmd *Daemon) AfterRun() error {
 func (dCmd *Daemon) Run() int {
 	log.WithField("version", version.Version).Info("Starting Quilt daemon")
 
+	// If the specified TLS credential path does not exist, autogenerate
```
kayousterhout: Why not remove the setup-tls command entirely as part of this commit?

cli/command/daemon.go
```diff
@@ -78,11 +76,9 @@ func (dCmd *Daemon) Run() int {
 				"Failed to parse private key %s", dCmd.adminSSHPrivateKey)
 			return 1
 		}
-	}
-
-	if sshKey == nil && dCmd.tlsDir != "" {
-		log.Info("No admin key supplied, but TLS is enabled, which requires an " +
-			"admin SSH key to copy TLS credentials to the cluster. " +
+	} else {
+		log.Info("No admin key supplied, which is required " +
```
kayousterhout: Are there any side-effects here the user needs to know about (or we should think about from a UX perspective)? E.g., does this mean if the daemon dies, it can't be re-started, because a new one will generate a new key and thus not be able to connect to the old cluster?

cloud/cfg/cfg.go
```diff
@@ -58,16 +53,11 @@ func Ubuntu(m db.Machine, inboundPublic string) string {
 	return cloudConfigBytes.String()
 }
 
-func minionOptions(role db.Role, inboundPublic, tlsDir string) string {
-	options := fmt.Sprintf("--role %q", role)
+func minionOptions(role db.Role, inboundPublic string) string {
+	options := fmt.Sprintf("--role %q --tls-dir %q", role, tlsIO.MinionTLSDir)
```
kayousterhout: Why not have the minion code directly use tlsIO.MinionTLSDir, rather than passing it as a command line arg?

docs/_Security.md
```diff
@@ -3,20 +3,20 @@
 ## TLS
 Quilt uses [grpc](http://www.grpc.io/) for communication with the daemon and
 deployed clusters. Functionality exposed through `grpc` include deploying new
-blueprints and querying deployment information. Thus, TLS should be enabled for
-all non-experimental deployments. It is currently disabled by default.
+blueprints and querying deployment information. Thus, TLS is automatically
```
kayousterhout: I'd just say ...deployment information, and all communication is automatically encrypted and verified using TLS." (not sure if that's the right use of TLS?).  In the current version, it's not totally obvious why it's "Thus, ..."

docs/_Security.md
```diff
@@ -3,20 +3,20 @@
...
+enabled for all deployments.
 
 ### Quickstart
 Generate the necessary TLS files.
 
 ```console
-$ quilt setup-tls ~/.quilt/tls
+$ quilt setup-tls
```
kayousterhout: Does this need to be in the quick start anymore?

integration-tester/tests/connection-credentials/credentials_test.go
```diff
@@ -32,29 +45,17 @@ func TestCredentials(t *testing.T) {
 
 	tlsDir := os.Getenv("QUILT_TLS_DIR")
 	if tlsDir == "" {
-		testInsecure(t, machines)
-	} else {
-		testTLS(t, machines, tlsDir)
+		t.Fatal("No QUILT_TLS_DIR specified. The test cannot continue without " +
+			"the TLS credentials used to interact with the cluster")
```
kayousterhout: why not use the default here, and not use the env var at all in the tests?

kklin:
I don't think the code is that complicated (at least the change to remove the environment variable and command line flag would just be deleting ~10 LOC and replacing them with the constant). I do agree that removing the env variable and flag would make the UX less complicated -- specifically, I'm worried about users having two credentials, and trying to connect with the wrong one.

The only current use case is for Jenkins, because the default credential location is in the home directory, and I'd like to have the Jenkins runs not create files outside its workspace.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Quilt Change Log
 
 Up Next
 -------------
+- Fix a bug where `quilt setup-tls` would fail when writing to a directory whose
+parent does not exist.
```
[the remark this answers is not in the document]
kklin: I dunno, I think it's helpful as a user to know when a bug you care about gets fixed. Plus, when we make a release we have a section in the beginning that details the major changes in prose, so I don't think the changes would go unnoticed.

cli/command/connection.go
```diff
@@ -16,7 +17,7 @@ type connectionFlags struct {
 
 func (cf *connectionFlags) InstallFlags(flags *flag.FlagSet) {
 	flags.StringVar(&cf.host, "H", api.DefaultSocket, "the host to connect to")
-	flags.StringVar(&cf.tlsDir, "tls-dir", "",
+	flags.StringVar(&cf.tlsDir, "tls-dir", os.Getenv("QUILT_TLS_DIR"),
```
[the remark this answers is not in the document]
kklin: I removed the ability to have multiple TLS credentials, which makes this moot

cli/command/connection.go
```diff
@@ -17,7 +17,12 @@ type connectionFlags struct {
 
 func (cf *connectionFlags) InstallFlags(flags *flag.FlagSet) {
 	flags.StringVar(&cf.host, "H", api.DefaultSocket, "the host to connect to")
-	flags.StringVar(&cf.tlsDir, "tls-dir", os.Getenv("QUILT_TLS_DIR"),
+
+	defaultTLSDir := os.Getenv("QUILT_TLS_DIR")
```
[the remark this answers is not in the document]
kklin: I removed the ability to have multiple TLS credentials, which makes this moot

cli/command/connection.go
```diff
@@ -17,7 +17,12 @@ type connectionFlags struct {
...
+	if defaultTLSDir == "" {
+		defaultTLSDir = tlsIO.DefaultTLSDir()
+	}
+	flags.StringVar(&cf.tlsDir, "tls-dir", defaultTLSDir,
```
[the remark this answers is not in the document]
kklin: I removed the ability to have multiple TLS credentials, which makes this moot

cli/command/setupTLS.go
```diff
@@ -32,13 +31,9 @@ func (sCmd *SetupTLS) InstallFlags(flags *flag.FlagSet) {
 
 // Parse parses the command line arguments for the setupTLS command.
 func (sCmd *SetupTLS) Parse(args []string) error {
-	if sCmd.outDir == "" {
-		if len(args) == 0 {
-			return errors.New("no output directory specified")
-		}
+	if len(args) != 0 {
```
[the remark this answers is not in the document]
kklin: I removed the ability to have multiple TLS credentials, which makes this moot

CHANGELOG.md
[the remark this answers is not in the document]
kayousterhout: Cool that seems reasonable

cloud/cfg/cfg.go
kklin: I removed the ability to have multiple TLS credentials, which makes this moot

docs/_Security.md
kklin: I removed the ability to have multiple TLS credentials, which makes this moot

docs/_Security.md
kklin: I thought it sounded a bit more natural with a period:

>>> Quilt uses [grpc](http://www.grpc.io/) for communication with the daemon and
deployed clusters. Functionality exposed through `grpc` include deploying new
blueprints and querying deployment information. All communication is
automatically encrypted and verified using TLS.

Let me know what you think.

cli/command/daemon.go
kklin: You're right that this could be improved a lot. Basically if the machines come up enough for the daemon to SSH in and install the certificates its fine, since even if you restart the daemon, the machines will be running with the right certificates. But if you kill the daemon after the machines have been booted, but before the keys have been SCP'd over, you won't be able to connect to the cluster (barring some ugly manual `admin-ssh-private-key` stuff).

The right thing to do here is to generate and save a SSH key pair to disk, and remove the `admin-ssh-private-key` flag I think. Do you think it should happen as part of this PR?

kklin: The more I think about it, the more I think we should just do it now since debugging this would be really confusing to a user.

cli/command/daemon.go
kklin: We include it now that I moved `SetupTLS.Run` into a helper function.

connection/credentials/tls/io/io.go
```diff
@@ -97,3 +97,8 @@ func signedCertPath(dir string) string {
 func signedKeyPath(dir string) string {
 	return filepath.Join(dir, signedKeyFilename)
 }
+
+// DefaultTLSDir returns the default location for users to store TLS credentials.
+func DefaultTLSDir() string {
+	return filepath.Join(os.Getenv("HOME"), ".quilt", "tls")
```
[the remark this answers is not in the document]
kklin: Yes, on POSIX machines https://unix.stackexchange.com/questions/123858/is-the-home-environment-variable-always-set-on-a-linux-system

CHANGELOG.md
kklin: It means that if I provide the path `~/.quilt/my-tls`, and that directory doesn't exist, the daemon will generate credentials and place them there. If no path is provided, the path will default to `~/.quilt/tls`.

This is kind of moot now that we're only supporting one TLS path, but leaving it made the code easier to write.

kklin: We should talk about this in person because one design for this could allow us to remove the need to set SSH keys in the common case altogether.

docs/_Security.md
kayousterhout: this looks great (super minor comment: include -> includeS)

connection/credentials/tls/io/io.go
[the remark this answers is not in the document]
kayousterhout: Do you think this is worth adding a comment? (I have a slight preference for a comment to make future-editor-of-this-code's life easier but don't feel strongly)

kklin:
@kayousterhout ready for more review!

docs/_Security.md
```diff
@@ -32,28 +32,18 @@ This can be simulated by restarting the daemon and running the same blueprint,
 but with different TLS credentials. Note that the machines never connect.
```
kayousterhout: I'd be in favor of cutting this or moving it to a FAQ -- it seems distracting here now that the user would have to do so much work to reset the credentials

cli/command/daemon.go
```diff
@@ -174,3 +157,25 @@ func setupTLS(outDir string) error {
 
 	return nil
 }
+
+// setupSSHKey generates a new RSA key for use with SSH, and writes it to disk.
+func setupSSHKey(outPath string) error {
+	if err := util.AppFs.MkdirAll(filepath.Dir(outPath), 0700); err != nil {
+		return fmt.Errorf("failed to create output directory: %s", err)
```
kayousterhout: can you include the dir path here?

cli/command/daemon.go
```diff
@@ -80,27 +78,21 @@ func (dCmd *Daemon) Run() int {
...
-	} else {
-		log.Info("No admin key supplied, which is required " +
-			"to copy TLS credentials to the cluster. " +
-			"Auto-generating an in-memory key.")
-		var err error
-		sshKey, err = newSSHPrivateKey()
-		if err != nil {
-			log.WithError(err).Error("Failed to generate SSH key")
+	if _, err := util.Stat(cliPath.DefaultSSHKeyPath); os.IsNotExist(err) {
+		log.Info("Auto-generating Quilt SSH key")
```
kayousterhout: Similar to the TLS comment -- how about something that includes the path, e.g., "No SSH key found at XX.  Auto-generating one and storing it at that location."

cli/command/daemon.go
```diff
@@ -80,27 +78,21 @@ func (dCmd *Daemon) Run() int {
...
+		if err := setupSSHKey(cliPath.DefaultSSHKeyPath); err != nil {
+			log.WithError(err).Error("SSH key generation failed")
 			return 1
 		}
 	}
 
+	sshKey, err := parseSSHPrivateKey(cliPath.DefaultSSHKeyPath)
+	if err != nil {
+		log.WithError(err).WithField("path", cliPath.DefaultSSHKeyPath).Error(
+			"Failed to parse private key")
```
kayousterhout: + from path x

cli/command/daemon_test.go
```diff
@@ -89,3 +79,14 @@ func TestSetupTLS(t *testing.T) {
 	_, err = tlsIO.ReadCredentials("tls")
 	assert.NoError(t, err)
 }
+
+// Test that the generated file can be parsed.
+func TestSetupSSHKey(t *testing.T) {
+	util.AppFs = afero.NewMemMapFs()
+
+	err := setupSSHKey("ssh_key")
```
kayousterhout: use variable for "ssh_key"?

cli/path/path.go
```diff
@@ -11,4 +11,8 @@ var (
 
 	// DefaultTLSDir is the default location for users to store TLS credentials.
 	DefaultTLSDir = filepath.Join(quiltHome, "tls")
+
+	// DefaultSSHKeyPath is the default location for the SSH key used for
```
kayousterhout: can you say "the default directory where the private SSH keys for accessing Quilt will be stored."?

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ parent does not exist.
 - Auto-generate TLS credentials when starting the daemon if the provided
 credentials don't exist.
 - Always encrypt communication with the daemon.
+- Try using the Quilt-managed SSH key when connecting to machines. The
```
kayousterhout: Do you think it's worth specifying that this will be the first path that gets tried?

kayousterhout:
BTW one thing that makes me slightly nervous is that we'll update the docs to say TLS is always on, but it actually isn't until our next release.  Makes me think we shouldn't wait too long before versioning the docs.

cli/command/daemon.go
kklin: I'm already logging `WithField("path")` here

CHANGELOG.md
kklin: What do you mean it's the first path that gets tried?

cli/path/path.go
kklin: That's not quite correct since the ssh key is just a file, not a directory. Happy to improve the comment if you have other ideas though!

CHANGELOG.md
kayousterhout: I meant that Quilt will first try to use the Quilt-managed SSH key, and only if that doesn't work will it try the user's other existing SSH keys

cli/path/path.go
kayousterhout: Ah ok can you just add the word "private" before SSH here then?  Or maybe "is the default filepath where the private SSH key used to access Quilt will be stored."

docs/_Security.md
kklin: I don't know if this belongs in a FAQ because I'm just trying to make it clear that only the correct credentials can connect to the cluster.

Do you think it maybe belongs somewhere else on this page?

kayousterhout: Ah I see I'd make that clear by just changing the first sentence of this paragraph to exactly what you said above:

"Only a user with access to the correct credentials can connect to the cluster.  As an example of this, if you delete your credentials, restart the daemon, and run the same blueprint, you won't be able to connect to the machines:"

CHANGELOG.md
kklin: I don't think that's important, I consider that just a detail of the implementation. Plus, the ssh-agent keys actually get tried first right now.

kklin:
@kayousterhout updated. Can you review the tester PR as well?

PR #1130, opened 2017-09-29
blueprint: Rename reference to stitch

kklin:
The recent rename of stitch to blueprint forgot to rename a test file.


kayousterhout:
Thanks for fixing this! Sorry I missed it.

PR #1132, opened 2017-10-02
Django updates

kayousterhout:
This should be merged after https://github.com/quilt/django/pull/10


integration-tester/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "license": "MIT",
   "dependencies": {
-    "@quilt/django": "quilt/django",
+    "@quilt/django": "quilt/django#d831932d3761a",
```
kklin: Note -- I think this hash will change if any changes are made in the accompanying PR, so the reviewer should make sure that this hash points to the expected commit before merging.

kayousterhout: @kklin do you think this is a bad idea? I was on the fence since using the commit hash is brittle, but it also seems broken that we need to synchronize changes like this.  In the long term we should publish to npm but in the short term I thought this would be a marginal improvement.  Curious what you think.

kayousterhout:
The commit hash is out of date again; I'll update when the other PR is finalized

kayousterhout:
Commit hash is good now

PR #1133, opened 2017-10-03
machine: Migrate chooseSize and chooseRegion from go to javascript

HongjianZhang:
Before this commit, chooseSize is implemented in go,
and called in toDBMachine in engine.go. This commit
moves this functionality to bindings.js, so that if
the user's configuration has problems, quilt can throw an error
earlier while creating the blueprint machine in js.


luise:
@HongjianZhang , it seems like there are some linting errors. If you run `make jslint` you can see those locally -- and you can run the underlying `eslint` command with the `--fix` option to automatically fix some of the errors.

blueprint/bindings.js
```diff
@@ -10,6 +10,17 @@ const os = require('os');
 // Use `let` to enable mocking of `fs` in tests.
 let fs = require('fs'); // eslint-disable-line prefer-const
 
+var descriptions = require('./descriptions');
+
+var amazonDescriptions = descriptions.amazonDescriptions;
```
kayousterhout: I'd just directly use descriptions.amazonDescriptions (and the other ones) in the places you need them below, rather than assigning them to variables here

blueprint/bindings.js
```diff
@@ -475,6 +486,10 @@ function boxRange(x) {
   return x;
 }
 
+function accepts(x, r) {
```
kayousterhout: Can you make this a function of the Range class, perhaps called `inRange`, so you'd call `r.inRange(x)`?

blueprint/bindings.js
```diff
@@ -621,6 +636,109 @@ function getBoolean(argName, arg) {
 }
 
 /**
+ * Chooses instance size if not specifed. If user specifies size, CPU and RAM, 
+ * verifies the specified size meets CPU and RAM requirements.
```
kayousterhout: I'd expand this a bit to something like:
```
Chooses an instance size (e.g., m2.xlarge) if the user did not specify one.  If the user did specify a 
size, this function verifies that the specified size satisfies any CPU and RAM requirements that the
 user specified (and throws an error if it doesn't).
```

blueprint/bindings.js
```diff
@@ -621,6 +636,109 @@ function getBoolean(argName, arg) {
...
+ * @private
+ * @param {string} size - User specified size ('' if not specified).
+ * @param {string} provider - Used to direct to corresponding descriptions.
+ * @param {Range} ram - User specified RAM requirement.
+ * @param {Range} cpu - User specified CPU requirement.
+ * @param {boolean} preemptible - Whether required machine can be preempted.
+ * @returns {string} The best size that fits the user's requirements if provider 
+ * is available in Quilt, otherwise throws an error.
+ */
+function chooseSize(size, provider, ram, cpu, preemptible) {
```
kayousterhout: Can you make this a method of the Machine class? Then you don't need to pass in all of these parameters, and can grab them directly from the instance.

blueprint/bindings.js
```diff
@@ -621,6 +636,109 @@ function getBoolean(argName, arg) {
...
+}
+
+/**
+ * Chooses provider's default region if region is not specified by the user.
+ * @private
+ * @param {string} provider - Used to direct to corresponding descriptions.
+ * @returns {string} The default region for the given provider if provider 
+ * is available in Quilt, otherwise throws an error.
+ */
+function chooseRegion(provider) {
```
kayousterhout: Similar to the chooseSize method, I'd make this a method of the Machine class (and move the check for whether the provider is blank to within this method)

blueprint/bindings.js
```diff
@@ -621,6 +636,109 @@ function getBoolean(argName, arg) {
...
+        return size;
+      } else {
+        throw new Error(`Requested size '${size}' does not meet ` +
+                        `RAM '(${ram.min}, ${ram.max})' or ` + 
+                        `CPU '(${cpu.min}, ${cpu.max})' requirements. ` +
+                        `Instance RAM: '${description.RAM}', ` +
+                        `Instance CPU: '${description.CPU}'`);
+      }
+    }
+    if (size == '') {
```
kayousterhout: It's confusing that the logic to check that the size meets the given requirements is grouped with the logic to choose the best size.   Can you move this into a separate function, e.g., "validateSize", that gets called only when the size is specified, and then only call chooseSize when the size isn't specified?

Also, did this logic exist in the old code (from a skim, it doesn't look like it does)?  If it wasn't part of the old code, it's great to add it, but can you do it in a separate commit?  That makes it easier for future people to see which commits added functionality versus just moved around existing functionality.

blueprint/bindings.js
```diff
@@ -621,6 +636,109 @@ function getBoolean(argName, arg) {
...
+          (bestSize == "" || description.Price < bestPrice)) {
+        bestSize = description.Size
+        bestPrice = description.Price
+      }
+    }
+  }
+  return bestSize
+}
+
+function vagrantSize(ramRange, cpuRange) {
```
kayousterhout: I realize this naming was inherited from the old code, but can you call this chooseBestVagrantSize, for consistency with the other function?

blueprint/bindings.js
```diff
@@ -689,6 +807,21 @@ function Machine(optionalArgs) {
   this.preemptible = getBoolean('preemptible', optionalArgs.preemptible);
 
   checkExtraKeys(optionalArgs, this);
+
+  // Throws an error if user specifies a size that can not be preemtible,
+  // but the user sets the preemptible flag True.
+  if (this.preemptible && this.size != ''
```
kayousterhout: (see other comment re: moving this to its own commit)

blueprint/bindings.js
```diff
@@ -689,6 +807,21 @@ function Machine(optionalArgs) {
...
+  // Throws an error if user specifies a size that can not be preemtible,
+  // but the user sets the preemptible flag True.
+  if (this.preemptible && this.size != '' 
+      && nonPreemptibleSizes[this.provider].includes(this.size)) {
+    throw new Error(`Requested size ${this.size} can not be preemptible`);
+  }
+
+  this.size = chooseSize(this.size, this.provider, this.ram, this.cpu, this.preemptible);
+  if (this.size == '') {
+    throw new Error(`No valid size for given requirements.`);
```
kayousterhout: Would it be simpler to throw this Error from inside chooseSize?

blueprint/bindings_test.js
```diff
@@ -18,6 +18,12 @@ describe('Bindings', () => {
     deployment = b.createDeployment();
   });
 
+  const checkSizes = function checkSizes(size, description, ram, cpu,
```
kayousterhout: Can you write this like the tests below, where they call toQuiltRepresentation and then check the JSON, to avoid needing to export the chooseBestSize method?

blueprint/bindings_test.js
```diff
@@ -48,6 +54,46 @@ describe('Bindings', () => {
...
   };
 
+  describe('Size', () => {
+    const testDescriptions = [{Size: "size1", Price: 2, RAM: 2, CPU: 2}];
+    const testDescriptionsMultiple = [
+      {Size: "size2", Price: 2, RAM: 8, CPU: 4},
+      {Size: "size3", Price: 1, RAM: 4, CPU: 4},
+      {Size: "size4", Price: 0.5, RAM: 3, CPU: 4},
+    ]
+    it('all constraints specified with valid price', () => {
```
kayousterhout: I know you inherited this description from the old code, but there's no price here (I think this should just be 'all constraints specified')?

blueprint/bindings.js
```diff
@@ -689,6 +807,21 @@ function Machine(optionalArgs) {
...
   checkExtraKeys(optionalArgs, this);
+
+  // Throws an error if user specifies a size that can not be preemtible,
+  // but the user sets the preemptible flag True.
+  if (this.preemptible && this.size != '' 
+      && nonPreemptibleSizes[this.provider].includes(this.size)) {
+    throw new Error(`Requested size ${this.size} can not be preemptible`);
+  }
+
+  this.size = chooseSize(this.size, this.provider, this.ram, this.cpu, this.preemptible);
```
kayousterhout: Also, I think this should go in the `toQuiltRepresentation` function.  Users may set things about a machine after creating it, e.g.,:

```
const m = new Machine();
m.CPU = new Range(5, 8);
```

which will cause the size to be set incorrectly with the current code.  toQuiltRepresentation is not called until after all of the users code executes, so it's definitely safe to set the size there.

engine/engine.go
```diff
@@ -69,14 +68,6 @@ func toDBMachine(machines []blueprint.Machine, adminKey string) []db.Machine {
 		m.Size = blueprintm.Size
 		m.Preemptible = blueprintm.Preemptible
 
-		if m.Size == "" {
-			m.Size = cloud.ChooseSize(p, blueprintm.RAM, blueprintm.CPU)
```
kayousterhout: Since you've removed this code, do the RAM and CPU variables still need to be in the blueprint class in go?  I suspect these can be removed from the JSON that gets passed between the JavaScript and Go code as well.

blueprint/bindings.js
kayousterhout: @kklin I wanted to run this by you to make sure you agree here re: the chooseSize functionality happening in toQuiltRepresentation().

blueprint/bindings.js
ejj: FWIW I might also call the internal varianle `amazon` instead of `amazonDescriptions`, so an importer will access it by doing `descriptions.amazon`

ejj:
Btw the patch has a bunch of trailing whitespace that should be removed.  You can tell git to highlight it in red when you do a `git show`

blueprint/bindings.js
kklin: The drawbacks I can think of if you delay converting the size is:
1) The stacktrace would be more confusing if `chooseSize` throws an error.
2) The resulting machine can't be used with placement constraints based on size (I'm less worried about this because we haven't put much design effort into placement, and we may support placement in terms of RAM and CPU in the future anyways).

Not sure what's best to do now though. I'm fine with whatever you decide.

kayousterhout: These are good points -- particularly (1).  One long-term decision could be to make machines immutable (or at least enforce that the size-related properties are immutable).  Anyway I agree with Kevin that it seems better to leave this functionality where it is (so ignore my previous comment).

HongjianZhang:
@kayousterhout Made adjustments and tested with Getting Started example. Ready for further reviews.

blueprint/bindings.js
kayousterhout: That's a good point re: performance.  For this kind of thing, the performance difference is minor enough that we tend to prefer readability!

blueprint/bindings_test.js
kayousterhout: Good point I agree that it's overkill to deploy the machines here, so this is fine

blueprint/bindings.js
```diff
@@ -619,6 +625,115 @@ function getBoolean(argName, arg) {
 }
 
 /**
+ * Chooses instance size if not specified (e.g., m2.xlarge). If user did specify size,
+ * this function verifies the specified size satisfies any CPU and RAM requirements that
+ * the user specified (and throws an error if it doesn't).
+ * @private
+ * @param {string} size - User specified size ('' if not specified).
```
kayousterhout: Can you update the parameter list?

blueprint/bindings.js
```diff
@@ -619,6 +625,115 @@ function getBoolean(argName, arg) {
...
+ * @param {string} provider - Used to direct to corresponding descriptions.
+ * @param {Range} ram - User specified RAM requirement.
+ * @param {Range} cpu - User specified CPU requirement.
+ * @returns {string} The best size that fits the user's requirements if provider
+ * is available in Quilt, otherwise throws an error.
+ */
+Machine.prototype.chooseSize = function machineChooseSize() {
+  switch (this.provider) {
+    case 'Amazon':
+      this.size = chooseBestSize(this.size, machineDescriptions.amazon,
```
kayousterhout: Can you move chooseBestSize and vagrantSize to be methods of Machine too, so then you only need to pass in machineDescriptions.amazon here, and chooseBestSize can handle setting this.size itself?

blueprint/bindings.js
```diff
@@ -619,6 +625,115 @@ function getBoolean(argName, arg) {
...
+        this.ram, this.cpu);
+      break;
+    case 'Vagrant':
+      this.size = vagrantSize(this.ram, this.cpu);
+      break;
+    default:
+      throw new Error(`Unknown Cloud Provider: ${this.provider}`);
+  }
+  if (this.size === '') {
+    throw new Error('No valid size for given requirements.');
```
kayousterhout: can you print the machine here?  Something like `No valid size for Machine (${JSON.stringify(this)})`

blueprint/bindings.js
```diff
@@ -619,6 +625,115 @@ function getBoolean(argName, arg) {
...
+  }
+};
+
+/**
+ * Chooses provider's default region if region is not specified by the user.
+ * @private
+ * @param {string} provider - Used to direct to corresponding descriptions.
```
kayousterhout: Can you update the parameter list?

blueprint/bindings.js
```diff
@@ -619,6 +625,115 @@ function getBoolean(argName, arg) {
...
+ * @returns {string} The default region for the given provider if provider
+ * is available in Quilt, otherwise throws an error.
+ */
+Machine.prototype.chooseRegion = function machineChooseRegion() {
+  switch (this.provider) {
+    case 'Amazon':
+      this.region = amazonDefaultRegion;
```
kayousterhout: Did you consider using a map to store these default regions?  So you'd have a variable like providerDefaultRegions, and then this code would look something like:

```
if (this.provider in providerDefaultRegions) {
  this.region = providerDefaultRegions[this.provider];
} else {
  throw new Error ...
}
```

blueprint/bindings.js
```diff
@@ -619,6 +625,115 @@ function getBoolean(argName, arg) {
...
+};
+
+/**
+ * Iterates through all the decriptions for a given provider, and returns
+ * the cheapest option that fits the user's requirements. If size is already
+ * specified, verifies that it fits user's requirements.
+ * @private
+ * @param {string} size - User specified size ('' if not specified).
+ * @param {description[]} providerDescriptions - Array of descriptions of
+ * a provider.
```
kayousterhout: super nit: line-wrapped params (and return descriptions) should be indented two spaces (it makes them marginally easier to read)

blueprint/bindings.js
```diff
@@ -619,6 +625,115 @@ function getBoolean(argName, arg) {
...
+        (bestSize === '' || description.Price < bestPrice)) {
+      bestSize = description.Size;
+      bestPrice = description.Price;
+    }
+  }
+  return bestSize;
+}
+
+/**
+ * Clips ram and cpu range to be > 1 for Vagrant
```
kayousterhout: I find "Clip" confusing here, because I think of clipping as making it smaller, but this is rounding up.  How about "Chooses a Vagrant size by rounding up the RAM and CPU requirements to each be at least 1."

blueprint/bindings.js
```diff
@@ -619,6 +625,115 @@ function getBoolean(argName, arg) {
...
+      break;
+    default:
+      throw new Error(`Unknown Cloud Provider: ${this.provider}`);
+  }
+};
+
+/**
+ * Iterates through all the decriptions for a given provider, and returns
+ * the cheapest option that fits the user's requirements. If size is already
+ * specified, verifies that it fits user's requirements.
```
kayousterhout: Can you update this comment?

blueprint/bindings.js
```diff
@@ -1142,6 +1264,17 @@ function Range(min, max) {
   this.max = max;
 }
 
+/**
+ * Checks whether x falls in the Range.
+ *
+ * @param {integer} x - The integer to check.
+ * @returns {boolean} True if x is in the Range, and False otherwise.
+ */
+Range.prototype.inRange = function accepts(x) {
```
kayousterhout: Can you change the two function names to be consistent (so change `accepts` to `inRange` or change `inRange` to `accepts`)?

blueprint/bindings.js
```diff
@@ -1142,6 +1264,17 @@ function Range(min, max) {
...
+  return this.min <= x && (this.max === 0 ||
+        this.max === this.min || x <= this.max);
```
kayousterhout: Can you add unit tests for this? (Is this code correct? It looks like if this.min == this.max, there's a bug where the value will be accepted if it's bigger than the max?)

cloud/machine/machine.go
```diff
@@ -1,60 +0,0 @@
...
-		return vagrantSize(ram, cpu)
-	default:
-		panic(fmt.Sprintf("Unknown Cloud Provider: %s", provider))
-	}
-}
-
-func chooseBestSize(descriptions []Description, ram, cpu blueprint.Range) string {
-	var best Description
-	for _, d := range descriptions {
-		if ram.Accepts(d.RAM) &&
```
kayousterhout: Can you remove the Accepts method from blueprint.go now?  It looks like it isn't used anywhere.

blueprint/bindings.js
```diff
@@ -1142,6 +1264,17 @@ function Range(min, max) {
   this.max = max;
 }
 
+/**
+ * Checks whether x falls in the Range.
```
kayousterhout: can you add `@private` here? (since it's not exported)

blueprint/bindingsTest.js
```diff
@@ -18,6 +18,12 @@ describe('Bindings', () => {
     deployment = b.createDeployment();
   });
 
+  const checkSizes = function checkSizes(size, description, ram, cpu,
+    preemptibe, expected) {
+    const resSize = b.chooseBestSize(size, description, ram, cpu, preemptibe);
```
kayousterhout: It looks like you have an extra argument to chooseBestSize (related -- can you remove preemptibe as an argument to the checkSizes function here, since it's not used anywhere?)

kayousterhout: nit: resSize is a little hard to parse.  How about "actualSize"?

blueprint/machineDescriptions.json
```diff
@@ -0,0 +1,532 @@
+{
+	"amazon": [
```
kayousterhout: Can you use 2-space indents here? (it's easier to read and I think better to be consistent with the JavaScript than with the go code, unless 8-spaces is a convention in JSON?)

blueprint/bindings.js
```diff
@@ -619,6 +625,115 @@ function getBoolean(argName, arg) {
...
+};
+
+/**
+ * Chooses provider's default region if region is not specified by the user.
+ * @private
+ * @param {string} provider - Used to direct to corresponding descriptions.
+ * @returns {string} The default region for the given provider if provider
+ * is available in Quilt, otherwise throws an error.
+ */
+Machine.prototype.chooseRegion = function machineChooseRegion() {
```
kayousterhout: Can you add unit tests for this?

cloud/machine/google.go
```diff
@@ -1,40 +0,0 @@
-package machine
-
-var googleDescriptions = []Description{
```
kayousterhout: Can you update scripts/gce_descriptions (which was used to generate this file) so that it correctly generates the new JSON? (Given the existence of that script, it might make sense to separate the machineDescriptions file into a different file for each Cloud provider)

engine/engine.go
kayousterhout: It looks like you thumbed-up this comment but didn't remove anything from blueprint.go -- are these variables still needed there?

kayousterhout: Also, can you update the method comment on line 45?

blueprint/bindings.js
HongjianZhang: Issue with making chooseBestSize a method of Machine is that I am not sure how I can unit test it without creating a machine object.

blueprint/bindings.js
HongjianZhang: There are unit tests for the method, but they are under the unit tests for Machine right now.

blueprint/bindings.js
HongjianZhang: The reason this.max == this.min only happens when use only specifies one number for RAM or CPU instead of a Range. In that case, the user probably wants  lower bound, so it's okay for x to be larger than max.

HongjianZhang: Im not very certain how to do this. If a user only specifies one number. Should we interpret it as the user wants exactly this amount of CPU/RAM or should we interpret it as the user accepts any CPU/RAM that is >= that number?

blueprint/bindings.js
HongjianZhang: Another issue is that internally chooseBestSize is called with the descriptions for Google, Amazon, etc. If we test chooseBestSize as part of the Machine tests, it will be hard to use our testDescriptions. I guess couple solutions are:
1. Instead of using testDescription, I will just test with the real descriptions from the providers.
2. Include testDescriptions in the JSON file of descriptions. 
3. Export chooseBestSize to have its own unit tests.
I am not too sure which way is the one to go.

HongjianZhang: Another thing is that since chooseBestSize, when called internally, is always passed a description from an actual provider. So if we were to unit test it in Machine test, it would be hard to use testDescriptions. Couple solutions are:
1. Include testDescriptions in the JSON file for all the descriptions.
2. Unit test using actual provider descriptions and sizes.
3. Export chooseBestSize to have its own unit tests.
I am not too sure which way to go.

HongjianZhang: Another thing is that since chooseBestSize, when called internally, is always passed a description from an actual provider. So if we were to unit test it in Machine test, it would be hard to use testDescriptions. Couple solutions are:
1. Include testDescriptions in the JSON file for all the descriptions.
2. Unit test using actual provider descriptions and sizes.
3. Export chooseBestSize to have its own unit tests.
I am not too sure which way to go.

kayousterhout: I don't think JavaScript classes allow private methods, so chooseBestSize would be public by default, so you can create a Machine object and then use the same test descriptions as in the old tests (this is essentially your (3) above, except that you don't have to specifically export chooseBestSize).  That seems like a good approach to me -- does that seem reasonable to you?

HongjianZhang: It makes sense. However the only issue in the current implementation is that when we are creating a new Machine object, chooseSize will be called as part of the constructor, which will check for a valid provider, and call chooseBestSize on the descriptions for that provider.

cloud/machine/google.go
HongjianZhang: That makes sense. Should this go into a separate commit?

blueprint/bindings.js
kayousterhout: You could still re-call chooseBestSize and make sure the size gets (re)set correctly. I also think it's fine to just have the tests use the real Google / Amazon descriptions and avoid calling chooseBestSize directly. It's up to you!

cloud/machine/google.go
kayousterhout: I think it should be in this commit, because otherwise this commit effectively breaks that script, and we try to organize commits to avoid any intermediate broken states between commits (even when they're in the same PR)

HongjianZhang: Also I think we don't necessarily have to separate machineDescriptions. One alternative is just to have the script open the machineDescription, and modify the corresponding provider's descriptions, and write back to machineDescriptions. I'm not sure which way is better tho.

kayousterhout: If it's easy to have the script modify the appropriate provider's descriptions, it seems fine to me to keep them all in one file! I think either way is OK so it's up to you.

HongjianZhang:
@kayousterhout Made adjustments and tested with Getting Started guide.

blueprint/amazonDescriptions.json
```diff
@@ -0,0 +1,328 @@
+{
+	"Descriptions": [
```
kayousterhout: Can you change this to use 2 spaces rather than 8? (Unless there's a reason to use 8? We use 2 for JavaScript so seems more consistent to use that here; Go is weirdly-large at 8).  Most editors include tools that allow you to easily update this!

blueprint/bindings.js
```diff
@@ -672,6 +683,92 @@ function getBoolean(argName, arg) {
...
+ * based on the Machine's specified provider, region, and hardware. Throw
+ * an error if provider is not valid or given machine requirements cannot
+ * be satisfied by any size.
+ * @private
+ * @returns {void}
+ */
+Machine.prototype.chooseSize = function machineChooseSize() {
+  switch (this.provider) {
+    case 'Amazon':
+      this.size = this.chooseBestSize(amazonDescriptions.Descriptions);
```
kayousterhout: Can you move the `this.size =` into the `chooseBestSize` function, or is there a reason to have it out here?

blueprint/bindings.js
```diff
@@ -672,6 +683,92 @@ function getBoolean(argName, arg) {
...
+      break;
+    case 'DigitalOcean':
+      this.size = this.chooseBestSize(digitalOceanDescriptions.Descriptions);
+      break;
+    case 'Google':
+      this.size = this.chooseBestSize(googleDescriptions.Descriptions);
+      break;
+    case 'Vagrant':
+      this.size = this.vagrantSize();
```
kayousterhout: same as above re:setting size in vagrantSize

blueprint/bindingsTest.js
```diff
@@ -18,6 +18,19 @@ describe('Bindings', () => {
...
   });
 
+  const checkSizes = function checkSizes(description, ram, cpu,
+    expected) {
+    const dummyMachine = new b.Machine({
+      role: 'Master',
+      ram,
+      cpu,
+      provider: 'Amazon',
+      size: 'm3.medium',
```
kayousterhout: Why set the size here? (can you eliminate this line?)

blueprint/bindingsTest.js
```diff
@@ -18,6 +18,19 @@ describe('Bindings', () => {
     deployment = new b.Deployment();
   });
 
+  const checkSizes = function checkSizes(description, ram, cpu,
+    expected) {
+    const dummyMachine = new b.Machine({
+      role: 'Master',
```
kayousterhout: Do you need to set the role here? (Can you eliminate this line?)

blueprint/digitalOceanDescriptions.json
```diff
@@ -0,0 +1,160 @@
+{
+	"Descriptions": [
```
kayousterhout: Same as above re: 2 spaces

blueprint/machineDescriptions.json
```diff
@@ -0,0 +1,532 @@
+{
```
kayousterhout: Is it intentional that you still have this file? (looks like it's been replaced by the provider-specific ones?)

blueprint/bindings.js
```diff
@@ -672,6 +683,92 @@ function getBoolean(argName, arg) {
...
+      break;
+    default:
+      throw new Error(`Unknown Cloud Provider: ${this.provider}`);
+  }
+  if (this.size === '') {
+    throw new Error(`No valid size for Machine ${JSON.stringify(this)}`);
```
kayousterhout: I realized I led you astray here -- I think you can just use `stringify` rather than `JSON.stringify`

blueprint/bindings.js
```diff
@@ -672,6 +683,92 @@ function getBoolean(argName, arg) {
...
 
 /**
+ * Sets the machine's size attribute to an instance size (e.g., m2.xlarge),
+ * based on the Machine's specified provider, region, and hardware. Throw
+ * an error if provider is not valid or given machine requirements cannot
+ * be satisfied by any size.
+ * @private
+ * @returns {void}
+ */
+Machine.prototype.chooseSize = function machineChooseSize() {
```
kayousterhout: I realized -- I think now you don't need to set `this.cpu` and `this.ram` in the `Machine` constructor.  It seems like instead, they should just be local variables that get passed into the chooseSize functions.  This means they won't be serialized as part of the json object, which is good because in general we're trying to reduce the size of the json object to minimize the information that gets passed between JS and go (and they shouldn't be in the json object if they're never read by the go code).  This also will help prevent user error later, because a user can't modify the machine's CPU or RAM after creating it.  This is an API-breaking change (albeit a small one) so should be in a new commit; can you add this to your to-do list for after this PR gets merged?

blueprint/bindingsTest.js
HongjianZhang: The reason is that when a Machine is constructed without specifying size, chooseSize will be called which will call chooseBestSize with Amazon's descriptions. Since the CPU and RAM in these tests are to test the test descriptions, we might see error. Therefore I passed in size just so that chooseSize is not called.

blueprint/machineDescriptions.json
HongjianZhang: I noticed that the new Google descriptions I parsed are slightly different from the ones in this file. I am pretty sure Google probably changes the prices once in a while, but just to be safe, I kept this file.

kayousterhout:
I forgot to mention in my comments last name -- can you fix the spelling typo in "functionarlity" (remove "r") in the commit msg?

blueprint/bindingsTest.js
kayousterhout: That makes sense -- can you add a comment saying this in the test code?

HongjianZhang:
@kayousterhout Fixed.

blueprint/bindings.js
```diff
@@ -672,6 +683,92 @@ function getBoolean(argName, arg) {
...
+      bestSize = description.Size;
+      bestPrice = description.Price;
+    }
+  }
+  this.size = bestSize;
+};
+
+/**
+ * Rounds up RAM and CPU requirements to be at least one for Vagrant
+ * @returns {string} The clipped Vagrant size.
```
kayousterhout: Can you mark this as private, add a period after "Vagrant", and update the return string to say "The rounded up Vagrant size."? (clipped is a bit confusing / non-intuitive)

blueprint/bindings.js
```diff
@@ -672,6 +683,92 @@ function getBoolean(argName, arg) {
...
+      throw new Error(`Unknown Cloud Provider: ${this.provider}`);
+  }
+  if (this.size === '') {
+    throw new Error(`No valid size for Machine ${stringify(this)}`);
+  }
+};
+
+/**
+ * Sets the machine's region using the default region of the specified provider
+ * if region is not specified by the user. Throw an error if provider is not
```
kayousterhout: Can you delete "if region is not specified by the user" (since that's checked by the caller, not here)?

blueprint/bindingsTest.js
```diff
@@ -18,6 +18,18 @@ describe('Bindings', () => {
     deployment = new b.Deployment();
   });
 
+  const checkSizes = function checkSizes(description, ram, cpu,
+    expected) {
+    const dummyMachine = new b.Machine({
+      ram,
+      cpu,
+      provider: 'Amazon',
+      size: 'm3.medium',
```
kayousterhout: Can you add a comment above this line about why this is needed?  Something like what you said to me in the earlier comment:
```
// Specify a placeholder size so that chooseBestSize isn't called, and the test can instead call
// chooseBestSize with it's own machine descriptions, so we can test more specific
// and restricted cases.
```

blueprint/bindingsTest.js
```diff
@@ -72,6 +144,106 @@ describe('Bindings', () => {
...
+    });
+    it('uses provided region when region is provided', () => {
+      deployment.deploy([new b.Machine({
+        role: 'Master',
+        provider: 'Amazon',
+        sshKeys: ['key1', 'key2'],
+        region: 'eu-west-1',
+      })]);
+      checkMachines([{
+        id: '21e797b7b251966809aa3d08e680c981b24db18e',
```
kayousterhout: Can you eliminate the Id from all of these tests?  I don't think it's needed (lmk if I'm misunderstanding here) and it makes the tests more brittle in terms of needing to be updated if any other attributes of the machine change and change the ID as a result

kayousterhout: (Ideally we should probably remove the ID from a bunch of the other tests too, but that's unrelated to this PR, and here we can at least avoid adding the ID to new tests that we know don't need it)

blueprint/blueprint.go
```diff
@@ -95,12 +93,6 @@ type Range struct {
 // network.
 const PublicInternetLabel = "public"
 
-// Accepts returns true if `x` is within the range specified by `blueprintr` (include),
```
kayousterhout: Can the Range class be eliminated now?

blueprint/bindingsTest.js
```diff
@@ -48,6 +60,66 @@ describe('Bindings', () => {
...
+      checkInRange(2, 2, 2, true);
+      checkInRange(2, 2, 3, false);
+    });
+    it('value is on boundary', () => {
+      checkInRange(2, 4, 4, true);
+      checkInRange(2, 4, 2, true);
+    });
+  });
+
+  describe('Size', () => {
```
kayousterhout: now that chooseSize is part of Machine, can these tests be moved inside of Machine?  I think you can move this whole "describe" block inside the "describe('Machine')" block of code (and I'd rename Size to chooseBestSize, to be more explicit about what's being tested)

blueprint/bindingsTest.js
HongjianZhang: The IDs in some of the existing machine tests might be useful. One example being the 'hash independent of SSH keys' I think. I think I will just leave these there, while removing the IDs from all the tests I wrote.

blueprint/bindings.js
```diff
@@ -672,6 +683,93 @@ function getBoolean(argName, arg) {
 }
 
 /**
+ * Sets the machine's size attribute to an instance size (e.g., m2.xlarge),
```
ejj: This is somewhat of an overall architectural piece of feedback, but I actually don't think it should be implemented in this patch:

A lot of the switching we're doing on provider could be implemented a lot more simply if we just made each provider a class with a chooseSize method and a defaultRegion member variable.  Didn't work that well in go, because it's not particularly object oriented, but in js I think that could be done extremely cleanly.  At any rate, I don't think that should block merging it as is.

luise: Yeah, at some point, it might make sense to merge this with the initializer code's `Provider`. That code also keeps some provider specific information is a few json files, so it would be clean to just consolidate that.

blueprint/bindings.js
```diff
@@ -740,6 +838,13 @@ function Machine(optionalArgs) {
   this.preemptible = getBoolean('preemptible', optionalArgs.preemptible);
 
   checkExtraKeys(optionalArgs, this);
+
+  if (this.size === '') {
```
ejj: So I think this is actually a bit awkward because chooseSize is responsible for setting this.size, but its not responsible for checking if this.size is `''`.  I would think either chooseSize should return a size, or chooseSize should be responsible for checking if this.size is not empty, and doing nothing in that case.

Also, on a side note, I'm not entirely convinced chooseSize and chooseRegion should be member variables of a machine.   They aren't really features of a machine so much as their features of a provider that a machine runs on.  One could make an argument that they should just be private functions of this file, rather than methods on the machine object.  That said, I think there are tradeoffs here, you could argue it either way.  If you think this is cleaner its fine to leave it.

kayousterhout: This was my suggestion (making these classes part of the Machine object) -- which I thought made sense because the size is an attribute of the Machine, and this functionality is only needed in determining how to construct machine objects.

blueprint/bindings.js
kayousterhout: Ah yeah that would be nice to consolidate with `Provider` in the `init` code (and agree that this shouldn't be part of / block this PR).  Seems like also should wait until the `init` changes that are tentatively planned for the next sprint.

blueprint/bindings.js
HongjianZhang: It definitely makes the code cleaner, since fewer parameters are passed around.

ejj: Yeah I think it's debatable either way. Sounds like you two thought about it and this was the best option.

ejj: Oh I will say though, I think the this.size == '' check should move into the chooseSize function if we're going this direction.

blueprint/bindings.js
```diff
@@ -672,6 +683,97 @@ function getBoolean(argName, arg) {
...
 /**
+ * Sets the machine's size attribute to an instance size (e.g., m2.xlarge),
+ * based on the Machine's specified provider, region, and hardware. Throw
+ * an error if provider is not valid or given machine requirements cannot
+ * be satisfied by any size.
+ * @private
+ * @returns {void}
+ */
+Machine.prototype.chooseSize = function machineChooseSize() {
+  if (this.size === '') {
```
ejj: So in general, it's better to check the negative and return.  That way you don't have to indent everything, and it's easier to read what all of the exceptions are.

blueprint/bindings.js
```diff
@@ -740,6 +842,10 @@ function Machine(optionalArgs) {
   this.preemptible = getBoolean('preemptible', optionalArgs.preemptible);
 
   checkExtraKeys(optionalArgs, this);
+
+  this.chooseSize();
+
```
ejj: Why this empty line?

ejj:
Two minor style nits. Please slack me when you've updated and I'll merge.

PR #1134, opened 2017-10-03
cli: Make `usage` dynamically generate the list of subcommands

aegamesi:
Previously, `quilt usage` used a hardcoded list of subcommands to
display to the user. This makes usage generate the list of subcommands
at runtime from those that are defined in cli.go.

I was considering writing a general purpose method to word wrap a
string to specified number of characters but decided that such a method
would require too much code to make it general (especially considering
it would only be used for printing the usage string).


ejj:
Did you consider using this? https://github.com/mitchellh/go-wordwrap

The hashicorp stuff is typically very good

aegamesi:
@ejj : yep, I saw that. I just didn't think it was worth it to bring in a whole other dependency for a small part of a relatively minor feature.

kklin:
Luise proposed wrapping here: https://github.com/quilt/quilt/pull/941. Not sure if it's relevant.

cli/cli.go
```diff
@@ -60,6 +61,16 @@ func HasSubcommand(name string) bool {
 	return ok
 }
 
+// GetSubcommands returns a slice containing the names of all subcommands
+func GetSubcommands() []string {
+	result := make([]string, 0, len(commands))
```
luise: This can just be `var result []string`.  That's what we typically do.

aegamesi:
@luise Updated with your feedback (and to match the new project name)

kelda.go
```diff
@@ -79,6 +78,27 @@ func main() {
...
 }
 
+// printUsageString generates and prints a usage string based off of the
+// subcommands defined in cli.go.
+func printUsageString() {
+	var commandBuffer bytes.Buffer
+	commandBuffer.WriteString("  ")
+
+	subcommands := cli.GetSubcommands()
+	for i, command := range subcommands {
```
luise: Have you looked at [strings.Join](https://golang.org/pkg/strings/#Join)?
I would be in favor of just using strings instead of `bytes.Buffer`, and then use `strings.Join` instead of the for loop. It would simplify a lot, and we don't really need the efficiency from `bytes.Buffer`. What do you think?

aegamesi: Can't believe I missed that! I suppose that was an artifact from when I was manually word-wrapping the string. Taking it out cleans up the code a lot.

cli/cli.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/kelda/kelda/cli/command/inspect"
 
 	log "github.com/sirupsen/logrus"
+	"sort"
```
kklin: This import should be in the first group

cli/cli.go
```diff
@@ -59,6 +60,16 @@ func HasSubcommand(name string) bool {
 	return ok
 }
 
+// GetSubcommands returns a slice containing the names of all subcommands
+func GetSubcommands() []string {
```
kklin: In response to @luise's comment re unit testing, I think leaving this uncovered is fine.

FWIW if I were to rewrite the `cli` package, I would move this entire file into `quilt.go`, and write a function like `getSortedKeys(map[string]subCommand)`. Then I could unit test `getSortedKeys` (the caller would pass in the `commands` map). In other words, the actual problem here is that `quilt.go` and this part of the `cli` package are unnecessarily split.

PR #1136, opened 2017-10-03
tester: Update use of Node.js blueprint

kayousterhout:
The Node.js blueprint renamed cluster to containers (https://github.com/quilt/nodejs/pull/8);
this commit updates the use of that blueprint in the
integration-tester to use the new API.


kayousterhout:
(FYI the tests are expected to fail here, because there's a synchronization issue between this and the nodejs PR)

PR #1141, opened 2017-10-04
Improve JSDoc

kayousterhout:
This PR includes a bunch of commits to use better JSDoc in our code (and to enforce it in our lint configuration).

@luise it would be great if you could sanity check the changes I made to the quilt init documentation.


cli/command/init/prompter.js
```diff
@@ -58,6 +58,15 @@ function getInquirerDescriptions(friendlyNameToValue) {
...
 
+/**
+ * @param {Object} question - An object with fields representing a question to
+ *   ask a user.
+ * @param {string} helpstring - The helpstring to show a user if they type '?' in
+ *   response to the question.
+ * @returns {Object} The given question, with a field added for the help message
+ *   and to validate user responses to the question (the validation function checks
+ *   for whether the user typed '?', and if so, returns the given help string).
+ */
```
luise: The `returns` comment is difficult to parse. How about 
```
The given question, with a modified `message` that now suggests using `?`, and a modified `validate` function that will print the helpstring if the user enters `?`.
```

kayousterhout:
Thanks -- fixed! And yes alas a weird github blib re: commit ordering.  You may already know this -- but you can always go look at the commit history for the branch corresponding to the PR to check this (I've started doing this for PRs because otherwise reviewing can be confusing).  E.g., here https://github.com/kayousterhout/quilt/commits/jsDoc

PR #1142, opened 2017-10-04
tls: Add minion's private IP to TLS certificate

kklin:
TLS certificates for services such as Vault and etcd check need to be
issued for the addresses the service listens on. This commit modifies
the certificate installed on minions to have the minion's private IP so
that it can be used with these services.


cloud/credentials_test.go
```diff
@@ -37,8 +37,8 @@ func TestSyncCredentials(t *testing.T) {
 	assert.NoError(t, err)
 
 	credentialedMachines := map[string]struct{}{}
-	syncCredentialsOnce(expSigner, ca,
-		[]db.Machine{{PublicIP: expHost}}, credentialedMachines)
+	syncCredentialsOnce(expSigner, ca, []db.Machine{
+		{PublicIP: expHost, PrivateIP: "9.9.9.9"}}, credentialedMachines)
```
aegamesi: I don't know if there are specific rules for this (and this is really nitpicky), but if you put the `[]db.Machine{` on the same line as the actual contents of the slice, it might be a bit easier to read? Though you'd have to put the `credentialedMachines)` on the next line.

kklin: Nitpicky is fine!

There's no formal rule, just whatever we think looks best. In this case I agree with you, and made the change.

kklin:
@aegamesi if you don't have any other comments can you approve it?

PR #1146, opened 2017-10-06
server: Warns user to run 'quilt stop' when Daemon gets SIGINT

HongjianZhang:
Before this commit, if user tries to kill the daemon before running
'quilt stop', the machines would still be active although daemon is
shut down. This commit prevents daemon from shutting down and 
shows a warning message when this happens.


api/server/server.go
```diff
@@ -57,10 +57,20 @@ func Run(conn db.Conn, listenAddr string, runningOnDaemon bool,
...
-		os.Exit(0)
+		for sig := range c {
+			// Checks whether there are still active machines when
+			// interrupted. If there are, warns user to run 'quilt stop'.
+			var activeMachinesCount = len(conn.SelectFromMachine(nil))
+			if activeMachinesCount > 0 {
+				log.Warnf("\n%d machines are still active. Run"+
+					" 'quilt stop' before shutting down daemon.\n",
+					activeMachinesCount)
+			} else {
```
ejj: I think we still want to allow the daemon to be killed if there are active machines.  I do that all the time FWIW.  It's just we shouldn't warn if all the machines are stopped.

api/server/server.go
```diff
@@ -58,6 +58,14 @@ func Run(conn db.Conn, listenAddr string, runningOnDaemon bool,
 	signal.Notify(sigc, os.Interrupt, os.Kill, syscall.SIGTERM, syscall.SIGHUP)
 	go func(c chan os.Signal) {
 		sig := <-c
+		// Checks whether there are still active machines when
```
ejj: I don't think this comment is providing a lot of extra information that isn't obvious from the code, I would just delete it.

api/server/server.go
```diff
@@ -58,6 +58,14 @@ func Run(conn db.Conn, listenAddr string, runningOnDaemon bool,
...
+		// interrupted. If there are, warns user to run 'quilt stop'.
+		var activeMachinesCount = len(conn.SelectFromMachine(nil))
+		if activeMachinesCount > 0 {
+			log.Warnf("\n%d machines are still active. Run"+
```
ejj: I would make a slight tweak to this warning message.  Perhaps "There are %d machines that will continue running when Quilt shuts down.  If you'd like to stop them, restart the daemon and run `quilt stop`."

Something like that.  The current error message is a little confusing (what does active mean?) and a bit too prescriptive (I may not want to shut down the machines by running `quilt stop`).

luise: Tiny rewording that might be easier to read: "%d machines will continue running after the Quilt daemon shuts down. If you'd like to stop them, restart the daemon and run quilt stop.""

aegamesi: I believe running `quilt stop` won't actually do anything -- you'd have to run the daemon, _run the blueprint_, and then do quilt stop.

ejj: So this isn't precisely true.  You could do `quilt stop <namespace>` without running the blueprint first.  Also I think if you're running in the default namespace, `quilt stop` will just work.

aegamesi: Wait, really? I haven't done any sort of special namespace setup and `quilt stop` does not work (on a newly booted daemon).

kklin: I think Eli is right.

aegamesi: Yep, and it works with `quilt stop default-namespace`. Maybe that should be some default behavior?

HongjianZhang: Yes. Running `quilt stop` without specifying namespace or have a running blueprint will just gives an error.

HongjianZhang:
Changed the warning message to tell user to run `kelda stop <namespace>`.

HongjianZhang:
@ejj Ready for more reviews.

PR #1148, opened 2017-10-07
cloud: Make provider functions non-blocking

ejj:
Before this patch, each provider was expected to block until requests
made to it had completed and were visible in the List() function.
This required them to implement a lot of duplicate functionality that
was a bit finicky to get right. This patch now allows the cloud
providers to return immediately, and causes the cloud package to do
the waiting instead.

Note that this blocks on https://github.com/quilt/quilt/pull/1144


ejj:
Eh, git is a bit confused and things all of the commits of my other Pr are part of this one ... I'll close for now and reopen when my other Pr is merged

PR #1151, opened 2017-10-09
docs: Document unsupported persistent storage

luise:
One could also argue that we should put a warning in the READMEs for the storage blueprints -- e.g. Elasticsearch and MongoDB.


docs/_FAQ.md
```diff
@@ -27,6 +27,13 @@ There are a few possible reasons:
 3. When exposing a different port than `80`, make sure to paste both the
   IP address _and_ the port number into the browser as `<IP>:<PORT>`.
 
+### How do I get persistent storage?
+Quilt currently doesn't support persistent storage, so we recommend using
+a hosted database like [Firebase](https://firebase.google.com/).
+If you still choose to run storage applications like [MongoDB](https://github.com/quilt/mongo)
+or [Elasticsearch](https://github.com/quilt/elasticsearch) on Quilt, be aware
+that the data will be lost if the DB containers or the VMs hosting them die.
```
ejj: "lost if the containers or VMs that host them die" perhaps.  I got tripped up on "the DB containers"

luise: Do you think it's still clear enough then that it's if e.g. the Mongo containers die, and not a Node container?

luise:
@ejj , updated

PR #1153, opened 2017-10-10
Implement ACLs on DigitalOcean (2)

aegamesi:
This fixes up the issues with @mmussomele's PR #1041 . I couldn't add commits to that PR, so I created a new one.

Issues fixed:
* using a port in an ICMP rule in a DigitalOcean API call would cause the call to fail
* making calls to DigitalOcean's Firewall API AddRules/RemoveRules without any rules to add/remove would cause a bunch of (harmless) annoying errors
* if rules to allow droplets to communicate are not specified, they will not be allowed to communicate

Everything appears to work now! DigitalOcean's API seems to be slightly flaky, so occasionally a single API call will fail, only to succeed when it's retried a few seconds later. This sometimes causes some errors to appear in the daemon even when nothing is wrong.


kklin:
Nice work debugging this, seems like there were some tricky subtle bugs!

A comment about the commit organization -- the rule of every commit not breaking the build should still be true. Right now it isn't, because a buggy `SetACLs` is introduced, and then you have some commits to fix it up. It'd be cleaner to squash your bug fixes into the original `SetACLs` patch. You should claim credit on Matt's commit by putting "Co-Authored-By" at the end of the commit (search our git history for examples)

aegamesi:
Okay, @kklin @kayousterhout I've squashed commits accordingly

cloud/digitalocean/client/client.go
```diff
@@ -21,7 +21,7 @@ type Client interface {
 	AssignFloatingIP(string, int) (*godo.Action, *godo.Response, error)
```
kklin: In the commit message, s/settings the ACLs/setting the ACLs

cloud/digitalocean/client/client.go
```diff
@@ -21,7 +21,7 @@ type Client interface {
...
 	UnassignFloatingIP(string) (*godo.Action, *godo.Response, error)
 
-	CreateFirewall(string, []godo.OutboundRule) (*godo.Firewall,
+	CreateFirewall(string, []godo.OutboundRule, []godo.InboundRule) (*godo.Firewall,
```
kklin: This change would be better in the commit that adds these methods

kklin: (the mock client should thus also be regenerated in that commit)

cloud/digitalocean/digitalocean.go
```diff
@@ -293,8 +326,127 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+	curACLSet := map[acl.ACL]struct{}{}
+	for _, cur := range current {
+		ports := strings.Split(cur.PortRange, "-")
+		if len(ports) == 1 {
+			ports = []string{ports[0], ports[0]}
+		}
+
+		from, err := strconv.Atoi(ports[0])
+		if err != nil {
+			log.WithError(err).Warn(
```
kklin: This log and the one below it would be more helpful it included the string it was trying to parse (`log.WithError(err).WithField("port", ports[0]).Warn...`)

aegamesi:
@kklin ^

cloud/digitalocean/digitalocean.go
```diff
@@ -174,6 +206,11 @@ func (prvdr Provider) Boot(bootSet []db.Machine) error {
 	return err
 }
 
+// Returns a unique tag to use for all entities in this namespace and region
```
kklin: This comment should end with a period

cloud/digitalocean/digitalocean.go
```diff
@@ -293,8 +331,144 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+		if firewall.Name == tagName {
+			return &firewall, nil
+		}
+	}
+
+	// The outbound rules and inter-droplet inbound rules are generated only
+	// once: when the firewall is first created. If these rules are externally
+	// deleted, there will be errors unless the firewall is destroyed
+	// (and then recreated by the daemon).
+	_, _, err = prvdr.CreateTag(tagName)
```
kklin: Will this error if the tag already exists?

cloud/digitalocean/digitalocean.go
```diff
@@ -293,8 +331,144 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+	}
+
+	tagName := prvdr.getTag()
+	for _, firewall := range firewalls {
+		if firewall.Name == tagName {
+			return &firewall, nil
+		}
+	}
+
+	// The outbound rules and inter-droplet inbound rules are generated only
```
kklin: I'd put this comment a little closer to the rules (i.e. not with the code to create the tag)

cloud/digitalocean/digitalocean.go
```diff
@@ -293,8 +331,144 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+	for _, intf := range toAdd {
+		add = append(add, intf.(acl.ACL))
+	}
+	for _, intf := range toRemove {
+		remove = append(remove, intf.(acl.ACL))
+	}
+	return toRules(add), toRules(remove)
+}
+
+func toRules(acls []acl.ACL) (rules []godo.InboundRule) {
```
kklin: This code is a bit complicated -- could you unit test it?

cloud/digitalocean/digitalocean.go
```diff
@@ -293,8 +331,144 @@ func (prvdr Provider) deleteAndWait(ids string) error {
...
+			PortRange: "all",
+			Sources:   internalDroplets,
+		},
+	}
+	firewall, _, err := prvdr.CreateFirewall(tagName, allowAll, allowInternal)
+	return firewall, err
+}
+
+func syncACLs(desired []acl.ACL, current []godo.InboundRule) (
+	addRules, removeRules []godo.InboundRule) {
```
kklin: Re-reading this, I realized that this code is a bit backwards.. I think what we should be doing is joining `godo.InboundRule`s, rather than `acl.ACL`s. So, converting `desired` into `[]godo.InboundRule`, and joining that with current. Right now, we're ignoring the protocol information in `current`.

I don't know if it's worth the effort, though. What's there now will work I think, and making this change might take some time. What do you think?

cloud/digitalocean/client/client.go
```diff
@@ -17,15 +17,25 @@ type Client interface {
 	GetDroplet(int) (*godo.Droplet, *godo.Response, error)
```
kklin: You should add Co-Authored-By to this commit as well

cloud/digitalocean/digitalocean.go
aegamesi: Nope.

cloud/digitalocean/digitalocean.go
aegamesi: Yeah, that's probably the better approach. The only thing I can think of (besides needing to change a bunch of code) is that the whole "allow connections from all other droplets/machines" would break... or, well, it would have to be made much more explicit. Which, honestly, is probably a good thing.

aegamesi: Yeah, this is definitely a change that should be made.

kklin: That's true. How long do you think it would take?

kklin: I'm thinking that we should just merge it as is and I'll change the join over the weekend or something. I'm fast at writing this sort of code, and I'd enjoy doing it. And I'd rather you start working on the foreman/cloud interaction stuff

aegamesi: Okay, sure. I don't think it would take terribly long, but yeah, that works too.

cloud/digitalocean/digitalocean.go
aegamesi: Do you think it's still worth writing a unit test for this part? If it's going to get completely rewritten (with behavior changed and everything) soon anyway.

aegamesi: Added a test

PR #1154, opened 2017-10-10
Added new Infrastructure() constructor

kayousterhout:
This series of commits adds a new Infrastructure() constructor, which is intended to replace Deployment().  These commits leave the Deployment() constructor in place, so we can update all of our other repositories without needing to synchronize, and then we can remove the Deployment() class (and createDeployment() method) after that.


kayousterhout:
@kklin the commits are slightly out of order here again (the boxObjects commit shows up 1 commit later than it should actually be; real order here: https://github.com/kayousterhout/quilt/commits/newDeploy)

blueprint/bindings.js
```diff
@@ -97,7 +97,7 @@ function baseInfrastructure(name = 'default') {
   return getInfraDeployment(infraPath);
```
kklin: I found the commit title hard to understand. Maybe it could be more explicit? Something like "Deprecate `createDeployment` in favor of `new Deployment`"

blueprint/bindingsTest.js
```diff
@@ -676,6 +676,42 @@ describe('Bindings', () => {
...
     });
   });
+  describe('Infrastructure', () => {
+    it('constructing Infrastructure overwrites Deployment', () => {
+      const namespace = 'testing-namespace';
+      const machine = new b.Machine({
+        provider: 'Amazon',
+        region: 'us-west-2',
+      });
+      const infra = new b.Infrastructure([machine], [machine], { namespace: 'testing-namespace' });
```
kklin: Did you mean to use the `namespace` variable in the optional args here?

blueprint/bindingsTest.js
```diff
@@ -676,6 +676,42 @@ describe('Bindings', () => {
         .to.throw('Unrecognized keys passed to Deployment constructor: badArg');
     });
   });
+  describe('Infrastructure', () => {
+    it('constructing Infrastructure overwrites Deployment', () => {
```
kklin: I don't really get what the test description means

cli/command/init/initializerTest.js
```diff
@@ -265,10 +264,10 @@ describe('Initializer', () => {
     preemptible: true
   });
 
-  inf.deploy(vmTemplate.asMaster().replicate(1));
-  inf.deploy(vmTemplate.asWorker().replicate(2));
-
-  return inf;
+  return new quilt.Infrastructure(
```
kklin: Looks nice!

kayousterhout:
@kklin I updated based on your comments.  I can replicate the Travis failure locally, unless I explicitly link the newer version of Quilt from the examples repo, but I'm suspicious because I don't remember this failing before.  Thoughts?

kklin:
Which failure? I'd expect the blueprint tests to fail now (I noticed it in your first PR as well).

I don't think the integration-tester unit test should fail though.

kayousterhout:
Ah k the blueprint tests are the ones that are failing (I remembered them as passing in the earlier version but sounds like I just remembered wrong!)

PR #1155, opened 2017-10-11
[DON'T MERGE] Vault RFC

kklin:
I'm opening this PR so that Kay can look at what I have so far, and possibly offer suggestions about design improvements.

@kayousterhout I'm specifically looking for feedback on what you think of the SecretOrString structure. Other feedback is appreciated, but I'd avoid making more minor comments like requesting renames, adding comments, breaking up functions etc.

There's a lot of commits in the PR because it depends on Ethan's cloud re-architecture stuff. The only one that you need to read I think is the final one, "secret implementation wip".

Thanks for taking the time to look at this!!


kklin:
Also, I think you should just ask me in person about any parts that are confusing since I don't think you should waste time trying to grok this sloppy version.

api/server/server.go
```diff
@@ -70,6 +71,24 @@ func Run(conn db.Conn, listenAddr string, runningOnDaemon bool,
 	return nil
 }
 
+func (s server) SetSecret(ctx context.Context, msg *pb.Secret) (*pb.SecretReply, error) {
```
kayousterhout: FWIW I wonder if AddSecret is better than SetSecret? (just food for thought -- I don't feel strongly about this)

minion/etcd/container.go
```diff
@@ -134,3 +135,16 @@ func joinContainers(view db.Database, etcdDBCs []db.Container) {
 		view.Commit(dbc)
 	}
 }
+
+func secretOrStringMapKey(x map[string]blueprint.SecretOrString) string {
```
kayousterhout: Since this is just for the join, rather than use this new function, why not split Env and FPToContent each into two variables: one for strings and one for secrets? That feels slightly less icky to me (also potentially more future-proof? If we add more fields to secrets to help with rotation, that you'd want to check for in the join)

minion/vault/policy.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Process the scheduled containers to decide what minions should be allowed
+	// which secrets.
+	workerToSecrets := map[string][]string{}
+	scheduledContainers := conn.SelectFromContainer(func(dbc db.Container) bool {
+		return dbc.Minion != ""
+	})
+	for _, c := range scheduledContainers {
+		referencedSecrets := append(getReferencedSecrets(c.Env),
+			getReferencedSecrets(c.FilepathToContent)...)
```
kayousterhout: Here, you're splitting in to secrets and non-secrets.  This seems to me like a reason to have 4 rather than 2 variables in the container in the database, so they're already explicitly split

minion/scheduler/worker.go
```diff
@@ -122,16 +235,16 @@ func doContainers(dk docker.Client, ifaces []interface{},
 }
 
 func dockerRun(dk docker.Client, iface interface{}) {
-	dbc := iface.(db.Container)
+	dbc := iface.(resolvedContainer)
```
kayousterhout: This is also a little icky -- it seems like this code shouldn't have to be "aware" of this concept of secrets / "resolved" containers

kklin:
Thanks for the comments @kayousterhout. Closing this now that I've gotten some feedback

api/server/server.go
kklin: BTW I named it SetSecret because the endpoint doesn't check whether you're adding it for the first time. It overwrites the secret if it already exists

PR #1159, opened 2017-10-13
scheduler: Comment the TestSyncWorker test

kklin:
It was hard to tell what it was testing before, and thus hard to modify.


minion/scheduler/worker_test.go
```diff
@@ -63,10 +63,21 @@ func runSync(dk docker.Client, dbcs []db.Container,
 	return changes
 }
 
+// 1. Initially, there are no containers running in Docker, and one container
```
kayousterhout: Can you add a brief summary first - e.g. "This test verifies three stages: "

minion/scheduler/worker_test.go
```diff
@@ -63,10 +63,21 @@ func runSync(dk docker.Client, dbcs []db.Container,
...
+// changes are made.
+//
+// 3. Finally, the test simulates removing the container from the database,
+// which causes the container in Docker to be removed as well.
 func TestSyncWorker(t *testing.T) {
 	t.Parallel()
 
 	md, dk := docker.NewMock()
+
+	// The containers that should be running. This is normally fetched from the
```
kayousterhout: what do you mean by "normally" here?

minion/scheduler/worker_test.go
```diff
@@ -90,40 +110,49 @@ func TestSyncWorker(t *testing.T) {
...
 
+	// Assert that the pairing specified in `changed` is consistent with the
+	// desired container in the database.
 	dbcs[0].DockerID = dkcs[0].ID
 	assert.Equal(t, dbcs, changed)
 
-	// Atempt a failed remove
+	// Change the desired containers to be empty. Any running containers should
+	// be stopped.
+	// However, the running container is not actually removed because we mock
```
kayousterhout: weird line wrapping?

minion/scheduler/worker_test.go
kayousterhout: Also this comment is at high risk of becoming out of date (since it's not right next to the corresponding code), and I'm not sure it's very useful given the detail below.  I might just make this a very brief summary: "This test verifies that the worker synchronizes with the database correctly, by booting or destroying containers as appropriate."

minion/scheduler/worker_test.go
kklin: As in, when the function runs outside a test, the value is obtained by querying the database. Maybe this isn't necessary context? I wanted to clarify since I refer to "database containers" and "docker containers" in other comments.

kayousterhout: How about "The containers that should be running, according to the database.  We populate this manually here to simulate different states of the database."

kklin: Done!

PR #1162, opened 2017-10-14
cloud: Sync TLS certificate mappings to minions

kklin:


kayousterhout:
@kklin can you resolve the merge conflict here?  I'll review tomorrow AM.

cloud/credentials.go
```diff
@@ -33,42 +33,48 @@ var credentialsCounter = counter.New("Cloud Credentials")
...
+		if !ok {
 			continue
 		}
 
-		credentialsCounter.Inc("Install " + m.PublicIP)
-		if generateAndInstallCerts(m, sshKey, ca) {
-			credentialedMachines[m.PublicIP] = struct{}{}
-		}
+		conn.Txn(db.MachineTable).Run(func(view db.Database) error {
+			for _, dbm := range view.SelectFromMachine(nil) {
```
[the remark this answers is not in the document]
kayousterhout: Also is the fact that this is n^2 a concern?  Worth adding a comment about?

cloud/credentials.go
```diff
@@ -33,42 +33,48 @@ var credentialsCounter = counter.New("Cloud Credentials")
...
+		syncCredentialsOnce(conn, sshKey, ca)
 	}
 }
 
-func syncCredentialsOnce(sshKey ssh.Signer, ca rsa.KeyPair,
-	machines []db.Machine, credentialedMachines map[string]struct{}) {
+func syncCredentialsOnce(conn db.Conn, sshKey ssh.Signer, ca rsa.KeyPair) {
 	credentialsCounter.Inc("Install to cluster")
+	machines := conn.SelectFromMachine(func(m db.Machine) bool {
+		return m.Status != db.Stopping && m.PublicIP != "" && m.PublicKey == ""
```
kayousterhout: I wonder if it would help readability if you called this "InstalledPublicKey" instead of just "PublicKey", to make the semantics here (that once this is set, the key has been installed) more obvious.  Feel free to ignore if you don't like this idea.

cloud/credentials_test.go
```diff
@@ -55,30 +59,32 @@ func TestSyncCredentials(t *testing.T) {
...
 		return nil, assert.AnError
 	}
-	credentialedMachines = map[string]struct{}{}
-	syncCredentialsOnce(nil, ca, []db.Machine{
-		{Role: db.Worker, PublicIP: "8.8.8.8"},
-	}, credentialedMachines)
-	assert.Empty(t, credentialedMachines)
+
+	syncCredentialsOnce(conn, nil, ca)
+	assert.Empty(t, conn.SelectFromMachine(nil)[0].PublicKey)
```
kayousterhout: Would be super helpful to add a comment here saying something like "The machine's PublicKey should not be set, because Quilt should have given up and not set the public key when getting an SFTP client failed." (or the alternate renaming I suggested above would also help with this issue)

db/machine.go
```diff
@@ -26,7 +26,8 @@ type Machine struct {
 	PrivateIP string
 
 	/* Populated by the cluster. */
-	Status string
+	Status    string
+	PublicKey string
```
kayousterhout: Especially if you don't do the re-naming I suggested, can you add a comment about the invariants of this variable (that if it's set, it means that the given key has been installed on the machine in the cluster)

minion/pb/pb.proto
```diff
@@ -22,6 +22,7 @@ message MinionConfig {
     string FloatingIP = 8;
     repeated string EtcdMembers = 9;
     repeated string AuthorizedKeys = 10;
+    map<string, string> MinionToPublicKey = 11;
```
kayousterhout: Can you add the key type to the name -- so MinionIPToPublicKey? (I think MinionPrivateIPToPublicKey is probably overkill)

It seems helpful given existing ambiguity about the identifier we use to reference things (e.g., various IDs are used elsewhere)

cloud/credentials_test.go
```diff
@@ -63,6 +63,45 @@ func TestSyncCredentials(t *testing.T) {
 	assert.Equal(t, string(certBytes), dbm.PublicKey)
 }
 
+// Test that we don't overwrite existing credentials.
+func TestCredentialsPresent(t *testing.T) {
```
kayousterhout: FWIW I'm a fan of just using long test names, because it's nice that they show up in the test output, so TestExistingCredentialsDontGetOverwritten or something (FFTI)

connection/tls/io/io.go
```diff
@@ -69,35 +69,35 @@ func ReadCA(dir string) (rsa.KeyPair, error) {
...
-func caKeyPath(dir string) string {
+// CAKeyPath defines where to write the private key for the certificate authority.
+func CAKeyPath(dir string) string {
 	return filepath.Join(dir, caKeyFilename)
 }
 
-// signedCertPath defines where to write the certificate for the signed certificate.
-func signedCertPath(dir string) string {
+// SignedCertPath defines where to write the certificate for the signed certificate.
+func SignedCertPath(dir string) string {
```
kayousterhout: This was the only function I saw used in your later commit -- are the other functions used in a later commit that's not in this PR?  Or did you decide to just make them all public in case of similar future use?

kklin: The Vault commit makes use of it

cloud/credentials.go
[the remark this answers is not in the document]
kklin: I don't think the n^2 matters. N is usually small, and the size of N will decrease as we successfully install certificates.

cloud/credentials.go
kklin: I left it as is because I prefer the readability of shorter names (but I did add the comment to the Machine row as you suggested)

cloud/credentials.go
```diff
@@ -33,42 +33,48 @@ var credentialsCounter = counter.New("Cloud Credentials")
...
-			return m.Status != db.Stopping && m.PublicIP != ""
-		})
-		syncCredentialsOnce(sshKey, ca, machines, credentialedMachines)
+		syncCredentialsOnce(conn, sshKey, ca)
 	}
 }
 
-func syncCredentialsOnce(sshKey ssh.Signer, ca rsa.KeyPair,
-	machines []db.Machine, credentialedMachines map[string]struct{}) {
+func syncCredentialsOnce(conn db.Conn, sshKey ssh.Signer, ca rsa.KeyPair) {
```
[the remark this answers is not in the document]
kklin: I wasn't sure what to add here that isn't covered by the SyncCredentials comment https://github.com/kklin/kelda/blob/be6bfb89148422b113f9c69ea16f341ec6e9d2b6/cloud/credentials.go#L24. Should I move some of that comment here? Or is there some missing details?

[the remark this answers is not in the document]
kayousterhout: With the new comments you added below I agree that this isn't necessary

cloud/credentials_test.go
```diff
@@ -82,6 +119,8 @@ func TestSyncCredentialsSkip(t *testing.T) {
 	}
 
 	syncCredentialsOnce(conn, nil, ca)
+	// The machine's PublicKey should not be set, because Kelda should have
```
kayousterhout: You added this in the wrong commit -- although not sure it matters enough to deal with fixing it

kklin: Ah thanks for noticing this, fixed

PR #1163, opened 2017-10-14
Test syncing Docker container attributes with database

kklin:


minion/docker/mock_docker.go
```diff
@@ -280,7 +280,12 @@ func (dk MockClient) InspectContainer(id string) (*dkc.Container, error) {
 		return nil, ErrNoSuchContainer
 	}
 
-	return container.Container, nil
+	containerCopy := container.Container
```
luise: Why the copy? Isn't it still a reference to the same underlying container?

luise:
@kklin , this also needs a rebase. Let me know if you think I should take another look after you rebase

minion/docker/mock_docker.go
kklin: Ah, I didn't realize that container.Container was a pointer. I'll rename this variable to not incorrectly imply that it's a copy.

PR #1166, opened 2017-10-16
cloud: Fix join when using floating IPs

kklin:
Before this commit, the cloud join fetched the roles of machines using
the public IP as a unique ID. However, because a floating IP replaces the
public IP of machine, the public IP is not a suitable identifier for a
machine. When a floating IP was reassigned from a Master to a Worker,
the public IP of the machine migrated as well, causing the foreman to
incorrectly report that the Worker was a Master. This broke the join.

This commit fixes this bug by modifying the foreman to use the machine's
CloudID as an identifier, rather than public IP. This is correct because
the CloudID is always associated with the same machine.


cloud/foreman/foreman.go
```diff
@@ -134,30 +134,30 @@ func RunOnce(conn db.Conn) {
 }
 
 // GetMachineRole uses the minion map to find the associated minion with
-// the IP, according to the foreman's last update cycle.
-func GetMachineRole(pubIP string) db.Role {
-	if min, ok := minions[pubIP]; ok {
+// the cloudID, according to the foreman's last update cycle.
```
kayousterhout: Can you add a comment on line 20 about the keys in the minion map?

PR #1167, opened 2017-10-16
 ci: Fix references to machine ID

kklin:
This PR fixes some integration tests that are failing to compile, and fixes the Makefile so that this will cause Travis to fail in the future.


integration-tester/Makefile
```diff
@@ -1,9 +1,12 @@
-SHELL := /bin/bash
+# Get the names of the test suites by getting the relative path to all
+# *_test.go files, getting the parent directory, and stripping the leading
+# path. For example, tests/100-logs/logs_test.go gets converted to
+# tests/100-logs, and finally 100-logs.
```
kayousterhout: Can you add in the commit message that this commit also fixes the error that currently happens for 60-duplicate-cluster-setup?

PR #1168, opened 2017-10-17
Simple name migrations

ejj:


minion/docker/docker.go
```diff
@@ -132,7 +133,7 @@ func (dk Client) Run(opts RunOptions) (string, error) {
 	if opts.IP != "" {
 		nc = &dkc.NetworkingConfig{
 			EndpointsConfig: map[string]*dkc.EndpointConfig{
-				"quilt": {
+				plugin.NetworkName: {
```
kklin: FYI the build is failing because I merged a commit that references "quilt" here https://github.com/kelda/kelda/blob/master/minion/docker/mock_docker.go#L314 (or as you've done here, it should reference plugin.NetworkName)

PR #1170, opened 2017-10-17
bindings.js: Add more JSdoc and use ES6 classes

kayousterhout:
This PR does two major cleanups to bindings.js:
- It adds more JSDoc.  This should be interpreted as a first cut of the JSDoc.  It's enough that the linter can now check for all JSDoc (which will force us to add JSDoc to new code) but some of the documentation should probably be improved in the future.  This should be viewed as a first cut and not an opportunity to perfect this documentation (although of course we shouldn't have any major errors).
- It uses ES6 classes, which groups the code more nicely and makes it easier to read.

@ejj it would be great to merge this before you do the JS rename, because this commit changes many many lines in bindings.js that will conflict with the rename.


kayousterhout:
Also as a warning the git diff for the commit to move to ES6 classes is quite unhelpful.  That commit includes no changes other than moving around code to use ES6 classes.

blueprint/bindings.js
```diff
@@ -443,6 +467,13 @@ class LoadBalancer {
  * @implements {Connectable}
  */
 const publicInternet = {
+  /**
+   * @param {Container|Container[]} srcArg - A Container or list of Containers
+   *   that should be allowed to connect to the public internet.
```
kklin: I think this is indented one extra space?

blueprint/bindings.js
```diff
@@ -899,6 +952,10 @@ class Machine {
     return res;
   }
 
+  /**
+   * @private
+   * @returns {String} A string describing all attributes of the machine.
```
kklin: Should "String" be lowercase?

blueprint/bindings.js
kklin: (or line 474 is not indented enough)

blueprint/bindings.js
```diff
@@ -1080,6 +1159,10 @@ class Container {
     return `${this.hostname}.q`;
   }
 
+  /**
+   * @private
+   * @returns {String} A string describing all attributes of the machine.
```
kklin: Same here re lowercase string

blueprint/bindings.js
```diff
@@ -1035,15 +1095,34 @@ class Container {
...
+   * @returns {Container} A new Container with the same attributes.
+   */
   clone() {
     return new Container(this.hostnamePrefix, this.image, this);
   }
 
+  /**
+   * Sets the given environment variable to the given value.
+   *
+   * @param {String} key - The name of the environment variable to set.
```
kklin: Same here re "String"s

kayousterhout:
Thanks for the quick review @kklin! I updated this and added one new commit to enforce consistent capitalization.

PR #1174, opened 2017-10-18
readme: Migrate the README.md from Quilt to Kelda

ejj:


scripts/blueprints-tester/tests/test_readme.go
```diff
@@ -88,8 +88,8 @@ func (parser readmeParser) blocks() (map[string]string, error) {
 
 var dependencies = `{
   "dependencies": {
-    "@quilt/quilt": "quilt/quilt",
-    "@quilt/redis": "quilt/redis"
+    "kelda": "kelda",
```
kklin: This should be `"kelda": "0.x"`

ejj:
Note that this PR also contains the commits from https://github.com/kelda/kelda/pull/1177  We think both are necessary for travis to be happy.

scripts/blueprints-tester/tests/test_readme.go
ejj: doh
thanks

PR #1177, opened 2017-10-18
ci: Update pinned Django revision

kklin:
The pinned version was before the Kelda rename.


kklin:
The commits in this PR was merged in the #1174.

PR #1179, opened 2017-10-18
More small name updates

ejj:


cloud/amazon/amazon.go
```diff
@@ -555,7 +555,7 @@ func (prvdr *Provider) getCreateSecurityGroup() (
 		return *groups[0].GroupId, groups[0].IpPermissions, nil
 	}
 
-	id, err := prvdr.CreateSecurityGroup(prvdr.namespace, "Quilt Group")
+	id, err := prvdr.CreateSecurityGroup(prvdr.namespace, "Kelda Group")
```
kklin: Is this going to do anything to existing instances? (I'm mostly worried about leaking instances because of the rename)

ejj: Nah, that's just a description.  Has no actual effect.

PR #1180, opened 2017-10-18
docs: Fix path to bindings.js

kayousterhout:
A recent commit moved bindings.js to a different directory, which
broke the docs build.  This commit updates the path in the docs
Makefile.


kayousterhout:
Did this in a separate PR so that I can make sure it's working correctly there by verifying that the Travis build on that PR fails: https://github.com/kelda/kelda/pull/1181

PR #1181, opened 2017-10-18
travis: Build docs

kayousterhout:
This commit adds functionality to Travis to build our docs, to make
sure to catch docs build failures.


kayousterhout:
I didn't test that it fails but I checked that Travis is compiling the docs as expected, and the Make command will exit with 1 when it fails.

PR #1185, opened 2017-10-18
machine: Remove CPU and RAM from machine attributes

HongjianZhang:
Before this commit, CPU and RAM are machine attributes. This has
two problems.

1. More information has to be passed between JS and Go.
2. A user might modify a machine's CPU and RAM after creating it.

This commit fixes it, by removing CPU and RAM from machine. The
constructor takes in CPU and RAM as arguments and passed them to
chooseSize. Instead of CPU and RAM, JS only needs to pass Size to
Go.


js/bindings/bindings.js
```diff
@@ -596,7 +596,8 @@ function checkExtraKeys(opts, obj) {
   // refID set in the options).
   if (objectHasKey.call(opts, '_refID')) { return; }
 
-  const extras = Object.keys(opts).filter(key => !objectHasKey.call(obj, key));
+  const extras = Object.keys(opts).filter(key => !objectHasKey.call(obj, key)
```
kayousterhout: Can you change the commit message to be "bindings: ...."? (that's what we usually use to refer to changes to the JS blueprint code; I should have noticed this in your previous PR too oops!)

js/bindings/bindings.js
```diff
@@ -596,7 +596,8 @@ function checkExtraKeys(opts, obj) {
...
+                                                  && key !== 'cpu' && key !== 'ram');
```
kayousterhout: we usually use 2-space indents in the JS (so this should be 2 spaces indented from the previous line's indentation level)

kayousterhout: Also by adding this check here, for other objects (other than Machine), we'll ignore it if those have a 'cpu' or 'ram' attribute, which is not ideal.  Can you move this special casing for the cpu/ram attributes so that it only is done for Machine?  It might make sense to write a special version of this code in the Machine constructor.

js/bindings/bindings.js
```diff
@@ -781,13 +782,12 @@ class Machine {
...
-    this.cpu = boxRange(optionalArgs.cpu);
-    this.ram = boxRange(optionalArgs.ram);
     this.preemptible = getBoolean('preemptible', optionalArgs.preemptible);
-
+    const cpu = boxRange(optionalArgs.cpu);
+    const ram = boxRange(optionalArgs.ram);
     checkExtraKeys(optionalArgs, this);
 
-    this.chooseSize();
+    this.chooseSize(cpu, ram);
```
kayousterhout: can you move lines 786 and 787 to be right above this line?  Just makes the code a little easier to read.  You could also inline them in this function call, so `this.chooseSize(boxRange(optionalArgs.cpu), boxRange(optionalArgs.ram))`

js/bindings/bindings.js
```diff
@@ -797,22 +797,24 @@ class Machine {
    * an error if provider is not valid or given machine requirements cannot
    * be satisfied by any size.
    * @private
+   * @param {Range} cpu - The desired number of CPUs.
```
kayousterhout: Can you change to "The desired number of CPU cores."

js/bindings/bindings.js
```diff
@@ -844,16 +846,18 @@ class Machine {
    * @private
    * @param {description[]} providerDescriptions - Array of descriptions of
    *   a provider.
+   * @param {Range} cpu - The desired number of CPUs.
```
kayousterhout: ditto re: comment above

js/bindings/bindings.js
kklin: (I told either Luke or Eli to use "js". But I'm fine with using "bindings" from now on. Sorry for the miscommunication!)

js/bindings/bindings.js
```diff
@@ -781,38 +781,71 @@ class Machine {
...
+  * of the object must be an unexpected optional argument. However since CPU and RAM 
+  * are valid user arguments but not Machine properties, we do not consider them to
+  * be invalid.
+  * @private
+  * @param {Object} opts - The Object to check for redundant keys.
+  * @returns {void}
+  * @throws Throws an error if redundant keys are found in `opts`.
+  */
+  checkExtraKeys(opts) {
+    // Sometimes, prototype constructors are called internally by Kelda. In these
```
kayousterhout: In this case, I'd just refer to the other comment to avoid duplication (so `// Refer to the comment in the global checkExtraKeys method about the reason for this check.`

js/bindings/bindings.js
```diff
@@ -761,7 +761,7 @@ class Machine {
    * @param {string} [optionalArgs.region] - The region the machine will run-in
    *   (provider-specific; e.g., for Amazon, this could be 'us-west-2').
    * @param {string} [optionalArgs.size] - The instance type (provider-specific).
-   * @param {Range|int} [optionalArgs.cpu] - The desired number of CPUs.
+   * @param {Range|int} [optionalArgs.cpu] - The desired number of CPU cores.
```
ejj: This change isn't a problem in and of itself, but it doesn't really have anything to do with removing RAM and CPU attributes -- it's an unrelated documentation change.  Thus, generally, it would be better to do it in a separate patch.

js/bindings/bindings.js
```diff
@@ -781,38 +781,66 @@ class Machine {
...
+  * object, any of those that appears in the optional argument but not as a property
+  * of the object must be an unexpected optional argument. However since CPU and RAM 
+  * are valid user arguments but not Machine properties, we do not consider them to
+  * be invalid.
+  * @private
+  * @param {Object} opts - The Object to check for redundant keys.
+  * @returns {void}
+  * @throws Throws an error if redundant keys are found in `opts`.
+  */
+  checkExtraKeys(opts) {
```
ejj: I really don't like that we're copying this function.  Could you please put a bit more thought into how to reuse the existing checkExtraKeys function?  I think if you think about it you could find a way.  Perhaps by adding `ram` and `cpu` to `this` before passing it in.

Anyways, Generally duplicate code is a a bad thing.

HongjianZhang:
In order to reuse the same checkExtraKeys, an easy way around is to temporarily add cpu and ram to this, and remove them after checkExtraKeys. It seems a little hacky to me, but I cannot think of a better approach. What do you think?

PR #1186, opened 2017-10-18
docs: Verify bundle is installed before building

ejj:
If bundler isn't installed, building the docs can get into a weird
broken state that's hard to get out of.  This patch checks bundler
before doing anything else as a precaution.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,5 +1,6 @@
 # Compiles all of the docs in docs/build.
 all: build/slate _jsdoc.erb
+	bundle --version && \
 	cd build/slate && \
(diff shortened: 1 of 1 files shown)


kayousterhout:
Should we use .DELETE_ON_ERROR for `build` instead? That seems more general to other potential issues that could happen when creating the build directory

docs/Makefile
```diff
@@ -1,5 +1,6 @@
 # Compiles all of the docs in docs/build.
 all: build/slate _jsdoc.erb
+	bundle --version && \
```
kayousterhout: Why do you need this one given the one in build/slate?

PR #1188, opened 2017-10-18
Implement secure Secret storage

kklin:


minion/scheduler/worker_test.go
```diff
@@ -62,12 +72,56 @@ func TestRunWorker(t *testing.T) {
...
+	secretVal := "secretVal"
+	conn.Txn(db.ContainerTable).Run(func(view db.Database) error {
+		dbc := view.SelectFromContainer(nil)[0]
+		dbc.Env = map[string]blueprint.SecretOrString{
+			envKey: blueprint.NewSecretOrString(secretName, true),
+		}
+		view.Commit(dbc)
+		return nil
+	})
+	mockVault.On("Read", secretName).Return("", vault.ErrSecretDoesNotExist).Twice()
```
kayousterhout: Why twice? (I thought it would get cached after the first time?)

minion/scheduler/worker_test.go
```diff
@@ -117,20 +173,21 @@ func TestSyncWorker(t *testing.T) {
 
 	// Assert that the pairing specified in `changed` is consistent with the
 	// desired container in the database.
-	dbcs[0].DockerID = dkcs[0].ID
-	dbcs[0].Status = dkcs[0].Status
-	assert.Equal(t, dbcs, changed)
+	evaluatedDbcs[0].DockerID = dkcs[0].ID
+	evaluatedDbcs[0].Status = dkcs[0].Status
+	assert.Len(t, changed, 1)
+	assert.Equal(t, evaluatedDbcs[0].Container, changed[0])
```
kayousterhout: why'd this change? (from checking that the whole list was equal) . Does equality not work when there's the extra layer of nesting?

api/server/server.go
```diff
@@ -70,6 +71,27 @@ func Run(conn db.Conn, listenAddr string, runningOnDaemon bool,
...
+		machines := s.conn.SelectFromMachine(nil)
+		leaderClient, err := newLeaderClient(machines, s.clientCreds)
+		if err != nil {
+			return &pb.SecretReply{}, err
+		}
+		defer leaderClient.Close()
+		return &pb.SecretReply{}, leaderClient.SetSecret(msg.Name, msg.Value)
+	}
+
+	client, err := vaultNew(s.conn.MinionSelf().PrivateIP)
```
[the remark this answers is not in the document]
kklin: The scheduler code makes use of the `SecretStore.Read` method, and I don't think this change would make as much sense there since we reuse the client for multiple `Read` calls

blueprint/blueprint.go
```diff
@@ -44,12 +44,64 @@ type Image struct {
...
+// maps. If the value is a secret, then the name of the secret is returned, and
+// the caller is expected to query Vault to resolve the secret value. Otherwise,
+// the value is used directly.
+type SecretOrString struct {
+	// Store the fields within a private member variable so that we can export
+	// them and use Go's automatic JSON marshalling and unmarshalling. We don't
+	// store the fields directly in the SecretOrString struct because then
+	// callers could bypass the Value() function and access the Value directly
+	// without checking whether it's a secret.
+	jsonObj privateSecretOrString
```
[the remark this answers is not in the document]
kklin: Go's JSON marshalling doesn't support unexported fields. (I added this exact comment into the code)

minion/vault/api.go
```diff
@@ -0,0 +1,131 @@
...
+
+	tlsIO "github.com/kelda/kelda/connection/tls/io"
+	"github.com/kelda/kelda/counter"
+
+	vaultAPI "github.com/hashicorp/vault/api"
+)
+
+var vaultCounter = counter.New("Vault")
+
+// APIClient is an interface for interacting with the Vault API. It is only used
```
kayousterhout: What do you mean when you say it's only used for unit testing?  It's implemented by vaultAPIClientImpl right?  Is the point here that the only reason to codify this in an interface is to generate mocks for unit testing?

blueprint/blueprint.go
```diff
@@ -44,12 +44,64 @@ type Image struct {
...
+func (secretOrString SecretOrString) String() string {
+	val, isSecret := secretOrString.Value()
+	if isSecret {
+		return "Secret: " + val
+	}
+	return val
+}
+
+// NewSecretOrString creates a new SecretOrString instance.
+func NewSecretOrString(val string, isSecret bool) SecretOrString {
```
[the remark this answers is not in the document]
kklin: I like that, I'll do it

minion/vault/api.go
```diff
@@ -0,0 +1,131 @@
...
+	vaultCounter.Inc(fmt.Sprintf("Write Path %q", path))
+	return c.client.Logical().Write(path, data)
+}
+
+func (c vaultAPIClientImpl) Delete(path string) (*vaultAPI.Secret, error) {
+	vaultCounter.Inc(fmt.Sprintf("Delete Path %q", path))
+	return c.client.Logical().Delete(path)
+}
+
+func (c vaultAPIClientImpl) Init(opts *vaultAPI.InitRequest) (*vaultAPI.InitResponse,
```
kayousterhout: Can you add comments for this and the below two methods, about what they do and when they need to be called? (These 3 are particularly non-obvious -- and a comment seems useful for your own future reference in updating this, if nothing else)

minion/vault/auth.go
```diff
@@ -0,0 +1,128 @@
...
+	joinRoles(vaultClient, getDesiredRoles(conn), currentRoles)
+}
+
+func getCurrentRoles(vaultClient APIClient) ([]certRole, error) {
+	listAuthResp, err := vaultClient.List(certListEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	// If no roles have been configured, the LIST will return nil.
```
kayousterhout: why the all caps LIST?

db/container.go
```diff
@@ -14,23 +15,39 @@ import (
...
 
+// GetReferencedSecrets returns the names of all Secrets referenced in the Env
+// and FilepathToContent maps.
+func (c Container) GetReferencedSecrets() []string {
+	return append(getReferencedSecrets(c.Env),
+		getReferencedSecrets(c.FilepathToContent)...)
+}
+
+func getReferencedSecrets(x map[string]blueprint.SecretOrString) (secrets []string) {
+	for _, possSecret := range x {
```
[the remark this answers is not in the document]
kklin: haha

minion/vault/auth.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+var (
+	certMountName     = "cert"
+	certRoot          = path.Join("/auth", certMountName)
+	certLoginEndpoint = path.Join(certRoot, "login")
+	certListEndpoint  = path.Join(certRoot, "certs")
+)
+
+type certRole struct {
```
kayousterhout: Can you add a comment about what this is used for? (e.g., looks like every authenticated user (?) needs a public key that Vault will use to authenticate connections from that user, which is cert, and a set of policies, which are described in ?? format -- here maybe just refer to policy.go)

minion/vault/auth.go
```diff
@@ -0,0 +1,128 @@
...
+	"path"
+	"strings"
+
+	"github.com/kelda/kelda/db"
+	"github.com/kelda/kelda/join"
+
+	log "github.com/sirupsen/logrus"
+)
+
+var (
```
kayousterhout: Can you add a comment that these are the Vault paths where vault stuff (? which stuff) is stored? (both to make it clear what these are used for, and also to make it clear that these are paths in Vault and not in the local filesystem)

minion/scheduler/worker_test.go
kklin: This line actually should be twice since the Read fails, but the one below is only once. Good catch! I added an `AssertExpectations` call which would have caught this

minion/scheduler/worker.go
```diff
@@ -48,13 +73,61 @@ func runWorker(conn db.Conn, dk docker.Client, myIP string) {
...
+				if err == nil {
+					secretMap[name] = secretVal
+					continue
+				}
+
+				// It is expected for secrets to not exist in Vault
+				// before users run the `kelda secret` command.
+				if err != vault.ErrSecretDoesNotExist {
+					log.WithError(err).WithField("name", name).
+						Warn("Failed to fetch secret")
```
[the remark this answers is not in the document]
kklin: This error message is printed for when the error is for something _other_ than the user not running `kelda secret` (e.g. if Vault was improperly configured). The logs would be pretty noisy otherwise.

Needing to run `kelda secret` is conveyed by the container Status, so maybe we should add it there? (although I do think the status would get pretty long)

minion/vault/auth_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+func TestGetCurrentRolesError(t *testing.T) {
+	t.Parallel()
+	mockClient := &mocks.APIClient{}
+
+	mockClient.On("List", certListEndpoint).Return(nil, assert.AnError).Once()
+	_, err := getCurrentRoles(mockClient)
+	assert.NotNil(t, err)
+
+	mockClient.On("List", certListEndpoint).Return(
```
kayousterhout: What's the problem here? The reply from List doesn't have the expected attributes?

minion/etcd/container_test.go
```diff
@@ -89,14 +107,19 @@ func TestRunContainerOnce(t *testing.T) {
...
-		Env:               map[string]string{"red": "pill", "blue": "pill"},
-		FilepathToContent: map[string]string{"foo": "bar"},
-		Hostname:          "host",
+		IP:          "10.0.0.2",
+		BlueprintID: "12",
+		Minion:      "1.2.3.4",
+		Image:       "ubuntu",
+		Command:     []string{"1", "2", "3"},
+		Env: map[string]blueprint.SecretOrString{
+			"red":  blueprint.NewSecretOrString("pill", true),
```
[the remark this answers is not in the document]
kklin: The fish assignment is to the minion's in-memory database, but etcd (which still contains the `{red: pill}` association) is the source of truth. So the expected behavior is that the local database will sync with what's in etcd.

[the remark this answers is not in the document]
kklin: (This test is extremely hard to follow)

minion/scheduler/worker.go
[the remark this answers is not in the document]
kayousterhout: Ahh I see I mis-interpreted the comment.  Would be nice to add the 'kelda secret' recommendation in the status unless you think it looks horrifically long, since it seems better there than for the user to have to poke around in the docs.

Can you add a sentence to the comment saying something like what you wrote above -- "If the error is for something _other_ than the use not running `kelda secret`, log a warning."

api/server/server.go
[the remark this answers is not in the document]
kayousterhout: I was wondering about that as I read more of the code, and I can see a consistency argument for keeping them the same.

minion/etcd/container_test.go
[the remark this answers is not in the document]
kayousterhout: Would be great to add a comment saying exactly what you said in your comment above (I realize unrelated to your change but your comment was super helpful and seems quick to add now that you've written it)

minion/vault/policy.go
```diff
@@ -0,0 +1,137 @@
...
+	"encoding/json"
+
+	"github.com/kelda/kelda/db"
+	"github.com/kelda/kelda/join"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type vaultPolicy struct {
+	name, policy string
```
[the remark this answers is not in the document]
kayousterhout: It might also be helpful here to have a comment like "For more about policies and their role in the overall Vault structure, refer to vault.go." (I understood this much better once I read your long comment there)

minion/vault/vault.go
```diff
@@ -0,0 +1,192 @@
...
+* In our design, there is a policy for each minion, and exactly one
+* authentication role that maps to the minion's policy. Authentication is
+* performed using TLS certificates. The daemon reports the minion to public key
+* mappings to the masters, and the masters create authentication rules that
+* require clients to possess the associated private key.
+**/
+
+// Run implements the logic necessary for the masters to run Vault. It blocks
+// until Vault is booted, and then continuously configures the Vault access
+// policies.
```
kayousterhout: It would be great to give a high level description of the role of this: something like "This function continuously checks the Kelda database to determine the policies that Vault should implement for secret access, and updates Vault to reflect those policies."

minion/vault/vault.go
```diff
@@ -0,0 +1,192 @@
...
+// tryStartVaultUntilSuccess repeatedly tries to start and bootstrap the Vault
+// container until it succeeds. This way, we recover from any transient errors
+// that might occur when starting Vault, such as a failure to pull the Vault
+// Docker image.
+func tryStartVaultUntilSuccess(dk docker.Client, listenAddr string) APIClient {
+	for {
+		if client, ok := startAndBootstrapVault(dk, listenAddr); ok {
+			return client
+		}
+		log.Warn("Failed to start Vault. Retrying in 30 seconds")
```
kayousterhout: Can you add a timeout here? (something generous -- like many minutes -- that you expect only to fail when it really didn't work, but it's helpful to have an explicit failure rather than waiting forever)

minion/vault/vault.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	vaultClient := tryStartVaultUntilSuccess(dk, myIP)
+
+	for range conn.TriggerTick(30, db.MinionTable, db.ContainerTable).C {
+		syncPolicies(vaultClient, conn)
+		syncAuth(vaultClient, conn)
+	}
+}
+
+const vaultContainerName = "vault"
```
kayousterhout: Can you comment that this is the name of the vault container stored in DockerHub? (I thought this might just be a kelda-internal name for the container)

minion/vault/vault.go
```diff
@@ -0,0 +1,192 @@
...
+		if err == nil {
+			return true
+		}
+
+		log.WithError(err).Debug("Failed to connect to Vault. " +
+			"This is expected when the container first boots.")
+		return false
+	}, 10*time.Second, 5*time.Minute)
+	if err != nil {
+		log.WithError(err).Error("Failed to connect to Vault")
```
kayousterhout: if it's not too much work, can you save the most recent error and print it here? (my understanding is that `err` will just be something saying that the 5 minute timeout expired, rather than saying the actual underlying problem connecting, which may be more useful)

minion/vault/vault.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, false
+	}
+
+	// Ensure that Vault is unsealed by generating an unseal key using the Init
+	// endpoint, and then Unsealing with the generated key.
+	initResp, err := client.Init(&vaultAPI.InitRequest{
```
kayousterhout: What does this mean for when this container gets re-started? Is it safe to re-call init? (can you document this?)

minion/vault/vault.go
```diff
@@ -0,0 +1,192 @@
...
+*
+* In our design, there is a policy for each minion, and exactly one
+* authentication role that maps to the minion's policy. Authentication is
+* performed using TLS certificates. The daemon reports the minion to public key
+* mappings to the masters, and the masters create authentication rules that
+* require clients to possess the associated private key.
+**/
+
+// Run implements the logic necessary for the masters to run Vault. It blocks
+// until Vault is booted, and then continuously configures the Vault access
```
kayousterhout: Can you also add that this actually boots Vault and also expand a bit on the last part -- something like "It boots a Vault container, blocks until Vault is booted, and then continuously checks the Kelda database to determine the policies that Vault should implement for secret access, and updates Vault to reflect those policies."

minion/run.go
```diff
@@ -83,6 +84,9 @@ func Run(role db.Role, inboundPubIntf, outboundPubIntf string) {
 		return
 	}
 
+	// Vault must be started after the credentials are resolved because it also
+	// makes use of the credentials.
+	go vault.Run(conn, dk)
```
kayousterhout: What happens if an application container is launched (and requires a secret) before Vault finishes coming up?

minion/vault/vault_test.go
```diff
@@ -0,0 +1,129 @@
...
+	assert.NoError(t, err)
+
+	// Ensure that the referenced TLS credentials in the config were properly
+	// uploaded to the container.
+	listenerConfig := parsedConfig["listener"].(map[string]interface{})
+	tcpConfig := listenerConfig["tcp"].(map[string]interface{})
+	serverCertPath := tcpConfig["tls_cert_file"].(string)
+	serverKeyPath := tcpConfig["tls_key_file"].(string)
+	caCertPath := tcpConfig["tls_client_ca_file"].(string)
+
```
kayousterhout: Looks like this is making sure that all of the files were properly added to the container? (add a comment here?)

minion/vault/vault.go
```diff
@@ -0,0 +1,192 @@
...
+	}, 10*time.Second, 5*time.Minute)
+	if err != nil {
+		log.WithError(err).Error("Failed to connect to Vault")
+		return nil, false
+	}
+
+	// Ensure that Vault is unsealed by generating an unseal key using the Init
+	// endpoint, and then Unsealing with the generated key.
+	initResp, err := client.Init(&vaultAPI.InitRequest{
+		SecretShares:    1,
```
kayousterhout: Can you add a brief description of what these params mean?

js/bindings/bindings.js
```diff
@@ -1341,14 +1343,71 @@ class Container {
...
-      filepathToContent: this.filepathToContent,
+      env: boxSecretOrStringMap(this.env),
+      filepathToContent: boxSecretOrStringMap(this.filepathToContent),
       hostname: this.hostname,
     };
   }
 }
 
 /**
+ * boxSecretOrStringMap takes a map with values that are either a string, or Secret,
```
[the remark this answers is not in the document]
kklin: This was function was removed entirely (yay!)

minion/scheduler/worker_test.go
kklin: Yeah, exactly. The types don't match up, plus `evaluatedDbcs` has some extra fields.

minion/vault/api.go
kklin: Yeah, I tweaked the wording to hopefully make this a little clearer:
```
// APIClient is an interface for interacting with the Vault API. It is only
// used for unit testing -- an interface must be exported in order to generate
// a mock with mockery.
```

minion/vault/auth.go
kklin: I thought it was an HTTP verb -- I'll make it lowercase.

minion/vault/auth_test.go
kklin: The error is that `Read` is mocked to return an error

kklin: I added a comment.

minion/run.go
kklin: Hm so the worker will log an error that it failed to read the secret (e.g because the connection was rejected). I'm actually not sure what we should do here. One option would be to not schedule containers until Vault boots?

minion/vault/policy.go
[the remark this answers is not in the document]
kklin: As a sidenote, I realized while writing the documentation for the `vaultPolicy` and `certRole` structs how the comment ties the entire file together. It really helps pull together the entire file. I kind of feel myself developing an intuition on what sort of _things_ really benefit from these high level comments (data structures, public facing methods)...

minion/vault/policy.go
```diff
@@ -0,0 +1,137 @@
...
+		if m.Role == db.Master {
+			desiredPolicies = append(desiredPolicies, vaultPolicy{
+				name:   m.PrivateIP,
+				policy: writeAllPolicy,
+			})
+		}
+	}
+
+	// Process the scheduled containers to decide what minions should be allowed
+	// to read which secrets.
```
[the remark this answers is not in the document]
kklin: BTW this feedback is really helpful to me -- sometimes I'm unsure whether the comment I'm writing is actually useful (and thus whether I should write it).

minion/vault/store.go
```diff
@@ -0,0 +1,74 @@
...
+	if !ok {
+		return "", errors.New("malformed secret")
+	}
+
+	return secret.(string), nil
+}
+
+func (client secretStoreImpl) Write(name, value string) error {
+	_, err := client.vaultClient.Write(pathForSecret(name),
+		map[string]interface{}{secretKey: value})
```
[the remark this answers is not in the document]
kklin: Yeah, exactly

minion/vault/store_test.go
```diff
@@ -0,0 +1,58 @@
...
+	mockClient.On("Read", pathForSecret(secretName)).Return(
+		&vaultAPI.Secret{
+			Data: map[string]interface{}{
+				secretKey: secretValue,
+			},
+		}, nil,
+	).Once()
+	actualValue, err := store.Read(secretName)
+	assert.Nil(t, err)
+	assert.Equal(t, secretValue, actualValue)
```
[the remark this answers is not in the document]
kklin: It doesn't hurt to check so I added it, but the reason why I didn't check originally was that I wasn't thinking about the number of times the method was called while writing the test. The `Once()` call is required when using mockery if you intend to make the mock return something different on the next call (e.g. the first part of the test returns an error for Read, and the second part returns an actual value).

minion/vault/vault.go
kklin: It actually is the kelda-internal name for the container. I added a comment to clarify this, and why assigning a name to the container is helpful.

minion/vault/vault.go
kklin: What should be the behavior once the timeout expires? Should we just panic/exit? That would seem aggressive to me..

minion/vault/vault_test.go
kklin: I had this comment a couple lines above, I think it's what you're requesting?
```
// Ensure that the referenced TLS credentials in the config were properly
// uploaded to the container.
```

minion/scheduler/worker.go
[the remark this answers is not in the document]
kklin: It's pretty unreadable IMO
```
CONTAINER       MACHINE         COMMAND                     HOSTNAME    STATUS                                                                                                      CREATED    PUBLIC IP
d044f3880fdc    sir-winicznj    alpine tail -f /dev/null    ignoreme    Waiting for secrets: [myEnvSecret myFileSecret]. Run `kelda secret <NAME> <VALUE>` to setup each secret.
```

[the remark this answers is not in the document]
kklin: It basically forces `quilt show`'s output to wordwrap, which is pretty hard to read

minion/vault/vault.go
kklin: This made me realize that if the container _doesn't_ get restarted, and Init was successfully called the first time, it'll error if it's called the second time. I changed the code to always start a new Vault container rather than reuse the existing one to avoid this case.

minion/scheduler/worker.go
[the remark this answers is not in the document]
kayousterhout: Ok I agree. I wonder if at some point later we could have a help message for STATUS (when you do quilt --help) that explains the action items for each status (in terms of things the user can do).  But anyway I agree with you that leaving just "Waiting for secrets: [x, y]" seems ideal for now

minion/run.go
kayousterhout: That would be ideal -- how hard is that to implement?  (I'd do it in a separate commit, but would be nice to get in soon, since it seems like something that will come up often when a user boots a blueprint and doesn't have any infrastructure running)

minion/vault/store_test.go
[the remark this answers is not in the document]
kayousterhout: I see I didn't realize that re: `Once()` (and changing the return value later) and now agree that it doesn't seem necessary in this case to have the check at the end.

minion/vault/vault.go
kayousterhout: I do think there should be a point when we panic/exit, because there's some amount of time after which we really don't expect this to succeed (like the 5 minute timeout on line 110 below), and in that case it's nice to tell the user pro-actively what went wrong.

minion/vault/vault_test.go
kayousterhout: I interpreted that comment as just referring to the next 5 lines (i.e., making sure the VAULT_LOCAL_CONFIG was set setup correctly, which I thought was being implicitly tested above, because if one of the configs isn't there, there will be a runtime error).  Was I misunderstanding that?  If my understanding is right, how about changing that comment to 

```
// Ensure that the referenced TLS credentials were properly added to the config and
// that the corresponding files were uploaded to the container.
```
Otherwise sounds find to leave the comment as-is.

minion/vault/vault.go
kklin: Maybe if we had a better debugging story, but in this case I think it'd be harder for the user to tell what happened since the error occurs on the remote minion. So in both cases, the user would have to run `kelda logs <MACHINE_THATS_NOT_CONNECTING>` in order to see the panic (or my current warning logs). And there still is some chance that it will eventually succeed (e.g. if the registry hosting Vault went down for 10 minutes).

Basically, I don't think panicking while convey more information to the user, and it removes the opportunity for us to maybe recover eventually.

integration-tester/tests/71-secret/secret_test.go
```diff
@@ -0,0 +1,77 @@
...
+	containers, err := clnt.QueryContainers()
+	if err != nil {
+		t.Fatalf("couldn't query containers: %s", err)
+	}
+
+	assert.Len(t, containers, 1)
+	containerID := containers[0].BlueprintID
+
+	// The secret values are defined in the 70-secret-setup test.
+	// The non-secret values are defined in the blueprint accompanying this test.
```
kayousterhout: These comments around this test are super helpful

integration-tester/tests/70-secret-setup/secretSetup.js
```diff
@@ -0,0 +1,8 @@
+const kelda = require('kelda');
+const infrastructure = require('../../config/infrastructure.js');
+
+const infra = infrastructure.createTestInfrastructure();
+
+(new kelda.Container('ignoreme', 'alpine', {
```
kayousterhout: Why do you need this container? Does the tester assume it needs at least one container to come up? (can you add a brief comment here? Or is this behavior the same in a bunch of other tests?)

minion/vault/vault.go
kayousterhout: Gah right I forgot that if a minion actually dies this error message won't get propagated back to the user in any useful way.  And I agree that there are some nice aspects of leaving it running -- like someone could login to try to debug it.  So I agree with you now :).

minion/run.go
kklin: The change is actually not that bad, so I'll just do it in the same commit

integration-tester/tests/70-secret-setup/secretSetup.js
kklin: It's a kind of half hearted test.. it has the same attributes as the container with secrets, so it might catch the case where the join is incorrect and doesn't properly kill this container. I could add a comment to explain this, or do you think I should just remove it?

kklin:
@kayousterhout all the tests you're describing would require non-trivial work. I agree with the sentiment that we should move towards better integration tests, but doing so would require a fair amount of work on the integration tester.

Do you think we could make this an Asana card, and debate tackling it in the next sprint?

integration-tester/tests/70-secret-setup/secretSetup.js
kayousterhout: I'd just remove it

minion/vault/api.go
```diff
@@ -0,0 +1,141 @@
...
+
+// InitStatus returns whether Vault has been initialized yet. See the documentation
+// on Init for more information on the initialization process.
+func (c vaultAPIClientImpl) InitStatus() (bool, error) {
+	vaultCounter.Inc("InitStatus")
+	return c.client.Sys().InitStatus()
+}
+
+// Unseal passes the Vault server the decryption key for the Vault secrets. In
+// contrast to Init, it must be called once per Vault process.
```
kayousterhout: The comments for these three methods are great now

minion/vault/api.go
```diff
@@ -0,0 +1,141 @@
...
+func (c vaultAPIClientImpl) Unseal(shard string) (*vaultAPI.SealStatusResponse, error) {
+	vaultCounter.Inc("Unseal")
+	return c.client.Sys().Unseal(shard)
+}
+
+func (c vaultAPIClientImpl) EnableAuth(path, authType, desc string) error {
```
kayousterhout: Can you add a comment for what this is too?

minion/vault/auth.go
```diff
@@ -0,0 +1,142 @@
...
+	// certMountName is the name to associate with the Vault certificate
+	// authentication backend. Clients interacting with the mounted backend
+	// will do so under /auth/cert.
```
kayousterhout: per our Slack convo, how about something like

```
Vault locations used for authentication.   An authentication backend runs in the Vault 
container, and stores authentication information in Vault using the path certMountName.  
certMountName is something we can choose; the other three paths are dictated by Vault (for 
example, if we set certMountName to `foo`, the login information would be stored at `foo/auth/login`).
```

minion/vault/auth.go
```diff
@@ -0,0 +1,142 @@
...
+	certMountName     = "cert"
+	certRootPath      = path.Join("/auth", certMountName)
+	certLoginEndpoint = path.Join(certRootPath, "login")
+	certListEndpoint  = path.Join(certRootPath, "certs")
+)
+
+// certRole represents the access configuration for an entity with the private
+// key associated with the certificate.
```
kayousterhout: can you change "the certificate" to "`cert`"? (took me a minute to figure out what "the certificate" was referring to)

minion/vault/auth_test.go
kayousterhout: Can you add an assertExpectations, to make sure that Read() actually got called (to guard against the case where there was an error in List)

minion/vault/policy.go
```diff
@@ -0,0 +1,171 @@
...
+	// {
+	//   "path": {
+	//     "/canRead": {
+	//       "policy":"read"
+	//     },
+	//     "/canWrite": {
+	//       "policy":"write"
+	//     }
+	//   }
+	// }
```
kayousterhout: This example is super useful

minion/vault/vault.go
kayousterhout: Took me a minute to figure out why you thumbs-upped this :).  I realize your comment above implicitly gets at this -- but can you add a short thing here like
```
SecretShares:    1, // Number of unseal keys to return
SecretThreshold: 1, // Number of unseal keys that must be provided to unseal vault
```

minion/vault/vault.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/kelda/kelda/db"
+	"github.com/kelda/kelda/minion/docker"
+	"github.com/kelda/kelda/util"
+
+	vaultAPI "github.com/hashicorp/vault/api"
+	log "github.com/sirupsen/logrus"
+)
+
+/**
+* This submodule configures the Vault container on the masters. It ensures
```
kayousterhout: Per-offline convo, add that: vault containers are started on each master, but only the vault container on the leader is used, and if the leader fails, vault will re-start, but all of the secrets will be lost and will need to be re-added by the user

kayousterhout: Also why bother starting one on each master if only the leader one is used?  Is the idea that the non-leader ones are just there to takeover in case the leader fails? (also would be great to have that in the comment)

minion/vault/auth.go
kklin: FYI I tweaked the comment a bit because `/auth` is always the prefix -- the login information is at `/auth/foo/login`, not `/foo/auth/login`.

PR #1191, opened 2017-10-19
ci: Wait for the credential files rather than just the directory

kklin:
Commit db2c21b2 (ci: Wait for the daemon to generate TLS credentials)
attempted to avoid a race between the daemon generating the credentials,
and the tester running commands that required those credentials.
However, the presence of simply the TLS directory is not sufficient for
the commands to work -- the actual credential files are needed. The
commands were still failing because the daemon creates the TLS directory
before actually writing the credentials, so the tester would start
running kelda commands while the daemon was writing the credentials.


integration-tester/tester.go
```diff
@@ -193,8 +194,12 @@ func (t *tester) setup() error {
 	// the subsequent Kelda commands (such as `kelda stop`) might fail if they
 	// are executed before the credentials are generated.
 	err = util.BackoffWaitFor(func() bool {
```
ejj: Is it `runKeldaDaemon` that generates the dependencies?  I wonder if it would be cleaner if it just blocked until these pre-requisites are done.  I.e. instead of doing `go runKeldaDaemon()` you would do `runKeldaDaemon()` it would set up dependencies, then spawn a go routine, and return.  Just a thought.

kklin: The dependencies are generated by shelling out to `kelda daemon`, so it's not like `runKeldaDaemon` generates the dependencies, then shells out to the daemon. I think the way it is now is fine.

kklin: Is this fine for me to merge @ejj?

PR #1193, opened 2017-10-19
ci: Use the local version of the bindings

kklin:
Before, the tester used the latest version of `kelda` hosted on NPM;
however, some tests might fail if they depend on changes to the bindings
that have not yet been released on NPM.


kklin:
@kayousterhout can you review this as well? it's needed for the secrets integration test to work

PR #1194, opened 2017-10-20
How To for workflow stories

luise:
More are coming, but these are ready for review.


docs/_HowTo.md
```diff
@@ -47,3 +47,90 @@ same `name:tag` as the currently running one.
...
+### Multiple Replicated Applications
+You might want to run _multiple_ replicated applications -- e.g. two different
+replicated websites. Calling `simpleLoadBalancer` for each of them would result
+in a separate load balancer for each application. That's not what we want. Let's
+instead see how to put two different applications behind a single load
+balancer. The approach is very similar to the one above. The only significant
+difference in the blueprint is that we create the load balancer using
+`withURLrouting()` instead of `simpleLoadBalancer()`. We additionally need each
+application to be associated with a domain name (registered with e.g.
+[GoDaddy](https://www.godaddy.com/) or [Namecheap](https://www.namecheap.com/)).
```
luise: When we have a "How To" for the DNS name workflow (coming soon), we will link to that section instead of Namecheap and GoDaddy.

docs/_HowTo.md
```diff
@@ -0,0 +1,49 @@
+# How To
+
+## How to Update an Application Running on Kelda
```
ejj: I think this title is a little bit hard to interpret.  But I'm not quite sure what would be better.  Basically, I think people may have trouble realizing that we're talking about the code that they've written.

docs/_HowTo.md
```diff
@@ -0,0 +1,49 @@
...
+For users who build and push their own images, we recommend always tagging
```
ejj: I think a lot of people won't necessarily know what an "image" is.  At least if we're assuming they don't need to know about docker to get going.  I wonder if there's some external docker documentation we can link to on tagging images? At the very least, we should use the word Docker a lot more frequently in this paragraph, so it's clear what we're talking about.

docs/_HowTo.md
```diff
@@ -0,0 +1,49 @@
...
+images, and not using the `:latest` tag.
+This way, updating from `me/myWebsite:0.1` to `me/myWebsite:0.2` is simply a
```
ejj: I think I might rewrite this sentence to be something a bit more like this.

If you're using tags, then updating to the latest version is simple.  For example, updating migrating from 
`me/myWebsite:0.1` to its latest version, `me/myWebsite:0.2`, is a simple two step process.  
* Update each reference in the blueprint to the new tag, `me/myWebsite:0.2`.
* Execute `kelda run` again to updating the running cluster.

That's very rough of course, but I think at a high level this sentence is shoving too much information into too short a space.

docs/_HowTo.md
```diff
@@ -0,0 +1,49 @@
...
+matter of updating the blueprint to use the new tag and then `kelda run`
+the updated blueprint.
+
+### Untagged Images and Images Specified in Blueprints
+For users who do not use tagged images, there are currently two ways of
```
ejj: tagged docker images

docs/_HowTo.md
ejj: Also, I think this comment here: "For users who build and push their own images, we recommend always tagging" implies that we only recommend doing so for users who want to build their own docker images.  I think it would be better to rephrase it to call out that the most robust way to handle application updates is to build and tag your own docker image rather on relying on the blueprint to do the build for you.

docs/_HowTo.md
```diff
@@ -0,0 +1,49 @@
...
+* You updated the contents of a file that is transferred to the application
+container using `withFiles`.
+* You changed the Dockerfile string passed to the `Image` constructor in the
+blueprint.
+
+
+#### Updating an Image
+Though we recommend using _tagged_ images, some
+applications might use untagged images hosted in a registry like Docker Hub or
+`Images` specified in blueprints. To pull a newer version of a hosted image or
```
ejj: ````Images` specified in blueprints.``` I don't even really know what this is.  May want to rephrase it or explain it more.

docs/_HowTo.md
```diff
@@ -0,0 +1,49 @@
...
+rebuild an `Image` with Kelda, you need to restart the relevant container:
+
+1. In the blueprint, remove the code that `.deploy()`s the container that
```
ejj: Eww. I'll be glad when we have update code  =)

docs/_HowTo.md
```diff
@@ -47,3 +47,90 @@ same `name:tag` as the currently running one.
 didn't change. This could for instance be because the Dockerfile clones a
 GitHub repository, so rebuilding the image would recreate the container with
 the latest code from GitHub.
+
+## How to Run a Replicated, Load Balanced Application Behind a Single IP Address
```
ejj: Should there be a comma after Load Balanced?

docs/_HowTo.md
```diff
@@ -47,3 +47,90 @@ same `name:tag` as the currently running one.
...
+balanced application behind a single IP address. We will use [HAProxy](https://www.haproxy.com/)
+(High Availability Proxy), a popular open source load balancer, to evenly
+distribute traffic between your application containers.
+
+Before we start writing the blueprint, make sure the application you want to
+replicate is listening on port 80. E.g., for an Node.js Express application
+called `app`, call `app.listen(80)` in your application code.
+
+### A Single Replicated Application
+In the blueprint, import [Kelda's HAProxy blueprint](https://github.com/kelda/haproxy):
```
ejj: "the blueprint" is vague.  Maybe, "in your application blueprint" or something

docs/_HowTo.md
```diff
@@ -47,3 +47,90 @@ same `name:tag` as the currently running one.
...
+This guide describes how to write a blueprint that will run a replicated, load
+balanced application behind a single IP address. We will use [HAProxy](https://www.haproxy.com/)
+(High Availability Proxy), a popular open source load balancer, to evenly
+distribute traffic between your application containers.
+
+Before we start writing the blueprint, make sure the application you want to
+replicate is listening on port 80. E.g., for an Node.js Express application
+called `app`, call `app.listen(80)` in your application code.
+
+### A Single Replicated Application
```
ejj: I dunno if this is the right PR for it.  But i think it would be cool for the HOW Tos to have example code associated with it that we could link to in the docs.  I'm imagining a "replicated-load-balancer.js" or something, which is the bare minimum to get that working that people could work off of.

docs/_HowTo.md
```diff
@@ -47,3 +47,90 @@ same `name:tag` as the currently running one.
...
+
+```javascript
+loadBalancer.allowFrom(publicInternet, haproxy.exposedPort);
+```
+
+Deploy the application containers and load balancer to your infrastructure
+(here, the infrastructure is called `inf`):
+
+```javascript
+appContainers.forEach(container => container.deploy(inf));
```
ejj: I wonder if we should just use baseInfrastructure() instead.   That way we don't have to explain what inf is, and it's easier to run if you copy and paste.

docs/_HowTo.md
```diff
@@ -47,3 +47,90 @@ same `name:tag` as the currently running one.
...
+loadBalancer.deploy(inf);
+```
+
+The application will be accessible on the `PUBLIC IP` of the `haproxy`
+container.
+
+### Multiple Replicated Applications
+You might want to run _multiple_ replicated applications -- e.g. two different
+replicated websites. Calling `simpleLoadBalancer` for each of them would result
```
ejj: Maybe this would be clearer: "e.g. two different websites with two different domain names".

docs/_HowTo.md
```diff
@@ -47,3 +47,90 @@ same `name:tag` as the currently running one.
...
+in a separate load balancer for each application. That's not what we want. Let's
```
ejj: I think you could slightly reorganize this paragraph in a way that flows better, something like this:

* What if you have two websites with two different domain names.
* You could get this working by calling simple load balancer twice.  That may work for you, but it would create two different HAProxy instances with two different IPs.
* In this section, we'll talk about how to have two websites, behind a single HAproxy (and therefore a single IP address).

Obviously not that text precisely, but I think if you introduce the problem before introduce the solution, this is going to read a bit more easily.

ejj: Also, this paragraph is too long I think.  Given that most people skim docs, I think we want a writing style that's a lot more bullet points and short paragraphs, and a lot less pros.

Also I think it would be really good in general with this PR, to rely a bit more heavily on bolding important words and phrases.  Again, since people skim, that will help them figure out what they want quickly.

docs/_HowTo.md
ejj: I'd ditch the registrar bit -- it kind of comes out of nowhere and is a bit distracting.  You could give an example of what a domain name is instead though.  I.e. "each application needs it's own domain name, like apples.io and oranges.io) or something.

docs/_HowTo.md
ejj: Yeah looking at this now, I think that would be really hugely helpful.  It's kind of hard to get a feel for what you're talking about without seeing it all together.  Particularly for the replicated load balancer bit.

docs/_HowTo.md
luise: Yeah, that makes sense. How about "How to Run the Latest Version of Your Application" ?

docs/_HowTo.md
luise: Good point

docs/_HowTo.md
luise: Yes! This is so ugly!

docs/_HowTo.md
ejj: I like that a lot better.

docs/_HowTo.md
luise: Nope :)

docs/_HowTo.md
luise: Ah, good idea, I'll use `baseInfrastructure()`!

docs/_HowTo.md
luise: Yeah, that's a good idea. I will submit PRs to the HAProxy repo for this one (might not be until the sprint deadline)

docs/_HowTo.md
luise: Hm, OK, I tried to update it. There is a lot of information, so it's not that easy to cut down.

luise:
@ejj , I addressed your comments. Also added another commit that explains the daemon (will be useful for the workflow docs on the remote daemon).

luise:
@ejj , added another How To -- this one is for running the daemon remotely

docs/_HowTo.md
```diff
@@ -0,0 +1,220 @@
...
+updated image.
+
+Examples of when you need to stop and start the container in order to update the
+application:
+
+* You use a hosted image (e.g. on Docker Hub), and pushed a new image with the
+same `name:tag` as the currently running one.
+* You want to rebuild an `Image` even though the Dockerfile in the blueprint
+didn't change. This could for instance be because the Dockerfile clones a
+GitHub repository, so rebuilding the image would recreate the container with
```
aegamesi: "this could for instance be because" sounds a little bit awkward. If you slightly rearranged the words, like "For instance, this could be because...", or maybe added commas: "This could, for instance, be because" (although I still think that sounds a bit awkward).

docs/_HowTo.md
```diff
@@ -0,0 +1,220 @@
...
+* **Robustness**. If any machine dies while your laptop is offline, the machine
+  won't be rebooted until your daemon is back up.
+* **Collaboration**. Currently, there are two issues with using a local daemon
+  if you share the deployment with others.
+  * **Multiple Daemons**. Each daemon assumes it's the only daemon attached to
+  its deployment, so if your teammates are also running local daemons for your
+  shared deployment, the daemons might misbehave.
+  * **TLS Keys**: Each daemon connecting to the deployment must have the same
+  TLS credentials, so you'd have to share your `~/.kelda/tls` directories.
+
```
aegamesi: The indentation of the list items here is inconsistent.

docs/_HowTo.md
```diff
@@ -0,0 +1,220 @@
...
+#### Setting up the Remote Daemon
+1. **Create a VM** (Ubuntu or Debian) on your preferred cloud provider. You can
+  choose a small instance type to keep costs low.
+  - If your provider blocks ports by default, allow ingress TCP traffic on ports 22 and 3333.
+2. **TLS**. Copy `~/.kelda/tls` from your local machine to `~/.kelda/tls` on the VM.
+3. **SSH** in to the VM and do steps a-c from the VM.
+  - [Install Node.js](https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions).
+  - [Install Kelda](#installing-kelda) with `npm`.
+  - **Provider Credentials**. Set up your [provider credentials](http://docs.kelda.io/#cloud-provider-configuration).
+  -  **Start the Daemon**. The following command starts the daemon in the
```
aegamesi: Super nitpicky but this line is also slightly misaligned

aegamesi:
This is the guide I wish had been available when I first started working with Kelda! I think it would have cleared up a good amount of my confusion.

docs/_HowTo.md
luise: This is actually intentional -- they're sub bullets. I'll change them to `-` instead though, so it's clearer!

luise:
Thanks for your comments, @aegamesi ! I updated the PR.
And I'm glad you think this will be useful :)

docs/_HowTo.md
ejj: Cool, just make sure you link to it here when you're ready.

PR #1196, opened 2017-10-21
js: Checks that non-preemptible machines are not set as spot machines

HongjianZhang:
Before this commits, the chooseSize function does not distinguish
between spot and reserved machines. It simply assumes that all
machines can be preempted. This is not always the case, e.g. Amazon
t2.micro can only work when using reserved machines. This commit
makes sure that user is either using a spot machine or preemptable
is set to False in blueprint, and warns user to change the config
otherwise.


js/bindings/bindings.js
```diff
@@ -795,11 +796,18 @@ class Machine {
    * Sets the machine's size attribute to an instance size (e.g., m2.xlarge),
    * based on the Machine's specified provider, region, and hardware. Throws
    * an error if provider is not valid or given machine requirements cannot
-   * be satisfied by any size.
+   * be satisfied by any size. Also throws an error if user requests a spot
```
kklin: I think "spot" should be removed here as well. Something like "Also throws an error if the user requests a preemptible instance for a size that cannot be preempted."

js/bindings/bindings.js
```diff
@@ -795,11 +796,18 @@ class Machine {
...
+   * machine for a size that cannot be preempted.
    * @private
    * @returns {void}
    */
   chooseSize() {
+    if (this.preemptible && this.size !== ''
+        && nonPreemptibleSizes[this.provider].includes(this.size)) {
+      throw new Error(`Requested size ${this.size} can not be preemptible.` +
+        ' Please choose a different size or set preemptible to be False in' +
```
kklin: I don't think "in blueprint" is necessary. This error message will get thrown when running the blueprint, so it's natural for the user to make the adjustment in the blueprint.

js/bindings/bindingsTest.js
```diff
@@ -240,6 +240,18 @@ describe('Bindings', () => {
         region: 'eu-west-1',
       }]);
     });
+    it('errors if preemptible set to true and requested '
+      + 'size is spot instance for the given provider', () => {
```
kklin: Same here re "spot". Also, I think you meant to write "is not a spot instance"?

js/bindings/nonPreemptibleSizes.json
```diff
@@ -0,0 +1,5 @@
+{
+  "Amazon": [
+  	"t2.micro"
+  ]
+}
```
kklin: This file seems to be missing a newline at the end of the file

js/bindings/package.json
```diff
@@ -7,6 +7,7 @@
     "bindings.js",
     "digitalOceanDescriptions.json",
     "googleDescriptions.json"
```
kklin: Do you need a comma here? Can you try `npm install` on this package and make sure it doesn't error?

HongjianZhang:
Fixed.

PR #1198, opened 2017-10-21
travis: Only deploy once per commit

kklin:
Before, we deployed twice for every commit because Travis starts a new
build for each NODEJS_VERSION in the build matrix.

Note -- this is untested.


kayousterhout:
Do you think it makes sense to set the first NODEJSVERSION=DEPLOYNODEJSVERSION to ensure consistency? I see pros and cons of this so fine either way

kklin:
I thought about it, but i liked that the build matrix is easier to read this way.

PR #1199, opened 2017-10-22
travis: Fix NODEJS_VERSION deployment check

kklin:
The check didn't actually evaluate the NODEJS_VERSION variable in the
conditional. In fact, it assigned to the variable.


.travis.yml
```diff
@@ -89,17 +89,17 @@ deploy:
...
-        condition: NODEJS_VERSION=$DEPLOY_NODEJS_VERSION
+        condition: $NODEJS_VERSION=$DEPLOY_NODEJS_VERSION
         tags: true
     - provider: script
       script: npm publish js/install
       skip_cleanup: true
       on:
         tags: true
-        condition: $TRAVIS_TAG != "dev" && NODEJS_VERSION=$DEPLOY_NODEJS_VERSION
+        condition: $TRAVIS_TAG != "dev" && $NODEJS_VERSION=$DEPLOY_NODEJS_VERSION
```
kayousterhout: Super nit: use spaces around the “=“ for consistency with the other conditional?

kklin: Done!

kayousterhout:
Is == the right syntax here? That’s what I expected when I first saw this change so I looked atthe Travis docs butthe examples there seem to use single =

kklin:
So `=` is the operator defined by POSIX. `==` is an extension supported by bash. https://unix.stackexchange.com/questions/168284/unix-test-when-to-use-eq-vs-vs-in-test-commands

Travis evaluates its conditionals using the bash `test` command:
```
condition: You may set bash condition with this option. This must be a string value, which will be pasted into a bash expression of the form if [[ <condition> ]]; then <deploy>; fi It can be complex, but there can be only one. For example, $CC = gcc.
```

That's why it supports the non-POSIX compliant `==`. I can see an argument for switching all of the Travis config to use `=`. What do you think? I'm fine either way.

kayousterhout:
Ehh I didn't realize there were other spots where we used this.  If all of our existing code uses ==, let's just leave it.

PR #1203, opened 2017-10-22
cloud: Fix SSH key syncing

kklin:
Before, SSH keys were only written to the database when the machines
were booting. This was obviously insufficient when the blueprint changed
the SSH keys, or the daemon restarted when machines were already running
the cloud, and it thus did not have to boot new machines. In the latter
case, the SSH keys would never get written to the database, and the
foreman would thus revoke all SSH access.


kklin:
Yeah, SSH keys and floating IPs.

PR #1205, opened 2017-10-23
scheduler: Log secret fetch error at INFO

kklin:
This commit demotes the error logged when fetching a secret from WARN to
INFO because the error is expected when the leader has scheduled a
container, but not yet configured the Vault access policy.


minion/scheduler/secret.go
```diff
@@ -47,10 +47,18 @@ func resolveSecrets(client vault.SecretStore, dbcs []db.Container) map[string]st
...
-				log.WithError(err).WithField("name", name).
-					Warn("Failed to fetch secret")
+			if err == vault.ErrSecretDoesNotExist {
+				continue
 			}
+
+			log.WithFields(log.Fields{
+				"error": err,
+				"name":  name,
+			}).Info("Failed to fetch secret. This error is probably safe " +
```
kayousterhout: I'd say this error is safe to ignore when the container has just been booted

kklin: What do you mean?

kayousterhout: I was thinking something like:

"Failed to fetch secret.  This error is benign if the container was launched recently -- permissions issues when containers are first scheduled are expected because the leader is still configuring the Vault ACLs for accessing the secret."

I think the "IF the container was launched recently" is important to state asap so users don't ignore this when it's actually important

kklin: Ah makes sense sorry. Updated!

PR #1207, opened 2017-10-24
Additional "How To" docs

luise:


luise:
@ejj , here are the remaining workflow stories. Would be great if you can continue the review (this is the second part of the PR I split up earlier)

docs/_HowKeldaWorks.md
```diff
@@ -1,24 +1,40 @@
...
+The key idea behind Kelda is a blueprint -- in this case `my_app.js`. A
+blueprint describes every aspect of
 running a particular application in the cloud, and is written in JavaScript.
 Kelda blueprints exist for many common applications.  Using Kelda, you can run
 one of those applications by executing just two commands on your laptop:
+`kelda daemon` and `kelda run`.
 
-![Kelda Diagram](Kelda_Diagram.png)
+### Daemon
+The first command,`kelda daemon`, starts a long-running process -- "the daemon".
```
aegamesi: should have a space after the first comma

luise: Nice catch!

docs/_HowKeldaWorks.md
```diff
@@ -1,24 +1,40 @@
...
 Kelda blueprints exist for many common applications.  Using Kelda, you can run
 one of those applications by executing just two commands on your laptop:
+`kelda daemon` and `kelda run`.
 
-![Kelda Diagram](Kelda_Diagram.png)
+### Daemon
+The first command, `kelda daemon`, starts a long-running process -- "the
+daemon". The daemon is responsible for booting machines in the cloud,
+configuring the machines, managing them, and rebooting them if anything goes
+wrong. This means that if a machine dies while no daemon is connected to it
```
ejj: "is connected to it" is ever so slightly misleading.  I would just say "is running".  The daemon doesn't need to be connected to reboot a failed machine.

docs/_HowKeldaWorks.md
```diff
@@ -1,24 +1,40 @@
...
+(e.g. because you closed the laptop that's running the daemon), the machine will
+not be restarted. However, as soon as the daemon is reconnected to the
```
ejj: "As soon as the daemon comes back online"

docs/_HowKeldaWorks.md
```diff
@@ -1,24 +1,40 @@
...
-those machines, and managing them (e.g., detecting when they've failed so need
-to be re-started).  The `kelda daemon` command starts the daemon, but doesn't
-yet launch any machines. To launch an application, call `kelda run` with a
-JavaScript blueprint (in this example, the blueprint is called `my_app.js`).
-The `run` command passes the parsed blueprint to the daemon, and the daemon
-sets up the infrastructure described in the blueprint.
+### Run
+To launch an application, call `kelda run` with a JavaScript blueprint -- in
+this example `my_app.js`. The `run` command passes the parsed blueprint to the
+daemon, and the daemon sets up the infrastructure described in the blueprint.
```
ejj: So the one thing that's missing here.  (Could be in a different PR, if you think that's best please write a card).  We aren't really capturing the subtlety around `kelda run`ing a spec that's already up and running.  In that case, the daemon will just notice that machines exist rather than booting it.

There's some broader UX questions here (the current flow is kinda icky).   Again, this PR shouldn't fix everything w.r.t. this problem. Whatever you think.

docs/_HowKeldaWorks.md
```diff
@@ -1,24 +1,40 @@
...
 
-Kelda runs applications using Docker containers. You can think of a container
+### Containers
+Kelda runs applications using [Docker](http://docker.com/) containers. As shown
```
ejj: "using Docker"  I don't think you need the containers bit.

docs/_HowKeldaWorks.md
```diff
@@ -27,9 +43,18 @@ isolated from other containers running on the same machine (unlike processes).
...
+**Workers**: The worker machines host the application containers. In this case,
+Kelda ran two containers on one worker machine and one container on the other.
+
+**Masters**: The master is responsible for managing the worker machines and the
+application containers running on the workers. No application containers run on
+the master. In the above explanation of [the daemon](#daemon), we described a
+scenario where a worker machine disappears. In that scenario, the master machine
+would reboot the application containers from the failed worker on one of the
+healthy worker machines. It is possible to have multiple master machines to
+safeguard against a master machine failing. If there are multiple master
```
ejj: I would ditch the "if there are multiple masters" sentence.  That's an implementation details that I don't think is relevant.

docs/_HowTo.md
```diff
@@ -148,3 +148,85 @@ and check you get the right response:
 ```console
 $ curl -H "Host: apples.com" HAPROXY_PUBLIC_IP
 ```
+
+
+## How to Run the Daemon in Production
```
ejj: Can we think of a less scary phrase than "in Production"?

docs/_HowTo.md
```diff
@@ -148,3 +148,85 @@ and check you get the right response:
...
+_We recommend reading about [the daemon](#daemon) before reading this section._
+
+The default way to run the daemon is to run it on a local machine like your
+laptop. While this is easy, and might work for you, there are two drawbacks to
```
ejj: "a couple of drawbacks"

docs/_HowTo.md
```diff
@@ -148,3 +148,85 @@ and check you get the right response:
...
+keep in mind:
+
+* **Robustness**. If any machine dies while your laptop is offline, the machine
```
ejj: The word "machine" is kind of kelda internals slang.  I think it might be clearer if we say "Cloud VM" the first use, and then "VM" in future users.  Perhaps you can think of something even better?

ejj: "your laptop" -> "the daemon"

"your daemon" -> "the daemon"

docs/_HowTo.md
```diff
@@ -148,3 +148,85 @@ and check you get the right response:
...
+  won't be rebooted until your daemon is back up.
+* **Collaboration**. Currently, there are two issues with using a local daemon
+  if you share the deployment with others.
+  - **Multiple Daemons**. Each daemon assumes it's the only daemon attached to
+  its deployment, so if your teammates are also running local daemons for your
```
ejj: I don't like "your teammates"  assumes too much about the reader.  maybe something along the lines of: if you're sharing your deployment with others that also have daemons or something.

docs/_HowTo.md
ejj: Btw do we really need this drawback section?  Can we just say "If you're running a long term deployment, we don't recommend runng the daemon on your laptop. " and just leave it at that?  

My concern is this subsection is kind of technical and scary sounding, and the drawbacks of running the daemon on your laptop are somewhat obvious (i.e. it's not running all the time).

docs/_HowTo.md
```diff
@@ -148,3 +148,85 @@ and check you get the right response:
...
+  shared deployment, the daemons might misbehave.
+  - **TLS Keys**: Each daemon connecting to the deployment must have the same
+  TLS credentials, so you'd have to share your `~/.kelda/tls` directories.
+
+
+### Solution: A Shared Daemon on a Separate VM
+To solve the above problems, we recommend running a single, shared daemon on a
+small VM on your preferred cloud provider. Even though the daemon is remote, you
```
ejj: "your preferred cloud provider" -> "in the cloud"

ejj: And perhaps ditch the "even though" sentence.  It doesn't add much

docs/_HowTo.md
```diff
@@ -148,3 +148,85 @@ and check you get the right response:
...
+can still run all `kelda` commands from your local machine.
+
+#### Setting up the Remote Daemon
```
ejj: Sigh ... this is somewhat painful.

We have a card on improving this workflow right?

docs/_HowTo.md
```diff
@@ -148,3 +148,85 @@ and check you get the right response:
...
+  background, and redirects its logs to the `daemon.log` file. The `nohup`
+  command ensures that the daemon keeps running even when you log out of the VM.
+  Here, the daemon will be listening for connections on port `3333`:
+
+    ```console
+    $ nohup kelda daemon -H tcp://<PRIVATE_IP_OF_DAEMON_VM>:3333 > daemon.log 2>&1 &
+    ```
+  **Note!** Some providers (e.g. DigitalOcean) don't give you a private IP.
+  If that's the case, try the public IP of the VM instead. If that doesn't work
+  either, try `0.0.0.0`.
```
ejj: Why not just always recommend 0.0.0.0?

docs/_HowTo.md
```diff
@@ -148,3 +148,85 @@ and check you get the right response:
...
+
+4. **Allow SSH**. To allow you and your teammates to SSH in to the deployment,
```
ejj: I wonder if this is complicated enough, that we should just tell people to ssh into the VM and do things from there.  What do you think?

docs/_HowTo.md
```diff
@@ -1,5 +1,49 @@
...
+2. Tell Kelda to place the application on the machine with your floating IP:
+
+    ```javascript
+    const app = new Container('myApp', 'myImage');
+    app.placeOn({ floatingIP });
+
+    // Deploy the application.
+    app.deploy(inf);
+    ```
+If your website is replicated, follow the guide for running a [replicated,
```
ejj: "replicated" is jargony.  Can we simplify the language?

ejj:
LGTM, just some minor suggestions.  Ping me when you're ready.

docs/_HowKeldaWorks.md
luise: Hm, yeah, I see how that's not completely accurate. I said "connected" because I wanted to signal that the daemon has to be registered with the namespace -- you can't just run `kelda daemon`.
How about "while no daemon is running for its deployment"? Or "while no daemon is tracking it"?

docs/_HowKeldaWorks.md
luise: Hm, yeah, I'll make a card for it. I'd be happy to add it in this PR, but I think it would require mentioning namespaces also which aren't documented anywhere.  So I think we should just add it all in a separate PR.

docs/_HowKeldaWorks.md
luise: I kind of like saying containers here because it then is clearly connected to the containers in the next sentence. Alternatively, I'd delete "containers" in the first sentence, and then say "with three Docker containers" in the following sentence. I'll do that for now.

docs/_HowTo.md
luise: Good point. Maybe "hosted on multiple servers"?

docs/_HowTo.md
luise: Hmm, I might just say "How to run the daemon".

docs/_HowTo.md
luise: Hm, yeah. I don't think the "collaboration" section is obvious though. We don't mention the part about multiple daemons anywhere, and the part about TLS keys also isn't mentioned directly anywhere. Let me try to rephrase and shorten it (and leave out the robustness point).

docs/_HowTo.md
luise: I actually think it makes a difference that I can run `kelda show` from my local laptop and don't have to ssh into the machines with the daemon, so I think I'll leave the  sentence.

docs/_HowTo.md
luise: Oh, I'd much rather do that. Didn't know if that was frowned upon.

docs/_HowTo.md
luise: Yup!

docs/_HowTo.md
luise: Hmm, yeah, I could see that too. Basically, the story would just be "create a VM, SSH in, install everything, and get the blueprints to this VM, and run all kelda commands from the VM".  It's not pretty, but it might be more straight forward than this.
I would actually be fine with doing that for now.

luise:
@ejj , addressed your comments. Changed the "How to run the daemon" workflow to simply be 1) create a VM in the cloud, 2) install Kelda etc. and 3) run all Kelda CLI commands from this machine.

ejj:
Sounds good.  Rebase and I"ll merge it.

luise:
@ejj  , rebased

PR #1208, opened 2017-10-24
spark: Don't rely on job() method to run SparkPi

ejj:
In future patches, we'll be removing the job() method from the Spark
Blueprint.  Also, executing the job directly is just a lot cleaner.


integration-tester/tests/20-spark/spark_test.go
```diff
@@ -38,21 +36,14 @@ func TestCalculatesPI(t *testing.T) {
...
-		fmt.Printf("`kelda logs %s` output:\n", id)
-		fmt.Println(string(logs))
-		return strings.Contains(string(logs), "Pi is roughly")
-	}, 15*time.Second, time.Minute)
-
+	cmd := exec.Command("kelda", "ssh", id, "run-example", "SparkPi")
+	output, err := cmd.Output()
 	if err != nil {
-		t.Fatalf("unable to get Spark master logs: %s", err.Error())
+		t.Fatalf("Failed to run spark job")
```
kayousterhout: Can you add the error message w/ the failure reason here? (since these are often hard to debug -- since it can be hard to recreate the jenkins failure)

kklin: In a similar vein, it'd be helpful if the command output were printed before this error case, since there might be some helpful information in stderr

ejj:
Updated

integration-tester/tests/20-spark/spark_test.go
```diff
@@ -38,21 +36,14 @@ func TestCalculatesPI(t *testing.T) {
...
-		}
-
-		fmt.Printf("`kelda logs %s` output:\n", id)
-		fmt.Println(string(logs))
-		return strings.Contains(string(logs), "Pi is roughly")
-	}, 15*time.Second, time.Minute)
-
+	cmd := exec.Command("kelda", "ssh", id, "run-example", "SparkPi")
+	output, err := cmd.Output()
+	fmt.Println(string(output))
```
kayousterhout: @kklin did you mean for this to be in the if statement? (Will this clutter the Jenkins output here?)

kklin: No, I think we should just always print it. Never hurts to have too many logs, and each test has its own output file, so this won't affect any other tests

PR #1213, opened 2017-10-26
bindings: Remove withFiles

kayousterhout:
This function is confusing because it's not obvious that it returns
a new container rather than modifying an existing container. Because
of this functionality, it's just as easy to set 'filepathToContent'
when constructing the container, and less error-prone.

This has bitten me a few times and has been annoying to debug, but I'm
curious how other folks feel about this.
I checked and none of our other blueprints use this method.  This is also
related to feedback we got from existing users, that it's confusing that
there are multiple ways of doing this (and non-obvious which is the
correct one).  We may also want to remove withEnv for the same reason.


kklin:
I'm fine with removing the `with` functions. I added `withFiles` because you couldn't specify the files through the constructor. Now that we can do that, I don't think it's all that useful.

ejj:
I'm generally in favor of having one way to do things FWIW.

You do have a mild theoretical issue where it's impossible to populate a file with information retrieved from the container object if the only way to set it is through the constructor ....  Dunno if that matters though.

kklin:
You could always assign to the object (and use `Container.clone` if you really have to).

integration-tester/tests/load-balancer/loadBalancer.js
```diff
@@ -5,10 +5,11 @@ const infra = infrastructure.createTestInfrastructure();
 
 const containers = [];
 for (let i = 0; i < 4; i += 1) {
-  containers.push(new kelda.Container('web', 'nginx:1.10').withFiles({
-    '/usr/share/nginx/html/index.html':
-        `I am container number ${i.toString()}\n`,
-  }));
+  containers.push(new kelda.Container('web', 'nginx:1.10', {
+    filepathToContent: {
```
ejj: This isn't the right PR for it, but I really hate the name "filepathToContent".  It's very scary.   I wonder if we should rename it "withFiles" perhaps.  I dunno, not a priority, just wanted to throw that out into the universe.

kayousterhout: There's some context here in the original PR: https://github.com/kelda/kelda/pull/793 

Originally Kevin had called the variable FileMap but we were concerned it was too non-obvious what the variable referred to in that case (what's the key and what's the value? is either of them a filepath on the local filesystem?  is either a file handle on the local filesystem)

Anyway agree this should be a separate discussion from this PR!

js/bindings/bindingsTest.js
```diff
@@ -498,23 +498,23 @@ describe('Bindings', () => {
 
     const filepathToContent = { qux: new b.Secret('quuz') };
     const jsonFilepathToContent = { qux: { nameOfSecret: 'quuz' } };
-    it('with*', () => {
+    it('withEnv', () => {
```
kayousterhout: FYI I updated this unit test since submitting the original PR, because the test failure made me realize this name was out of date now

kayousterhout:
@ejj I updated the unit tests so this is ready for another look / merge

PR #1215, opened 2017-10-26
make: Clean up coverage files

luise:


Makefile
```diff
@@ -41,8 +41,7 @@ check: gocheck jscheck
 
 clean:
 	govendor clean -x +local
-	rm -f *.cov.coverprofile cloud/*.cov.coverprofile minion/*.cov.coverprofile
-	rm -f *.cov.html cloud/*.cov.html minion/*.cov.html
+	find . -name "*.cov*" -type f -delete
```
kklin: The second `*` is a bit too greedy, it'd be safer to be explicit about the extensions. E.g. "./node_modules/cliui/.coveralls.yml" was matched when I ran that find command.

Btw `find . -name '*.cov' -or -name '*.cov.html'` is how I would have multiple matches.

luise: Oh, nice! I'll update!

kklin: Oh, I tried it locally, and I think you need to group the match for the `-delete` to work?

```
find . -name '*.cov' -or -name '*.cov.html' -type f
```

luise: Oh, I see. I just tested it without the `-delete`, and that found all the files. Thanks! Updated!

PR #1217, opened 2017-10-26
scheduler: Wait for Vault before attempting to boot containers

kklin:
Before, the workers would sometimes log scary errors about being unable to
connect to Vault. This patch makes it so the worker first ensures it can
successfully connect to Vault before attempting to boot containers. This
way, if we fail to connect to Vault later on, we know it's a real issue,
and deserves to be logged at ERROR.


minion/scheduler/scheduler.go
```diff
@@ -48,12 +48,12 @@ func bootWait(conn db.Conn) {
...
 	}
 
-	for workerCount := 0; workerCount <= 0; {
-		workerCount = 0
-		for _, m := range conn.SelectFromMinion(nil) {
-			if m.Role == db.Worker {
-				workerCount++
-			}
+	for {
+		workers := conn.SelectFromMinion(func(m db.Minion) bool {
```
ejj: You know, I think I would just use the backoff wait for thing.

kklin: What do you think the timeout should be? Do you think we should exit completely if the timeout expires?

minion/scheduler/scheduler.go
```diff
@@ -44,16 +44,23 @@ func Run(conn db.Conn, dk docker.Client) {
...
-		return
-	}
-
 	for {
-		workers := conn.SelectFromMinion(func(m db.Minion) bool {
-			return m.Role == db.Worker
-		})
-		if len(workers) > 0 {
-			return
+		switch conn.MinionSelf().Role {
```
ejj: I think this would be a lot easier to read if you just had a separate waitMaster() and waitWorker() function and did the role check in the caller.  Would make the history really clean too if you could progage that back in the PR.

minion/scheduler/scheduler.go
ejj: I might make it really long and then panic if it expires.

kklin: OK -- how long is long? Is 1 hour reasonable?

ejj:
Yeah I think so.

PR #1218, opened 2017-10-26
docs: Tweak overview and add blueprint index

luise:
For the blueprint index, I chose to leave out only Tensorflow and Django. I'm open to
adding them, and to removing some of the blueprints that are currently on the list.

For the overview, the new version looks like this:
screen shot 2017-10-26 at 4 03 19 pm

As opposed to:
screen shot 2017-10-26 at 4 03 12 pm


docs/_SupportedBlueprints.md
```diff
@@ -0,0 +1,17 @@
+# Blueprints Supported by Kelda
+Kelda supports and maintains blueprints for a number of popular applications.
```
kayousterhout: I wonder if we shouldn't say "Kelda" here to allow us to later link others' blueprints (as you mentioned below) -- so more like "Kelda blueprints have been written for many popular applications, which are listed below."

docs/_SupportedBlueprints.md
```diff
@@ -0,0 +1,17 @@
...
+If you would like your blueprint to be on the list, feel free to [submit a pull request](https://github.com/kelda/kelda/blob/master/docs/_SupportedBlueprints.md)!
+
+* [Elasticsearch](https://github.com/kelda/elasticsearch)
+* [etcd](https://github.com/kelda/etcd)
+* [HAProxy](https://github.com/kelda/haproxy)
+* [Kibana](https://github.com/kelda/kibana)
+* [Lobste.rs](https://github.com/kelda/lobsters)
```
kayousterhout: I'd probably remove this (since unlike the other lower level systems, I don't think this is something that people are super likely to build on).  Or maybe put it in a separate "Applications" section? (and this could be "Systems" or something?)

docs/_SupportedBlueprints.md
```diff
@@ -0,0 +1,17 @@
...
+* [MEAN stack](https://github.com/kelda/mean) (MongoDB, Express, Angular, and Node.js)
+* [MongoDB](https://github.com/kelda/mongo)
+* [Nginx](https://github.com/kelda/nginx)
+* [Node.js with MongoDB](https://github.com/kelda/nodejs)
+* [Redis](https://github.com/kelda/redis)
+* [Slackin](https://github.com/kelda/slackin)
```
kayousterhout: ditto re:lobsters comment

README.md
```diff
@@ -79,7 +79,8 @@ let vm = new Machine({
 });
 ```
 
-See [full example blueprints](https://github.com/kelda/) and [check out our docs](http://docs.kelda.io).
+See [full example blueprints](http://docs.kelda.io/#blueprints-supported-by-kelda)
```
kayousterhout: I'd say something more bold here since "example" sounds a little like we're not directly recommending to use them. Maybe "See [the Library of existing Kelda blueprints](link...)"?

docs/_SupportedBlueprints.md
```diff
@@ -0,0 +1,17 @@
+# Blueprints Supported by Kelda
```
kayousterhout: What about calling this "Kelda blueprint library"?

docs/index.html.md
```diff
@@ -20,6 +20,7 @@ includes:
   - CloudProviders
   - Dev
   - Security
+  - SupportedBlueprints
```
kayousterhout: Can you add a link to this in the overview section too?

docs/_SupportedBlueprints.md
luise: Yeah, I thought something similar, but then I decided that the Kelda subtitle would look weird if there isn't a "non-kelda section".
I like your rewording of the text though, as it's more flexible in terms of who has written the blueprints. 
Does it sound ok to just update the text, but leave out the subtitle?

README.md
luise: Ah, yeah, I see your point. I'd probably reword a little to clarify that it's not all the blueprints that exist, but just the ones we vouch for. Maybe just "See the Library of blueprints supported by Kelda"?

docs/_SupportedBlueprints.md
luise: Ah, maybe "Kelda blueprints have been written for many popular applications. Here are some that we recommend trying out:"
Then it's clear that this is just a subset

docs/_SupportedBlueprints.md
luise: I like that it's more succinct, but I'm a little wary of making it sound like these are the only blueprints that exist. 

Maybe the title could be "Blueprint library", and then the text could be what I suggested above with "here are some that we recommend." What do you think?

docs/_SupportedBlueprints.md
luise: I'm fine removing it. It does stick out a little, and  I think I'd rather avoid categorizing - it isn't quite clear where they all fit. E.g. MEAN or Node.js with Mongo.

It might still be helpful as an example for people trying to build blueprints, but I guess they'll find it through the blueprints writers guide.

docs/_SupportedBlueprints.md
luise: I guess we should decide what the purpose of this list is. I'm not sure the idea is to only list blueprints that users will build on. 
I think it might as well be for users to see example blueprints, and also potentially to attract blueprint writers. A person with a popular open source project might see how easy we make it to deploy slackin and lobste.rs and realize Kelda would be helpful for their own project as well.

I would be in favor of just adding a good number, and then we can always filter later if we end up having too many. But I don't think that's a problem right now.

docs/index.html.md
luise: Yup, good idea!

luise:
@kayousterhout , I updated the PR based on your comments and the discussion on slack

docs/_BlueprintLibrary.md
```diff
@@ -0,0 +1,11 @@
+# Blueprint Library
+Kelda blueprints have been written for many popular applications. Here are some
+that we recommend trying out:
+
+* [HAProxy](https://github.com/kelda/haproxy)
+* [MEAN stack](https://github.com/kelda/mean) (MongoDB, Express, Angular, and Node.js)
+* [Slackin](https://github.com/kelda/slackin)
+* [Spark](https://github.com/kelda/spark)
+
+If you would like your blueprint to be on the list, feel free to
```
kayousterhout: super nit: the --> this

docs/_BlueprintWritersGuide.md
```diff
@@ -6,6 +6,9 @@ blueprint for a new application, using the lobste.rs application as an example.
 lobste.rs is an open source project that implements a reddit-like web page,
 where users can post content and vote up or down other content.
 
+For more blueprint examples, check out [some blueprints we think are
```
kayousterhout: how about just [the blueprints in the blueprint library]? I'm worried the current text is an over-endorsement of how great we think those blueprints are.  Feel free to leave if you feel strongly.

luise: I don't feel strongly either, so we should probably rather under-endorse! I'll update!

luise:
@kklin , ready!

PR #1219, opened 2017-10-27
ci: Identify Spark master by its hostname

kklin:
A recent change to the Spark blueprint changed the command run to start
Spark masters. This commit changes the Spark tests to identify the
masters by their hostname, rather than command. This fixes the
integration tests that were failing because they were unable to find the
masters.


kayousterhout:
Thanks for fixing this Kevin!

PR #1222, opened 2017-10-28
provider: Support DiskSize for GCE

HongjianZhang:
Before this patch, DiskSize was only supported for AWS. This patch
allows booting machines with user-specified DiskSize on GCE.


ejj:
How did you test this btw?

kklin:
The `DiskSize` should also be properly set when `List` is called

HongjianZhang:
I think each machine can be attached multiple disks, each disk having a diskSize. Since the DiskSize property of db.Machine has type int, should that property simply be set as the sum of all the disks?

ejj:
That definitely won't work.  Have a look at how the caller uses List and see if you can figure out a solution for that situation.

HongjianZhang:
I think I misunderstood how `List()` is used. I noticed that the `parseDiskSize` function in the Amazon code only extracts the size of the first disk from the BlockDeviceMappings. I am not sure why, but I did something similar here, where I only extract the size of the first attached disk. Can someone confirm this?

HongjianZhang:
Need some more changes before it can be reviewed again.

kklin:
Closing this since nobody is really making use of the DiskSize feature, and this PR isn't ready to be merged yet.

PR #1225, opened 2017-10-29
tests: Refactor the network test

ejj:
Before this patch, the network test code was extremely complicated and
slow.  SSH attempts tended to come in batches, causing the system to
either bottleneck on the concurrency limit, or waste time trying to
figure out what to do next.  This patch refactors the test and in the
process dramatically simplifies the code. It additionally causes SSH
executions to smooth out at a rate of once every 100ms which results
in a roughly 3x improvement in test time in my local environment.
This makes the network test much more palatable as a development tool.

### DIFF
--- a/integration-tester/tests/10-network/network_test.go
+++ b/integration-tester/tests/10-network/network_test.go
@@ -3,7 +3,9 @@ package main
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"os/exec"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -59,7 +61,7 @@ func TestConnectivity(t *testing.T) {
 }
 
 type testerIntf interface {
-	test(c db.Container) []error
+	test(t *testing.T, c db.Container)
 }
 
 type commandTime struct {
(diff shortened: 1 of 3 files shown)


integration-tester/tests/10-network/network_test.go
```diff
@@ -75,47 +77,27 @@ func (ct commandTime) String() string {
...
+// We have to limit our parallelization because each `kelda ssh` creates a new SSH login
+// session. Doing this quickly in parallel breaks system-logind on the remote machine:
+// https://github.com/systemd/systemd/issues/2925.  Furthermore, the concurrency limit
+// cannot exceed the sshd MaxStartups setting, or else the SSH connections may be
+// randomly rejected.
+var sshRateLimit = time.Tick(100 * time.Millisecond)
+
 func keldaSSH(id string, cmd ...string) (string, error) {
+	<-sshRateLimit
+	fmt.Println("kelda ssh", id, strings.Join(cmd, " "))
```
ejj: Note I wasn't sure if it's a good idea to include this or not.  It makes the logs really really long, but when you're running as a dev it's somewhat nice to see.  Thoughts?

ejj:
Eh, this still needs some work and I'm to tired to figure it out.  Will reopen later.

PR #1226, opened 2017-10-30
Tweak bindings and remove deprecated functions for Deployment

luise:
This PR removes the deprecated Deployment class and related functions in favor of the new Infrastructure class. 

It additionally renames the optional object parameters to `opts`, so attribute names in the documentation will be shorter.


luise:
I kept the `Deployment` class around because we use it for initializing`global._keldaDeployment` and in `bindingsTest.js` to initialize a deployment without machines. @kayousterhout , I think you might have said to remove the whole class during our planning meeting, so let me know if you still think that's a better solution.

kayousterhout:
@luise can you move the other bindings tweaks to a different PR, or update the title and description (to make these easy to find in the future)? It's confusing that the title and description don't currently match because they describe different commits in the PR.

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ it would never enter the "connected" state.
 JavaScript API-breaking changes:
 - Remove the Container.replicate() method. Users should create multiple
 containers by looping.
```
kayousterhout: I think we decided "bindings" is the desired prefix for changes like this

js/bindings/bindings.js
```diff
@@ -1054,11 +1054,11 @@ class Container {
    * @param {Object} [optionalArgs] - Additional, named, optional arguments.
    * @param {string} [optionalArgs.command] - The command to use when starting
    *   the container.
-+  * @param {Object.<string, string|Secret>} [optionalArgs.env] - Environment
```
kklin: My bad, thanks!

js/bindings/bindings.js
```diff
@@ -756,11 +717,8 @@ class Machine {
    *   and Vagrant. This argument is optional, but the provider attribute of the
    *   machine must be set before it is deployed.
    * @param {string} [opts.role] - The role the machine will run as
-   *   (accepted value are Master and Worker). A Machine's role must be set before
-   *   it can be deployed.  This argument is not required, so that users can
-   *   create a template to use for all machines in the cluster;
-   *   {@link Machine#asWorker} and {@link Machine#asMaster} can be called on the
-   *   template to create a machine with the appropriate role, as in the example.
+   *   (Master or Worker). Users don't have to provide this argument since the
+   *   {@link Infrastructure} constructor will set the appropriate value.
```
kayousterhout: Should we just delete this? It seems like users shouldn't know about it and shouldn't pass it in here, because it will be overridden

js/bindings/bindings.js
```diff
@@ -911,7 +869,9 @@ class Machine {
   }
 
   /**
-   * Clones this machine, and adds the given role.
+   * Clones this machine, and adds the given role. Users should not have to use
+   * this function, since the {@link Infrastructure} constructor will set the
+   * machine role.
```
kayousterhout: Same comment as above - can we just delete withRole?

js/bindings/bindings.js
```diff
@@ -1537,7 +1475,6 @@ module.exports = {
   Secret,
   LoadBalancer,
   allow,
-  createDeployment,
   getDeployment,
```
kayousterhout: Can you rename this to getInfrastructure, and rename the underlying variable accordingly?

js/bindings/bindings.js
luise: The `Infrastructure` constructor uses it, but it would probably be better to get rid of the function to simplify the exposed API. I'll fix it

js/bindings/bindings.js
luise: Ah, yes, good point.

luise: Oh, actually, we should probably leave this because users can still deploy Machines with the `Machine.deploy()` function.
And then I should reword the description to say that the user has to set the role unless they're using the Infrastructure constructor (recommended).
Does that sound ok?

kayousterhout: Do we need to let users call Machine.deploy()?

kklin: FWIW I'd be fine with removing it.

luise: I think it's kind of nice for consistency -- i.e., you can deploy everything else, so why not machines? But that being said, it is just another way of doing something you can already do (with Infrastructure). I'll go ahead and delete it.

js/bindings/bindings.js
```diff
@@ -52,11 +55,32 @@ class Deployment {
...
+
+    boxedMasters.forEach(master =>
+      this.machines.push(machineWithRole(master, 'Master')));
+    boxedWorkers.forEach(worker =>
+      this.machines.push(machineWithRole(worker, 'Worker')));
   }
 
   /**
-   * Converts the deployment to the QRI deployment format.
+   * Converts the infrastructure to the Kelda infrastructure format.
```
kayousterhout: Can you change this to be "to the JSON format expected by the Kelda go code"?

js/bindings/bindings.js
```diff
@@ -724,29 +674,24 @@ class Machine {
    * Creates a new Machine object, which represents a machine to be deployed.
    * @constructor
    *
-   * @example <caption>Create a template Machine on Amazon, and then use the
-   * template to create a master and 2 workers. This will use the default size
-   * and region for Amazon.</caption>
+   * @example <caption>Create a template Machine on Amazon. This will use the
```
kayousterhout: I'd eliminate the word "template" here, since now it's not obvious what that means

js/bindings/bindings.js
```diff
@@ -1346,7 +1239,7 @@ class Container {
   }
 
   /**
-   * Converts the Container to the QRI deployment format.
+   * Converts the Container to the Kelda infrastructure format.
```
kayousterhout: Same as above re: json

js/bindings/bindings.js
```diff
@@ -243,12 +207,9 @@ function baseInfrastructure(name = 'default') {
     throw new Error(`no infrastructure called ${name}. Use 'kelda init' ` +
       'to create a new infrastructure.');
   }
-  return getInfraDeployment(infraPath);
+  return getBaseInfrastructure(infraPath);
 }
 
-// The default deployment object. The Deployment constructor overwrites this.
-global._keldaDeployment = new Deployment({});
```
kayousterhout: Now that this has been removed, does the behavior when it's not defined change? (e.g., if you deploy an empty blueprint, what happens?)

js/bindings/bindingsTest.js
```diff
@@ -15,7 +15,8 @@ describe('Bindings', () => {
   let deployment;
```
kayousterhout: can you rename this to infra?

js/bindings/bindings.js
luise: Yeah, this isn't ready -- but you're quick on the reviews 😄  As you say, this doesn't work!
I was planning to make the `toKeldaRepresentation` a global function that takes the infrastructure as an argument. If the Infrastructure is nil, it just returns an empty deployment object. I'll push a commit over the weekend.

luise:
@kayousterhout , updated! Maybe @kklin also has an opinion on the changes to the global deployment object.

scripts/blueprints-tester/tests/test_readme.go
```diff
@@ -122,13 +122,20 @@ func TestReadme() error {
 		return fmt.Errorf("failed to parse README: %s", err.Error())
 	}
 
+	if err := run("npm", "link", "../../js/bindings"); err != nil {
```
kklin: Could you get the absolute path to the bindings here, and then `npm install` that path directly once you're in the workdir? I think that's better than `npm link` since that messes with the global node_modules

kklin:
@luise seems fine to me, it's a hack either way really..

scripts/blueprints-tester/tests/test_readme.go
luise: Makes sense, I updated it!

js/bindings/bindingsTest.js
```diff
@@ -138,16 +141,16 @@ describe('Bindings', () => {
...
         provider: 'Amazon',
         region: 'us-west-2',
         size: 'm4.large',
         cpu: new b.Range(2, 4),
         ram: new b.Range(4, 8),
         diskSize: 32,
         sshKeys: ['key1', 'key2'],
-      })]);
+      });
+      infra = new b.Infrastructure(machine, machine);
```
kayousterhout: This is slightly confusing now -- because there's an infra with 2 of these machines with a selected size, but you're only checking one (the worker).  Do you think it makes sense to remove the "role" from all of these tests, since it doesn't really matter which worker gets checked?  It would also be great to add a comment about this in the first test that does it

js/bindings/bindingsTest.js
```diff
@@ -159,39 +162,42 @@ describe('Bindings', () => {
       }]);
     });
     it('chooses size when provided ram and cpu', () => {
-      deployment.deploy([new b.Machine({
-        role: 'Worker',
+      const machine = new b.Machine({
         provider: 'Google',
+        region: 'us-east1-b',
```
kayousterhout: Why did you need to add this?

js/bindings/bindingsTest.js
```diff
@@ -316,13 +322,14 @@ describe('Bindings', () => {
...
           provider: 'Amazon',
         },
-      ]);
+      ], 3);
     });
     it('replicate independent', () => {
-      const baseMachine = new b.Machine({ provider: 'Amazon' });
-      const machines = baseMachine.asMaster().replicate(2);
+      const baseMachine = new b.Machine(
+        { provider: 'Amazon' });
```
kayousterhout: why'd this line wrap need to change?

js/bindings/bindingsTest.js
```diff
@@ -316,13 +322,14 @@ describe('Bindings', () => {
           role: 'Master',
           provider: 'Amazon',
         },
-      ]);
+      ], 3);
     });
     it('replicate independent', () => {
-      const baseMachine = new b.Machine({ provider: 'Amazon' });
-      const machines = baseMachine.asMaster().replicate(2);
+      const baseMachine = new b.Machine(
```
kayousterhout: I think you can remove this test, since the Infrastructure constructor already copies things, so it's not necessary that replicate does (and, if a test like this doesn't exist already, add a test that the Infra constructor clones machines so handles the same machine being passed in as both master and worker)

js/bindings/bindingsTest.js
```diff
@@ -331,18 +338,18 @@ describe('Bindings', () => {
...
-      const baseMachine = new b.Machine({
-        provider: 'Amazon',
-        floatingIp: 'xxx.xxx.xxx.xxx',
-      });
-      deployment.deploy(baseMachine.asMaster());
+      const worker = new b.Machine({ provider: 'Amazon' });
+      const machineWithFloatingIP = new b.Machine({
+        provider: 'Amazon', floatingIp: 'xxx.xxx.xxx.xxx' });
+
+      infra = new b.Infrastructure(machineWithFloatingIP, worker);
```
kayousterhout: why does the master (as opposed to the worker) get the floating IP here?

js/bindings/bindingsTest.js
```diff
@@ -835,75 +845,59 @@ describe('Bindings', () => {
...
+      const amazonMachine = new b.Machine({ provider: 'Amazon', region: '' });
+      const doMachine = new b.Machine({ provider: 'DigitalOcean', region: '' });
+      infra = new b.Infrastructure(amazonMachine, doMachine);
       expect(deploy).to.throw('All machines must have the same provider and region. '
         + 'Found providers \'Amazon\' in region \'us-west-1\' and \'DigitalOcean\' in '
         + 'region \'sfo1\'.');
     });
   });
-  describe('Deployment', () => {
-    it('no args', () => {
```
kayousterhout: Should these tests move to the Infrastructure test?

js/bindings/bindingsTest.js
```diff
@@ -1048,7 +1045,18 @@ describe('Bindings', () => {
         connections: [],
         containers: [],
         loadBalancers: [],
-        machines: [],
+        machines: [
+          {
+            id: 'a45b4bef049ee155e0c481af209bf40fe676fdd1',
```
kayousterhout: Do you need the id here?

js/bindings/bindings.js
```diff
@@ -234,7 +200,7 @@ function getInfraDeployment(infraPath) {
  *
  * @param {string} name - The name of the infrastructure, as passed to
  *   `kelda init`.
- * @returns {Deployment} A deployment object representing the infrastructure.
+ * @returns {Infrastructure} The infrastructure object.
```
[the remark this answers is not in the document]
luise: Good catch!

js/bindings/bindingsTest.js
```diff
@@ -34,32 +35,34 @@ describe('Bindings', () => {
     expect(dummyMachine.size).to.equal(expected);
   };
 
-  const checkMachines = function checkMachines(expected) {
-    const { machines } = deployment.toKeldaRepresentation();
-    expect(machines).to.have.lengthOf(expected.length)
-      .and.containSubset(expected);
+  // numMachines defaults to 2 because most test Infrastructures deploy one
+  // worker and one master.
+  const checkMachines = function checkMachines(expectedSubset, numMachines = 2) {
```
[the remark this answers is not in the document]
luise: Yeah, that sounds great. I wasn't sure it was OK to skip the length check, but if you think it is, I'd much rather get rid of it here!

js/bindings/bindingsTest.js
luise: Yup, I'd be fine removing the role, and I'll add a comment.

js/bindings/bindingsTest.js
luise: Oh, that was from a previous iteration. Will delete!

js/bindings/bindingsTest.js
luise: Wouldn't it matter in other cases? E.g. something like this: 
```javascript
const workers = baseMachine.replicate(5);
workers[0].floatingIP = 'x.x.x.x';
const inf = new Infrastructure(baseMachine, workers);
```

js/bindings/bindingsTest.js
luise: No need to

js/bindings/bindingsTest.js
luise: I figured it doesn't matter since we're just testing that the IP is set on the machine we expect, and existing test checked the floating IP on the master and not the worker.

js/bindings/bindingsTest.js
luise: Ah, yes, they should!

js/bindings/bindingsTest.js
luise: I kept them because they depend on the machine attributes. So if the IDs are the same, the Machines are indeed the same -- and this test actually wants to check that the infrastructure is the right one. 
Do you think they should be left out?

js/bindings/bindingsTest.js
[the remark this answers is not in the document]
luise: Ah, yeah, I wasn't sure if it was ok to remove the length check. If you think it is, I agree it would simplify things!

js/bindings/bindingsTest.js
kayousterhout: Ah good point you're right

js/bindings/bindingsTest.js
kayousterhout: In general I think we've over-used the id attribute -- the downside of using it is that it makes the tests super brittle, because anytime we change anything about the Machine class we have to go and update all of the IDs.  I'm not convinced it's necessary here because it seems more useful to just directly verify that the attributes of the machines are as expected.  But if you feel strongly fine to leave as-is!

js/bindings/bindingsTest.js
```diff
@@ -1036,4 +1036,23 @@ describe('Bindings', () => {
       revertGetInfra();
     });
   });
+  describe('getInfrastructureKeldaRepr()', () => {
+    it('should return an empty object when no Infrastructure was created', () => {
+      const revertKeldaInf = b.__set__('_keldaDeployment', undefined);
+      expect(global.getInfrastructureKeldaRepr()).to.deep.equal({});
+      revertKeldaInf();
```
kayousterhout: Why do you need this line? (everything seems to work without it)

js/bindings/bindingsTest.js
```diff
@@ -1036,4 +1036,23 @@ describe('Bindings', () => {
...
+    });
+    it('should return the correct infra object when an Infrastructure exists', () => {
+      const expected = {
```
kayousterhout: Can you add a comment that this is the expected infra based on the Infrastructure object initialized in the beginning of the test? (otherwise it's confusing why this is the expected infra).  Alternately it might be better to re-initialize an infrastructure here so that the expectation is super obvious

js/bindings/bindingsTest.js
```diff
@@ -835,75 +829,59 @@ describe('Bindings', () => {
...
+      (new b.Container('host', new b.Image('img', 'dk'))).deploy(infra);
+      (new b.Container('host', new b.Image('img', 'dk2'))).deploy(infra);
       expect(deploy).to.throw('img has differing Dockerfiles');
     });
     it('machines with same regions/providers', () => {
-      deployment.deploy([new b.Machine({
-        provider: 'Amazon',
-        region: 'us-west-2',
-      }), new b.Machine({
+      const machine = new b.Machine({
```
kayousterhout: I think it would better fit the intention of this test for two of the same machine to be passed in as workers -- since this is trying to check the behavior of the conversion, and by the time the machines are converted, they will be different, because one will have a master role and one will have a worker role

luise: I think the intention of this test is to check that there are no errors when two machines in the same provider and region are deployed (as opposed to the below test which checks that we throw an error when two machines are in different locations). So I think the test still has the original intention, even if the machines are not of the same role.
Do you still think I should change it?

kayousterhout: Ah I see fine to leave then

js/bindings/bindingsTest.js
luise: Yeah, so it still works without because the `beforeEach()` function sets a new infrastructure before the next test runs. I think it is just best practice to revert the values you mock out -- and you could imagine it causing bugs later on if we leave it out and change the `beforeEach()`.

Do you think I should rather get rid of it, even i

js/bindings/bindingsTest.js
luise: OK, I'll switch over using to the attributes! It's the same effect, but I see your point about the brittle tests that rely on IDs!

js/bindings/bindingsTest.js
kayousterhout: I'd take this out because aren't there a bunch of other tests that also change the infra object?

kayousterhout: (also I think the point of the beforeEach is to avoid needing to do things like this in the individual tests)

luise: Ok, I'll take it out

luise:
@kayousterhout , thanks for the thorough review. I addressed your comments. @kklin , I guess it's ready for you.

scripts/blueprints-tester/tests/test_readme.go
```diff
@@ -88,7 +88,6 @@ func (parser readmeParser) blocks() (map[string]string, error) {
 
 var dependencies = `{
   "dependencies": {
-    "kelda": "0.x",
```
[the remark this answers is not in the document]
luise: Yeah, so I don't think this line matters any more. Since we `npm install` the path, we don't need to install kelda with the `npm install` that uses `package.json`.
Are you ok with removing it? I'm also fine with leaving the line, but the `kelda` package will be overwritten anyways when we install with the path.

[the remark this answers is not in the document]
kklin: Eh, I have a slight preference for leaving it but I don't think it actually matters. I'll merge now!

PR #1228, opened 2017-10-31
foreman: Populate machines roles and connection status directly into the database

aegamesi:
Previously, cloud (and cloud/join) had to call various methods on the foreman to get a machine's role and connection status. This is unclean and a bit slow.

Now, the foreman directly populates machines' roles and connection statuses into the database. This has the added bonus of being a little bit faster, because the foreman directly knows the connection statuses (and doesn't have to infer them like cloud/join did).


cloud/foreman/foreman.go
```diff
@@ -87,12 +87,16 @@ func RunOnce(conn db.Conn) {
 		bp, _ := view.GetBlueprint()
 		blueprint = bp.Blueprint.String()
 
+		updateMinionMap(machines, view)
```
ejj: I really don't like the idea of moving these things into the database transaction.  Both functions make network connections to the minions which can take forever, during which time the entire daemon will freeze up.

cloud/foreman/foreman.go
```diff
@@ -93,6 +93,17 @@ func RunOnce(conn db.Conn) {
 	updateMinionMap(machines)
 	forEachMinion(updateConfig)
 
+	conn.Txn(db.MachineTable).Run(func(view db.Database) error {
+		for _, m := range machines {
+			view.Commit(m)
```
kklin: This doesn't seem safe to me since you might overwrite any changes that occurred in the database while you were outside the transaction.

aegamesi: Should I fetch the list of machines from the database again (in the transaction), and just update their statuses and roles?

kklin: I've done that in a couple other places, but it's pretty ugly. Maybe @ejj has a better idea.

cloud/cloud.go
```diff
@@ -262,4 +262,3 @@ func (cld cloud) String() string {
 // Stored in variables so they may be mocked out
```
luise: Nits for the commit message: "This makes the foreman...." --> "This commit makes the foreman..." and  "retrieve the a machines role" has a redundant the/a and is missing an `'` in machines (and possibly "role" --> "roles" if you meant `machines'` rather than `machine's`).

luise:
Hey @aegamesi! I have been running on top of your branch, and there seems to be some issues. The machine statuses and roles seem to get confused, and I sometimes get a panic when I `kelda stop` the deployment. Here are the daemon logs from one of the failed runs: https://gist.github.com/luise/5de33207c03087775501493bad8511fa. It seems to be pretty reproducible with the nginx/nginxExampleWithInfra.js blueprint (I changed it to have 2 workers).

aegamesi:
Okay, this definitely has some larger issues that I didn't realize before. I'm going to withdraw this for now and make a PR later.

PR #1229, opened 2017-10-31
foreman: Generate etcd and public key configuration based on database

kklin:
Before, we generated the etcd configuration and public key mappings
based on the `minions` map. However, when a minion disconnects, it is
removed from the map, causing the configuration that we send to the
minions to change. This can sometimes result in transient errors -- for
example, if a minion is removed from the public key mappings, then the
leader will revoke Vault access to it, causing the minion to log some
errors about connecting to Vault.

This commit avoids this problem by generating the minion config based on
the database, whose contents are not affected by the connection status
to the minion.


kklin:
This should fix the failing log test.

cloud/foreman/foreman.go
```diff
@@ -95,13 +95,13 @@ func RunOnce(conn db.Conn) {
 
 	minionIPToPublicKey := map[string]string{}
 	var etcdIPs []string
-	for _, m := range minions {
-		if m.config.Role == pb.MinionConfig_MASTER && m.machine.PrivateIP != "" {
-			etcdIPs = append(etcdIPs, m.machine.PrivateIP)
+	for _, m := range machines {
```
kayousterhout: Can you add a comment here about why you're using `machines` and not the minion map?  It's not intuitive, especially given the `updateMinionMap` call above

kklin: Done!

PR #1230, opened 2017-11-01
docs: Add documentation for secrets

kklin:


docs/Makefile
```diff
@@ -1,3 +1,6 @@
+# Always recompile the JavaScript documentation to pick up changes to bindings.js.
+.PHONY: _jsdoc.erb
```
kayousterhout: Why not add bindings.js as a dependency to _jsdoc.erb? (I think that's the more typical make-y way to do this)

docs/_HowTo.md
```diff
@@ -148,3 +148,56 @@ and check you get the right response:
 ```console
 $ curl -H "Host: apples.com" HAPROXY_PUBLIC_IP
 ```
+
+## How To Run Applications That Rely On Configuration Secrets
```
kayousterhout: super nit: don't capitalize "To", "That", and "On"

docs/_HowTo.md
```diff
@@ -148,3 +148,56 @@ and check you get the right response:
...
+This guide walks through an example of running an application that has sensitive
+information in its configuration. Note that Kelda secrets are currently only
+useful for _configuration_. Secrets generated at runtime, such as customer
+information that needs to be stored in a secure database, are not handled.
```
kayousterhout: I'd add "yet" after "are not" here

docs/_HowTo.md
```diff
@@ -148,3 +148,56 @@ and check you get the right response:
 ```console
 $ curl -H "Host: apples.com" HAPROXY_PUBLIC_IP
 ```
+
+## How To Run Applications That Rely On Configuration Secrets
+This guide walks through an example of running an application that has sensitive
```
kayousterhout: I realize that "guide" is used elsewhere here, but I think it's clearer to call this "section" to be totally obvious that it's just this part of the overall "how to" guide section.  Same thing for the first sentence of the next paragraph.

docs/_HowTo.md
```diff
@@ -148,3 +148,56 @@ and check you get the right response:
...
+information in its configuration. Note that Kelda secrets are currently only
+useful for _configuration_. Secrets generated at runtime, such as customer
+information that needs to be stored in a secure database, are not handled.
+
+In this guide, we will walk through deploying a GitHub bot that requires a
+GitHub OAuth token. We will deploy the `keldaio/bot` Docker image, and configure
```
kayousterhout: I'd add, after "token", " in order to get notifications from a private GitHub repository." (to disambiguate from the case where an application needs to continuously get and store OAuth tokens for each new user that uses the app)

docs/_HowTo.md
```diff
@@ -148,3 +148,56 @@ and check you get the right response:
...
 $ curl -H "Host: apples.com" HAPROXY_PUBLIC_IP
 ```
+
+## How To Run Applications That Rely On Configuration Secrets
+This guide walks through an example of running an application that has sensitive
+information in its configuration. Note that Kelda secrets are currently only
+useful for _configuration_. Secrets generated at runtime, such as customer
+information that needs to be stored in a secure database, are not handled.
+
+In this guide, we will walk through deploying a GitHub bot that requires a
```
kayousterhout: I think we were trying to move away from using "you" / "we" in the docs -- you can change this to be similar to the above paragraph: "This guide walks through deploying... token, by deploying a `keldaio/bot' Docker image and configuring its..." (one advantage of this change is that it's slightly more concise)

docs/_HowTo.md
```diff
@@ -148,3 +148,56 @@ and check you get the right response:
...
+    
+3. Set the secret value with `kelda secret githubToken <TOKEN_VALUE>`. If the
+    command succeeds, there will be no output, and the exit code will be zero.
+    
+    ```console
+    $ kelda secret githubToken the_value_of_the_secret
+    ```
+    
+    Note that Kelda does not handle the lifecycle of the secret before `kelda
+    secret` is run. For our GitHub token example, we might copy and paste the
```
kayousterhout: For the GitHub token example, the GitHub token can be copied directly from the GitHub web UI...

docs/_HowTo.md
```diff
@@ -148,3 +148,56 @@ and check you get the right response:
...
+    GitHub token directly from the GitHub web UI to the `kelda secret` command.
+    Another approach could be to store the token in a password manager, and
+    paste it into `kelda secret` when needed.
+    
+4. `kelda show` should show that the bot container has started. It may take up to
+    a minute for the container to start.
+    
+5. If you need to change the secret value, simply run `kelda secret githubToken
```
kayousterhout: To change the secret value, run `kelda ....

(same as above re: eliminating "you", and it's better to avoid [weasel words](http://matt.might.net/articles/shell-scripts-for-passive-voice-weasel-words-duplicates/) like "simply" in technical docs -- we shouldn't have to tell the user that it's simple because it should be obvious! This is a stylistic preference so if you feel strongly I'm fine with leaving it)

docs/_HowTo.md
```diff
@@ -148,3 +148,56 @@ and check you get the right response:
...
+    ```javascript
+    const container = new kelda.Container('bot', 'keldaio/bot', {
+      env: {
+        GITHUB_OAUTH_TOKEN: new kelda.Secret('githubToken'),
+      },
+    });
+    ```
+    
+2. Deploy the blueprint with `kelda run`, and wait for `kelda show` to show that
+    our container is waiting for the `githubToken` secret. The output should look
```
kayousterhout: This section makes it sound like the user _has_ to wait for this before installing a secret.  How about merging part of this with 3., resulting in something like:

(the reason for the change to 2. is that some users have said they skim quickly and only run things that are obviously code snippets, so it's too easy to miss the `kelda run` in the text)

"2. Deploy the blueprint:

```
$ kelda run <blueprintName.js>
```

3. Kelda will not launch a container until all secrets needed by the container have been added to Kelda.  If you run `kelda show` after deploying the blueprint, the output will look like this:
...
This means that Kelda is waiting to launch the container until the secret called `githubToken` is set.   To set the secret value, use `kelda secret`:

```
$ kelda secret githubToken <TOKEN_VALUE>
```
If the command succeeds... (existing text here)"

docs/_Security.md
```diff
@@ -62,3 +62,16 @@ Used for connecting to the cluster.
 Used for connecting to the cluster.
 
 Other files in the directory are ignored by Kelda.
+
+## Secrets
+Kelda uses Vault to securely store values for container environment variables
+and files. For an example of how to use secrets, see [How To Run Applications
+That Rely On Configuration Secrets](#how-to-run-applications-that-rely-on-configuration-secrets).
+
+Although all Kelda control traffic is encrypted in transit, only secrets are
```
kayousterhout: Can you just say "Secrets are encrypted both in transit and at rest."? I'm not sure what to make of the "Kelda control traffic" part -- it sent me down this path of wondering whether that was bad and what the repercussions of that are

js/bindings/bindings.js
```diff
@@ -1047,6 +1047,16 @@ class Container {
    * const container = new Container(
    *   'myApp', 'nginx', {filepathToContent: {'/etc/myconf': 'foo'}});
    *
+   * @example <caption>Create a Container that has one regular, and one secret
+   * environment variable value. The value of `mySecret` is defined by running
```
kayousterhout: is defined --> must be defined

and add " (if the blueprint with the container is launched before `mySecret` has been added, Kelda will wait to launch the container until the secret's value has been defined)." (helpful for the user to make these semantics super clear, I think)

docs/Makefile
kklin: Ah that's much nicer!

docs/_HowTo.md
kklin: Does saying "in order to push to a private GitHub repository" sound OK? You don't use OAuth tokens to get notifications from repositories, I think.

kklin:
@kayousterhout updated according to your comments!

docs/_Security.md
```diff
@@ -62,3 +62,15 @@ Used for connecting to the cluster.
 Used for connecting to the cluster.
 
 Other files in the directory are ignored by Kelda.
+
+## Secrets
+Kelda uses Vault to securely store values for container environment variables
+and files. For an example of how to use secrets, see [How To Run Applications
```
kayousterhout: super super nit: fix capitalization of "To", "That", and "On" for consistency

kayousterhout:
This looks super helpful thanks for doing this!

PR #1234, opened 2017-11-02
Various small tweaks.

ejj:
These things came up as I was hacking on scale.  More to come.

### DIFF
--- a/cloud/credentials.go
+++ b/cloud/credentials.go
@@ -41,26 +41,25 @@ func syncCredentialsOnce(conn db.Conn, sshKey ssh.Signer, ca rsa.KeyPair) {
 	machines := conn.SelectFromMachine(func(m db.Machine) bool {
 		return m.Status != db.Stopping && m.PublicIP != "" && m.PublicKey == ""
 	})
+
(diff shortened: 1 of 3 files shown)


cloud/credentials.go
```diff
@@ -41,26 +41,25 @@ func syncCredentialsOnce(conn db.Conn, sshKey ssh.Signer, ca rsa.KeyPair) {
...
+		for _, dbm := range view.SelectFromMachine(nil) {
+			if cert, ok := ipCertMap[dbm.PublicIP]; ok {
+				dbm.PublicKey = cert
+				view.Commit(dbm)
 			}
-			return nil
-		})
-	}
+		}
+		return nil
```
kayousterhout: do you need this return?

ejj: I think so, the transaction function returns an error.  I can try it without it and see if it builds.

ejj: Nope doesn't build without it.

kayousterhout: Ahh got it didn't realize that

PR #1235, opened 2017-11-03
bindings: Check for invalid machine sizes

HongjianZhang:
Before this patch, the Kelda daemon would keep re-trying when a
user specified an invalid size, which would result in an infinite
stream of log messages about it. This patch verifies size is valid
if a user specifies one, and throws an error if it is not valid.


js/bindings/bindingsTest.js
```diff
@@ -159,6 +108,56 @@ describe('Bindings', () => {
         size: 'n1-standard-2',
       }]);
     });
+    it('chooses size when ram and cpu do not have max', () => {
```
kklin: Why'd you move these unit tests? It makes the diff a bit harder to read.

js/bindings/bindingsTest.js
```diff
@@ -171,6 +170,45 @@ describe('Bindings', () => {
         size: 'm3.medium',
       }]);
     });
+    it('errors if provided size does not meet RAM or ' +
```
kklin: Thanks for writing these tests!

js/bindings/bindings.js
```diff
@@ -813,29 +815,65 @@ class Machine {
...
+    }
+    for (let i = 0; i < providerDescriptions.length; i += 1) {
+      const description = providerDescriptions[i];
+      if (this.size !== '' && this.size === description.Size) {
+        if (ram.inRange(description.RAM) &&
+            cpu.inRange(description.CPU)) {
+          return;
+        }
+        throw new Error(`Requested size '${this.size}' does not meet`
+          + ` RAM '(${ram.min}, ${ram.max})' or`
```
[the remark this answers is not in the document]
HongjianZhang: Good point. In the case of undefined `max`. Would it be better to represent it with 'inf', or should I leave it as 'undefined'?

js/bindings/bindingsTest.js
HongjianZhang: It's a bit complicated. So before this patch, since `chooseBestSize` is not exported, the way I tested `chooseBestSize` independently is by creating a dummy `machine`, and with a dummy `size`, so that `chooseSize` would not be called upon constructing the machine, then I can call `machine.chooseBestSize`. However after implementing `verifySize`, it will try to verify the dummy `size`, so all those tests failed. So instead of using dummy sizes, I just tested on real ones. And also it's more clear to have them under unit tests for machine.

js/bindings/bindings.js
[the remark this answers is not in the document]
kklin: I think "inf" is good!

js/bindings/bindings.js
```diff
@@ -813,26 +814,60 @@ class Machine {
...
         break;
       case 'DigitalOcean':
-        this.chooseBestSize(digitalOceanDescriptions.Descriptions, cpu, ram);
+        providerDescriptions = digitalOceanDescriptions.Descriptions;
         break;
       case 'Google':
-        this.chooseBestSize(googleDescriptions.Descriptions, cpu, ram);
+        providerDescriptions = googleDescriptions.Descriptions;
         break;
       case 'Vagrant':
```
kklin: Nit -- could you make the Vagrant check a separate if statement about the `switch` statement? I think that's a bit better since then this switch is just for getting the provider descriptions, and it's easy to gloss over the Vagrant part if it's within the switch

js/bindings/bindings.js
```diff
@@ -813,26 +814,60 @@ class Machine {
...
+   * If so, verifies the requested machine size satisfies CPU and RAM requirements.
+   * @private
+   * @param {description[]} providerDescriptions - Array of descriptions of
+   *   a provider.
+   * @param {Range} cpu - The desired number of CPUs.
+   * @param {Range} ram - The desired amount of RAM in GiB.
+   * @returns {void}
+   */
+  verifySize(providerDescriptions, cpu, ram) {
+    if (providerDescriptions === undefined) {
```
kklin: Is there any need for this `if` since I don't think the caller ever passes in an `undefined` providerDescriptions? I think this method would be simpler to read if we removed this check, and I don't think any callers would assume that you could pass in an `undefined` providerDescriptions.

js/bindings/bindings.js
```diff
@@ -853,7 +888,8 @@ class Machine {
 
   /**
    * Iterates through all the decriptions for a given provider, and returns
-   * the cheapest option that fits the user's requirements.
+   * the cheapest option that fits the user's requirements. Throws an error
```
kklin: Thanks for updating the jsdoc! It's easy for this stuff to get stale if we're not careful

js/bindings/bindings.js
HongjianZhang: Do you mean check/handle the Vagrant case before the `switch` statement?

kklin: Yeah, sorry!

kklin: (edited to fix the typo)

HongjianZhang:
Right now it is failing the integration test because nginx is trying to create an AWS machine with size `t2.micro` which is not in the Amazon descriptions file. It is however in the file of nonPreemptible machines, but that file has no RAM or CPU information of the size; therefore we can't check if this size fulfill user's RAM and CPU requirements. I am not too sure what's the best way to resolve this.

kklin:
I think we have to add the `t2.micro` information to the amazon descriptions (it'd be cleanest if you did this in a commit before your commit that checks for valid size)

PR #1236, opened 2017-11-03
Update Spark integration test and add integration test docs

kayousterhout:


docs/_Dev.md
```diff
@@ -146,6 +146,44 @@ To run the JavaScript tests for `bindings.js`, use the `jscheck` build target:
 $ make jscheck
 ```
 
+### Running the Integration Tests
```
kklin: Thanks for adding this!

docs/_Dev.md
```diff
@@ -146,6 +146,44 @@ To run the JavaScript tests for `bindings.js`, use the `jscheck` build target:
...
+$ kelda daemon
+$ # Open a new window to run the blueprint
+$ kelda run ./tests/20-spark/spark.js
+```
+
+Use `kelda show` to check the status of the blueprint.  Once all of the
+virtual machines and containers are up, run the Go test code:
+
+```console
+$ go test -v ./tests/20-spark/spark_test.go
```
kklin: `go test` can take a package, so this could just be go test -v ./tests/20-spark. I'm not sure what would happen if you run go test on a test file that also depends on other files in the package (such as the tests in 10-network).

Also, it might be worth mentioning what the -v flag does. I wouldn't run the logs test with -v since it dumps so much output to the console. I'd also be happy to make this change, what's PR'd here is way better than what we had before!

kklin:
@HongjianZhang I'll just review this, since I'm really familiar with our integration tests

docs/_Dev.md
kayousterhout: Updated!

integration-tester/tests/20-spark/spark_test.go
```diff
@@ -36,14 +37,32 @@ func TestCalculatesPI(t *testing.T) {
 		t.Fatal("unable to find BlueprintID of Spark master")
 	}
 
-	cmd := exec.Command("kelda", "ssh", id, "run-example", "SparkPi")
+	// Run a job that shuffles 100,000 key-value pairs using 100 mappers and 10
+	// reducers.
+	numMappers := 100
+	keyValuePairsPerMapper := 1000
+	cmd := exec.Command("kelda", "ssh", id, "run-example", fmt.Sprintf(
+		"GroupByTest %d %d %d %d", numMappers, keyValuePairsPerMapper, 4, 10))
```
kklin: What do you think about putting the `4` and `10` directly in the formatter string? They're also kind of mysterious, do you think it would be helpful to give them a name via the variable?

integration-tester/tests/20-spark/spark_test.go
```diff
@@ -36,14 +37,32 @@ func TestCalculatesPI(t *testing.T) {
...
 	output, err := cmd.Output()
```
kklin: I know the previous code was using `Output`, but this will only return `stdout`, and not `stderr` (which might be useful for debugging). If nothing gets printed to `stderr` during a successful run, I'd just call `CombinedOutput` instead. Otherwise, I'd set `cmd.Stderr` to a buffer we create, and capture the `stderr` output that way.

integration-tester/tests/20-spark/spark_test.go
kayousterhout: I was wondering about that -- I just made them named variables to minimize mysteriousness

integration-tester/tests/20-spark/spark_test.go
kayousterhout: There's a ton of stderr so I made it just log with the verbose flag; let me know if there's an alternative that you think is better here.

integration-tester/tests/20-spark/spark_test.go
```diff
@@ -36,14 +38,39 @@ func TestCalculatesPI(t *testing.T) {
...
+	valuesPerKey := 4
+	numReducers := 10
+	cmd := exec.Command("kelda", "ssh", id, "run-example",
+		fmt.Sprintf("GroupByTest %d %d %d %d",
+			numMappers, keyValuePairsPerMapper, valuesPerKey, numReducers))
+	stderrBytes := bytes.NewBuffer(nil)
+	cmd.Stderr = stderrBytes
+
+	stdoutBytes, err := cmd.Output()
+	t.Logf(stderrBytes.String())
```
kklin: Is there a `t.Log`? (we're not doing any formatting here)

kayousterhout:
This can be merged now and passes with the current blueprint

PR #1239, opened 2017-11-06
kelda: Allow containers to reference their public IP

kklin:


blueprint/blueprint.go
```diff
@@ -47,88 +47,127 @@ type Container struct {
...
-	val, isSecret := secretOrString.Value()
-	if isSecret {
-		return "Secret: " + val
+func (cv ContainerValue) String() string {
+	switch v := cv.Value.(type) {
+	case string:
+		return v
+	case Secret:
+		return "Secret: " + v.NameOfSecret
+	case RuntimeInfo:
```
ejj: part of me would prefer this is called `RuntimeValue` to be consistent with `ContainerValue`

Part of me also would like them both end with Data instead of Value, though that one I'm less confident in.

blueprint/blueprint.go
```diff
@@ -47,88 +47,127 @@ type Container struct {
...
-	}
+// ContainerPubIPKey is the RuntimeInfo resource key for the public IP of the
+// machine running the container. In other words, it's the IP at which the
+// container can be reached from the public internet.
+const ContainerPubIPKey = "host.ip"
+
+// RuntimeInfo represents metadata about a container that is only known when
+// the container is about to be booted. Currently, the only valid resource key
+// is defined by the ContainerPubIPKey constant.
+type RuntimeInfo struct {
```
ejj: I think I'd prefer these be defined further up in the file.  Near the top.

blueprint/blueprint.go
```diff
@@ -47,88 +47,127 @@ type Container struct {
...
+	ResourceKey string
```
ejj: Btw did you consider simply `type RuntimeInfo string`?

kklin: We need the `ResourceKey` field for JSON unmarshalling

ejj: Oh also, I wonder if simply `Key` or `Name` would be better.  Sounds less jargony, and it doesn't really tie us to resources.

integration-tester/tests/runtime-info/runtimeInfo.js
```diff
@@ -0,0 +1,19 @@
+const kelda = require('kelda');
```
ejj: Thanks for adding an integration test!

integration-tester/tests/runtime-info/runtimeInfo.js
```diff
@@ -0,0 +1,19 @@
...
+const infrastructure = require('../../config/infrastructure.js');
+const crypto = require('crypto');
+
+const infra = infrastructure.createTestInfrastructure();
+
+const indexPath = '/usr/share/nginx/html/index.html';
+
+const webContainer = new kelda.Container('web', 'nginx', {
```
ejj: Would be cool if you could get the nginx container to return the content of it's environment variables in the body of it's response.  That way you're actually testing that it makes it's way into the container.  Should be trivial if you just deploy a docker file that references nginx and set up a config file or something.

That said, not a huge deal, I don't think it should block merging of the PR.

kklin: Talked in person, the test is actually fine.

blueprint/blueprint.go
kklin: I just renamed `RuntimeValue`, not sure I agree with naming it `ContainerData` as that sounds like container metadata to me.

blueprint/blueprint.go
kklin: This isn't exposed to blueprint users (other than through the `kelda run` diff) since users reference the `hostIP` variable. And the way the JSON unmarshalling works it only contains the field names. So I like that `ResourceKey` is a bit more descriptive than just `Key` (and less likely to conflict)

PR #1242, opened 2017-11-07
Fix network integration tests

kklin:


minion/network/load_balancer.go
```diff
@@ -125,7 +125,8 @@ func updateLoadBalancerARP(client ovsdb.Client, loadBalancers []db.LoadBalancer)
 	}
 	// Ignore the order of `loadBalancers`.
 	sort.Strings(loadBalancedIPs)
-	expAddresses := strings.Join(append(loadBalancedIPs, ipdef.LoadBalancerMac), " ")
+	expAddresses := strings.Join(
+		append([]string{ipdef.LoadBalancerMac}, loadBalancedIPs...), " ")
```
ejj: Sigh, that really should have worked ...  OVN isn't following the OVSDB spec properly.

Anyways, thanks for fixing this.  You brought it up in review, I should have booted the system.

minion/network/acl.go
```diff
@@ -78,7 +78,7 @@ func updateACLs(ovsdbClient ovsdb.Client, connections []db.Connection,
 			expACLs = append(expACLs, directedACLs(
 				ovsdb.ACL{
 					Core: ovsdb.ACLCore{
-						Action:   "allow",
+						Action:   "allow-related",
```
ejj: I think this deserves a comment.  The reasoning for doing it with ICMP and nowhere else is somewhat subtle.

minion/network/load_balancer.go
kklin: No problem, I know this is just a small set of a large number of changes

PR #1243, opened 2017-11-08
Command line lookup container by hostnames

ejj:


api/util/util.go
```diff
@@ -3,11 +3,51 @@ package util
...
 	"fmt"
 
+	"github.com/kelda/kelda/api/client"
 	"github.com/kelda/kelda/db"
 )
 
-// GetContainer retrieves the container with the given blueprintID.
-func GetContainer(containers []db.Container, blueprintID string) (db.Container, error) {
+// FuzzyLookup finds either a container or a machine that's called `name` by querying the
+// provided `client`.  `name` could a container's hostname, the prefix of a container's
```
kklin: `name` could _be_ a

api/util/util.go
```diff
@@ -3,11 +3,51 @@ package util
...
+		container, containerError = findContainer(containers, name)
+	}
+
+	resolvedMachine := machineError == nil
+	resolvedContainer := containerError == nil
+	switch {
+	case !resolvedMachine && !resolvedContainer:
+		return nil, "", fmt.Errorf("%s, %s", machineError, containerError)
+	case resolvedMachine && resolvedContainer:
+		return nil, "", fmt.Errorf("ambiguous IDS: machine %q, container %q",
```
kklin: IDS => IDs

cli/command/counter.go
```diff
@@ -71,11 +72,11 @@ func queryCounters(c client.Client, tgt string) ([]pb.Counter, error) {
 		return c.QueryCounters()
 	}
 
-	mach, err := getMachine(c, tgt)
+	_, ip, err := apiUtil.FuzzyLookup(c, tgt)
```
kklin: Should you check that it didn't match a container? If a user thinks that we do counter metrics on a container and  tries it, they might be misled because the command didn't fail.

api/util/util.go
```diff
@@ -47,16 +47,18 @@ func FuzzyLookup(client client.Client, name string) (interface{}, string, error)
 	return nil, "", fmt.Errorf("no machine with private IP %s", privateIP)
```
kklin: In the commit message: s/hostanme/hostname

api/util/util.go
```diff
@@ -69,8 +71,11 @@ func findContainer(containers []db.Container, blueprintID string) (db.Container,
 		return *choice, nil
 	}
 
-	err := fmt.Errorf("no container with blueprintID %q", blueprintID)
-	return db.Container{}, err
+	if c, ok := hostnameToContainer[id]; ok {
```
kklin: Do you think this change is worthy of being in the changelog?

api/util/util_test.go
```diff
@@ -69,8 +69,8 @@ func TestFindContainer(t *testing.T) {
 	assert.Equal(t, b, res)
 
 	_, err = findContainer([]db.Container{a, b}, "4")
-	assert.EqualError(t, err, `ambiguous blueprintIDs 4567 and 432`)
+	assert.EqualError(t, err, `ambiguous choices 4567 and 432`)
 
 	_, err = findContainer([]db.Container{a, b}, "1")
```
kklin: Can you add a test for matching by hostname?

api/util/util.go
ejj: sure

cli/command/counter.go
ejj: Meh.  I'm not even sure we should document counters.  Users shouldn't be using it as it's a debugging feature for us.

That said, easy to check so I can go ahead and add it.

PR #1244, opened 2017-11-08
foreman: Run a thread per minion

luise:
Before, the foreman used a fork-join logic, where the code was mostly
single threaded, but would create multiple threads when connecting to the minion.
This would cause the main thread to be slowed down by stragglers.
With this commit, the main foreman thread instead creates a thread per minion.

I will submit a separate patch in the next few days to have the foreman write the role and connection status directly to the database (to replace https://github.com/kelda/kelda/pull/1228).


kklin:
@aegamesi I'll review this since we need to merge it pretty quickly

cloud/foreman/foreman_test.go
```diff
@@ -15,163 +15,208 @@ type clients struct {
...
+	assert.Contains(t, clients.clients, "2.2.2.2")
 
-	conn.Txn(db.AllTables...).Run(func(view db.Database) error {
+	machines = []db.Machine{{PublicIP: "2.2.2.2", CloudID: "ID2"}}
+	expectStop := minionChans["1.1.1.1"]
+	updateMinions(conn, machines, minionChans)
+	assert.Equal(t, 2, clients.newCalls)
+	assert.NotContains(t, minionChans, "1.1.1.1")
+	assert.Contains(t, minionChans, "2.2.2.2")
+	_, more := <-expectStop
```
luise: This might be a bad way of testing that the thread received the stop signal, since if it isn't true, we'll just block forever.  I wasn't sure what would be a better way, but let me know if you have one!

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,199 @@ type clientImpl struct {
...
-		log.WithField("machine", m.machine).Debug("New connection")
-	} else {
-		c.Inc("Minion Disconnected")
+func setMinionStatus(minionIP string, config pb.MinionConfig, isConnected bool) {
+	statusLock.Lock()
+	minionStatuses[minionIP] = minionStatus{
+		connected: isConnected,
+		role:      db.PBToRole(config.Role),
+	}
+	statusLock.Unlock()
```
ejj: It's marginally better style to do `defer statusLock.Unlock()` immediately after the lock call.  Doesn't matter much, but its slightly more future proof.

luise: Cool, I'll do that

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,199 @@ type clientImpl struct {
...
+var c = counter.New("Foreman")
 
-	var blueprint string
+// Run checks for updates to the machine and blueprint tables, and starts and
+// stops minion threads in response.
+func Run(conn db.Conn, creds connection.Credentials) {
+	credentials = creds
+	// A map from public IP to the stop channel for the corresponding minion thread.
+	minionChans := make(map[string]chan struct{})
 	var machines []db.Machine
```
ejj: This could be defined inside the loop.

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,199 @@ type clientImpl struct {
...
-	conn.Txn(db.BlueprintTable,
-		db.MachineTable).Run(func(view db.Database) error {
 
-		machines = view.SelectFromMachine(func(m db.Machine) bool {
-			return m.PublicIP != "" && m.PrivateIP != ""
+	for range conn.TriggerTick(60, db.BlueprintTable, db.MachineTable).C {
```
ejj: You don't seem to access the BlueprintTable, so no need to trigger on it.

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,199 @@ type clientImpl struct {
...
+		conn.Txn(db.MachineTable).Run(func(view db.Database) error {
```
ejj: You can just use `conn.SelectFromMachine`

luise: Ah, that's much nicer.

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,199 @@ type clientImpl struct {
...
-	// not alter the configuration because of temporary connectivity issues.
-	minionIPToPublicKey := map[string]string{}
-	var etcdIPs []string
+	seen := make(map[string]struct{})
 	for _, m := range machines {
-		if m.Role == db.Master && m.PrivateIP != "" {
-			etcdIPs = append(etcdIPs, m.PrivateIP)
+		_, ok := minionChans[m.PublicIP]
+		if !ok {
+			cli, err := newClient(m.PublicIP)
```
ejj: I really think creation of the client should happen inside the newMinion thread.  That's in large part the expensive bit that we're trying to parallelize.

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,199 @@ type clientImpl struct {
...
-			Provider:            string(m.machine.Provider),
-			Size:                m.machine.Size,
-			Region:              m.machine.Region,
-			EtcdMembers:         etcdIPs,
-			AuthorizedKeys:      m.machine.SSHKeys,
-			MinionIPToPublicKey: minionIPToPublicKey,
-		}
+func newMinionImpl(conn db.Conn, minionIP string, cli client, stop chan struct{}) {
+	log.Debugf("Start minion at %s", minionIP)
+	trigger := conn.TriggerTick(60, db.BlueprintTable, db.MachineTable)
```
ejj: This doesn't feel frequent enough to me.  In the case where we're still trying to connect to the minion, I think we could go as quickly as once every 5 seconds or so.

Really, I think what we want, is to check more frequently for nodes that aren't connected.  Shouldn't be hard to add, but it could wait for later as well.  Up to you.

cloud/foreman/foreman.go
luise: Yeah, I thought about that, but it would just make the code a good deal. I will take a look at it though. I see that it makes sense to parallelize that part as well.

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,199 @@ type clientImpl struct {
...
-	return db.None
 }
 
-// IsConnected returns whether the foreman is connected to the minion running
-// on the machine with the given cloud ID.
-func IsConnected(cloudID string) bool {
-	min, ok := minions[cloudID]
-	return ok && min.connected
-}
+func runOnce(conn db.Conn, minionIP string, cli client) {
```
ejj: So I think there's a fairly high level problem that needs some work with this function.  Since the connection isn't created by this go routine, we don't have a way to deal with failed connections.  I.E. suppose we have a connection that's created happily, and then for some reason it times out later (you time our or something), I think with this code, getMinion and setMinion will fail basically forever and never recover, because we don't try to reconnect.

I think there's two ways to deal with this.  Not sure which is preferable (perhaps should discuss in person along with kevin).

1) runOnce could store the connection object in the minions stack somewhere.  If it's not there, it could try to create it before doing anything else.  And if any of the connection methods error, it could remove it so that the next run through runOnce sets it up.

2) Each run through runOnce could create a new connection.  Basically this means we have to do an extra round trip, but that doesn't seem so painful to me.  Certainly the simpler option.

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,199 @@ type clientImpl struct {
...
-		if m.connected {
-			log.WithError(err).Error("Failed to get minion config")
-		} else {
-			log.WithError(err).Debug("Failed to get minion config")
+func makeConfig(conn db.Conn, minionIP string) pb.MinionConfig {
+	var blueprint string
+	var machines []db.Machine
+	conn.Txn(db.BlueprintTable, db.MachineTable).Run(func(view db.Database) error {
+		bp, _ := view.GetBlueprint()
+		blueprint = bp.Blueprint.String()
```
ejj: No need to do this in the transaction.

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,199 @@ type clientImpl struct {
...
+		})
+		return nil
+	})
+
+	minionIPToPublicKey := map[string]string{}
+	var etcdIPs []string
+	var minionMachine db.Machine
+	for _, m := range machines {
+		if m.PublicIP == minionIP {
+			minionMachine = m
```
ejj: Probably should handle the case where minionMachine isn't there.

cloud/foreman/foreman.go
ejj: Yeah.  See my comment below though, I think you really have to do it to be correct.

Ethan

cloud/foreman/foreman.go
ejj: (Btw that will happen when namespaces change for example).

cloud/cloud_test.go
```diff
@@ -76,7 +76,7 @@ func (p *fakeProvider) Boot(bootSet []db.Machine) ([]string, error) {
 		// We simulate this by setting the role of the machine returned by
 		// `List()` to be None, and only return the correct role in
 		// `getMachineRole`.
-		p.roles[toBoot.CloudID] = toBoot.Role
+		p.roles[toBoot.PublicIP] = toBoot.Role
```
kklin: Is it safe now to go back to using public IPs for identifiers? I fixed a bug that resulted from that https://github.com/kelda/kelda/commit/9542c492

ejj:
Yeah you're probably right.  Should likely use the CloudID.

Btw, luise, you should make sure to test how your code deals with floating IP changes.  That's probably another argument for creating the connection each time rather than trying to preserve it across run loops.  If the IP changes, its not clear how to deal with that.

cloud/foreman/foreman.go
luise: There is currently only a `GetBlueprint` function on  a `Database` instance, not on a `Conn`. But I'll just go ahead and create the function in a separate commit since it would simplify things.

cloud/cloud_test.go
luise: Ah, good point! I didn't see that commit. I'll use cloudID instead!
And @ejj , if the IP of a machine changes, it's just treated like a dead machine and a new machine (so a new thread and client will be created).  But I'll definitely fix the IP->CloudID!

cloud/foreman/foreman.go
luise: Hmm, yeah, I'm not quite sure how much to do about it. Right now we just try to send a mostly empty config to the machine that is potentially no longer in the database, which probably isn't that bad (or is it?).

The `minionMachine` will also not exist when the machine is `stopping`. So we shouldn't treat the missing machine as an error (that would lead to a bunch of errors). We could log it at debug and otherwise leave it as is?

ejj: Actually I think sending an empty config to the machine is really bad.  If the machine is a master we're in-effect, telling it to delete all containers.  That would be ok if we're stopping the entire cluster, but on a namespace change that could be quite problematic.

I agree logging isn't the correct behavior.  I think the right thing to do is just exit runOnce() and wait until it's called again.  By then either the machine should be back, or the stop channel should have been closed.

cloud/cloud_test.go
ejj: Probably makes sense to do so everywhere btw.  I.e. have a go routine per cloudID instead of a go-routine per public IP.  I don't see a downside and it's likely a bit more robust.

luise: Yeah, I agree, I'm switching everything over.

cloud/foreman/foreman.go
luise: Ah, that makes sense, thanks! I'll exit runOnce

luise:
@ejj  and @kklin, thanks for your comments. Some of them made me realize that I had definitely overlooked some complexity around changing IPs and cloud IDs -- and there are indeed some issues when reassigning floating IPs (also in the latest version I pushed). I'll take a look at it again tomorrow morning, and maybe ask for some help debugging if either of you is free.

cloud/foreman/foreman.go
```diff
@@ -37,189 +32,223 @@ type clientImpl struct {
...
+// Run checks for updates to the machine table and starts and stops minion threads
+// in response.
+func Run(conn db.Conn, creds connection.Credentials) {
+	credentials = creds
+	// A map from cloud ID to the stop channel for the corresponding minion thread.
+	minionChans := make(map[string]chan struct{})
 
-	conn.Txn(db.MachineTable).Run(func(view db.Database) error {
-		machines := view.SelectFromMachine(func(m db.Machine) bool {
+	for range conn.TriggerTick(60, db.MachineTable).C {
```
ejj: More I think about it, I don't think you need the tick.  A regular trigger should be fine.  There's no need to check these regularly.

cloud/foreman/foreman.go
```diff
@@ -37,189 +32,223 @@ type clientImpl struct {
...
+		seen[m.CloudID] = struct{}{}
 	}
 
-	// Assign all of the minions their new configs
-	forEachMinion(func(m *minion) {
-		if !m.connected {
-			return
+	for id := range minionChans {
+		if _, ok := seen[id]; !ok {
+			log.Debugf("Stop minion at %s", id)
```
ejj: The message could use a bit more context.  I.E. it's not clear you're referring to a foreman thread.  That said, I'm not sure the message is necessary.

cloud/foreman/foreman.go
```diff
@@ -37,189 +32,223 @@ type clientImpl struct {
...
-			Region:              m.machine.Region,
-			EtcdMembers:         etcdIPs,
-			AuthorizedKeys:      m.machine.SSHKeys,
-			MinionIPToPublicKey: minionIPToPublicKey,
+var newMinion = newMinionImpl
+
+func newMinionImpl(conn db.Conn, cloudID string, stop chan struct{}) {
+	log.Debugf("Start minion at %s", cloudID)
+	// Threads that aren't currently connected to their minion should run more often.
+	frequentTrigger := conn.TriggerTick(5, db.BlueprintTable, db.MachineTable)
```
ejj: Just use a time.Ticker.  No need for both of these to be triggers.

cloud/foreman/foreman.go
```diff
@@ -37,189 +32,223 @@ type clientImpl struct {
...
+		default:
 		}
 
-		if err := m.client.setMinion(newConfig); err != nil {
+		currConfig, connected := runOnce(conn, cloudID)
+		setMinionStatus(cloudID, currConfig, connected)
+	}
+}
+
+func runOnce(conn db.Conn, cloudID string) (currConfig pb.MinionConfig, connected bool) {
```
ejj: I think you could reorganize this to be just function to be just a bit simpler.   Specifically, I don't much like that we need to do two separate database transactions, and thus have two separate opportunities for the world to change out from under us as we're running.

What if you first thing you did a select to get the list of the machines.  Then Given that list of machines, you search through it to find the machine with my cloudID (which will give you the public IP).  And then later in makeConcifg, you can pass in the list of machines you grabbed earlier, and the machine that represents yourself.

ejj: That would also simplify makeConfig so that it doesn't need to return an error.  You can just deal with the "lack of self" error once at the top of runOnce()

cloud/foreman/foreman.go
```diff
@@ -37,189 +32,223 @@ type clientImpl struct {
...
+	}
+	return
 }
 
-// GetMachineRole uses the minion map to find the associated minion with
-// the cloudID, according to the foreman's last update cycle.
-func GetMachineRole(cloudID string) db.Role {
-	if min, ok := minions[cloudID]; ok {
-		return db.PBToRole(min.config.Role)
+func getPublicIP(conn db.Conn, cloudID string) (string, error) {
```
ejj: Also, when you make the reorganization I suggest, I don't think you need this function anymore.

cloud/foreman/foreman.go
```diff
@@ -37,189 +32,223 @@ type clientImpl struct {
...
+		// In a race between a closed stop and a trigger, choose stop.
+		select {
+		case <-stop:
 			return
+		default:
 		}
 
-		if err := m.client.setMinion(newConfig); err != nil {
+		currConfig, connected := runOnce(conn, cloudID)
+		setMinionStatus(cloudID, currConfig, connected)
```
ejj: Why not call this from within runOnce()?

cloud/foreman/foreman.go
luise: Hmm, I'll probably just delete it (and the same with the "starting" log).

cloud/foreman/foreman.go
luise: Ah, I like that. Thanks!

cloud/foreman/foreman.go
luise: Because then I'd have to call it before each return (which would be a total of 5 extra times). Would you prefer that?

cloud/foreman/foreman.go
```diff
@@ -37,189 +31,202 @@ type clientImpl struct {
...
-			if err != nil {
-				continue
-			}
-			min = &minion{client: client}
-			minions[m.CloudID] = min
+		if m.CloudID == cloudID {
+			minionMachine = m
 		}
+	}
+	if reflect.DeepEqual(minionMachine, db.Machine{}) {
```
ejj: I don't really like this.  Just make minionMachine a pointer and check for nil?  Or have a bool "found"?  Generally best to avoid superfluous uses of reflection.

luise: Oh, cool, didn't know that reflection should be avoided. I'll fix it

ejj:
LGTM.  Just needs a rebase.

luise:
@ejj , updated

PR #1245, opened 2017-11-08
etcd: Persist data to host disk

kklin:
Before, etcd data was saved to the container's disk, but not persisted in
the host. Therefore, if the etcd container was restarted, all the
information in etcd was lost. This resulted in a bug where if a new etcd
configuration was sent to the minions (for example, right after the daemon
restarted, or the number of masters was changed in the blueprint), all
containers would be restarted. This is because sending a new etcd config
triggers the etcd container to restart, which in turns causes the minions
to read that there are no scheduled containers, resulting in it destroying
whatever containers it was previously running.


integration-tester/tests/101-switch-namespace/switch_namespace_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+		expDbc.ID = 0
+		actualDbc.ID = 0
+
+		if reflect.DeepEqual(expDbc, actualDbc) {
+			return 0
+		}
+		return -1
+	}
+	_, extraContainers, missingContainers := join.Join(
```
ejj: May be simpler just to convert initialContainers and currentContainers to maps (setting ID to zero), and then done an assert.Equal().  You'll get a better error message too if you do so.

kklin:
Hm, something about the commit that creates a thread per minion breaks this PR. I'm going to hold off on merging until I get the integration test passing.

PR #1247, opened 2017-11-08
bindings: Add `t2.micro` instance to AWS descriptions

HongjianZhang:
This patch manually adds `t2.micro` to AWS descriptions in order
to pass integration test. It also modifies `chooseBestSize`, so
that it does not choose non-preemptible instance sizes when
preemptible flag is set.


kklin:
@HongjianZhang looks like some unit tests for the size logic failed because they now make use of the `t2.micro` instance

js/bindings/bindingsTest.js
```diff
@@ -194,6 +194,18 @@ describe('Bindings', () => {
       infra = new b.Infrastructure(machine, machine);
       checkMachines([{
         provider: 'Amazon',
+        size: 't2.micro',
+      }]);
+    });
+    it('ignores non-preemtiple machines when preemptible flag is set', () => {
```
kklin: preemtiple => preemptible

js/bindings/bindings.js
```diff
@@ -810,11 +811,14 @@ class Machine {
     let bestPrice = Infinity;
     for (let i = 0; i < providerDescriptions.length; i += 1) {
       const description = providerDescriptions[i];
-      if (ram.inRange(description.RAM) &&
-          cpu.inRange(description.CPU) &&
-          (bestSize === '' || description.Price < bestPrice)) {
-        bestSize = description.Size;
-        bestPrice = description.Price;
+      if (!this.preemptible ||
```
kklin: This if block is getting pretty hard to read. I think it'd be easier to follow if we broke up the `if` statement:

```
const nonPreemptible = nonPreemptibleSizes[this.provider].includes(description.Size);
if (this.preemptible && nonPreemptible) {
  continue
}

const isValid = ram.inRange(description.RAM) && cpu.inRange(description.CPU);
if (!isValid) {
  continue
}

if (bestSize === '' || description.Price < bestPrice) {
  ....
}
```

I also think checking whether the size is valid given the preemptible flag with `!(this.preemptible && nonPreemptible)` rather than `(!this.preemptible || !nonPreemptible)` is easier to read since it doesn't rely on the shortcircuiting of the first case (`!this.preemptible`), and there's less negating.

HongjianZhang: Yeah I tried to use `continue` statement, but got a lint error saying `Unexpected use of continue statement  no-continue`, so I got rid the `continue`. Am I misinterpreting the lint error?

kklin: Eh, I think disallowing it is dumb. I like them. I'll post and see if people are OK with us allowing it in our code base.

HongjianZhang: I agree.

kklin: I think you can go ahead and add a commit to allow it in the .eslintrc.json

PR #1248, opened 2017-11-09
tester: Fix name for Spark master container

kklin:
The Spark master hostname was recently renamed from "spark-ms" to
"spark-master". This commit also makes use of the new feature to `kelda
ssh` given a hostname rather than blueprint ID.


kklin:
No worries these tests are super fragile!

PR #1253, opened 2017-11-09
api: Query cluster information from cache

kklin:
Before, every call to `kelda show` would cause the daemon to query all
machines for the state of the containers. This was inefficient, especially
at scale. With this commit, querying the cluster is separated from
retrieving the information for `kelda show` -- the daemon constantly
queries the cluster, and `kelda show` simply displays the most recent
information (which should be no more than 10 seconds old).

Another benefit of this change is that temporary issues to connect to the
cluster are hidden from the user. `kelda show` will simply display the last
successful call, and any errors communicating with the cluster is logged in
the daemon.


api/server/server.go
```diff
@@ -152,36 +171,98 @@ func (s server) queryLocal(table db.TableType) (interface{}, error) {
...
+	default:
+		return nil, fmt.Errorf("unrecognized table: %s", table)
+	}
+}
+
+// syncClusterInfo periodically connects to the cluster and writes the state
+// of the cluster into the `clusterInfo` map. It tracks the Container, LoadBalancer,
+// Image, and Connection tables.
+func (s *server) syncClusterInfo() {
+	for range time.Tick(10 * time.Second) {
```
ejj: For me personally, I think 5 seconds would be better.  `kelda stop -containers` will feel more responsive that way.  Did you do any testing that suggested it was too expensive to query more frequently?

kklin: 5 seconds should be fine

api/server/server.go
```diff
@@ -152,36 +171,98 @@ func (s server) queryLocal(table db.TableType) (interface{}, error) {
...
+	if len(machines) == 0 {
+		return
 	}
 
-	var leaderClient client.Client
-	leaderClient, err := newLeaderClient(s.conn.SelectFromMachine(nil), s.clientCreds)
+	leaderClient, err := newLeaderClient(machines, s.clientCreds)
 	if err != nil {
-		return nil, err
+		log.WithError(err).Warn("Failed to connect to leader")
```
ejj: Is this warning going to be annoying while the cluster is booting?

kklin: I actually fix this in the next commit

api/server/server.go
```diff
@@ -152,36 +171,98 @@ func (s server) queryLocal(table db.TableType) (interface{}, error) {
...
+		},
+		{
+			db.LoadBalancerTable,
+			func(c client.Client) (interface{}, error) {
+				return c.QueryLoadBalancers()
+			},
+		},
+	}
+
+	wg := sync.WaitGroup{}
```
ejj: nit, but I usually just do `var wg sync.WaitGroup`

api/server/server.go
ejj: nvm, just saw the next patch.

PR #1255, opened 2017-11-10
tester: Remove non-determinism from runtime-info test

kklin:
Before, the runtime info test picked a random hash to use as its
index.html at runtime. This resulted in the blueprint producing
different containers each time it was run -- this confused the
integration test runner since it checks whether the expected containers
had been booted by re-compiling the blueprint (which would result in
a different container than what was actually deployed, causing the test
to fail).


integration-tester/tests/runtime-info/runtimeInfo.js
```diff
@@ -10,8 +9,10 @@ const webContainer = new kelda.Container('web', 'nginx', {
   env: {
     myPubIP: kelda.hostIP,
   },
+  // Serve a random hash so that the test can ensure it really reached this
```
kayousterhout: Can you add a comment about why this should be deterministic? (from the commit message)

PR #1256, opened 2017-11-10
util: Fix backoff wait logic

kklin:
The previous code didn't actually make use of `cap` so it increased the
wait interval indefinitely..

Sigh.. I wrote the original patch, and I'm not sure what happened 🙈


ejj:
Since Kevin is away I'm just going to merge this ...

PR #1259, opened 2017-11-10
docs: Change the Getting Started guide to use Node/Mongo

luise:
This commit updates the Getting started guide to deploy a Node/MongoDB
application rather than the very simple Nginx website. Since Node/Mongo is
typically more complex to boot than Nginx, this new guide should better reflect
the benefits of Kelda.

As part of the rewrite, a lot of the explanations have been stripped away.
The guide now has clearer and simpler step-by-step instructions, which should
make it easier for users to follow it and not miss anything important.
Users who want to know more about how Kelda works should read the How Kelda
Works section.

This PR should be merged with the following PRs:
* https://github.com/kelda/nginx/pull/15
* https://github.com/kelda/nodejs/pull/10
* https://github.com/kelda/node-todo/pull/1
* https://github.com/kelda/mean/pull/16


README.md
```diff
@@ -32,7 +32,7 @@ Install Kelda with npm:
 ```bash
```
kklin: In the commit message, "Started" in "Getting started" should be capitalized

docs/_GettingStarted.md
```diff
@@ -44,199 +43,142 @@ tutorial before running `kelda init`**. In the next step, you will run
...
+3. **Check the status**. The VMs and application containers are now booting.
+Use Kelda's `show` command, to see how things are progressing.
+The output looks similar to this:
 
-The output above means that Kelda has launched two machines, one as a master and
-one as a worker, in Amazon.  Both machines are `disconnected`, because they're
-still being initialized. When a machine is fully booted and configured, it will
-be marked as `connected` in the `STATUS` column.
+    ```console
+    $ kelda show
```
kklin: I think this `kelda show` output is out of date? The machine IDs shouldn't be blueprint IDs, and "disconnected" isn't printed as a status anymore. And I don't think the Role will show up now before the machine is connected.

kklin: It's minor but I think there's some chance users will get confused and think something is wrong if their output is sufficiently different from this

kklin: I would also make sure that the machine IDs match with the machine IDs in the output below that has containers

docs/_GettingStarted.md
```diff
@@ -44,199 +43,142 @@ tutorial before running `kelda init`**. In the next step, you will run
...
-will prompt you to accept the changes that you're making to your deployment;
-type `y`.
-
-Run `kelda show` again and notice that Kelda has stopped the old
-container and is starting a new one.  When the new container is `running`,
-navigate to the new IP address and check that the modified page is up.
+Note that the commands in this section (and all Kelda commands that take IDs)
+don't need the full ID; a unique prefix of the ID is enough.
+
+1. **Check the logs**. To get the logs of the Node.js app, find the
```
kklin: I think these step by step instructions for debugging are much easier to follow!

docs/_GettingStarted.md
```diff
@@ -44,199 +43,142 @@ tutorial before running `kelda init`**. In the next step, you will run
...
-```console
-$ kelda stop
-```
-
-At this point, you can safely kill the Kelda daemon; go to the terminal window
-that's running `kelda daemon` and stop the process with `Ctrl+C`, or just close
-the window.
+1. **Stop the VMs**. When you're done playing around, __make sure to stop the machines
+you've started!__ Otherwise, your cloud provider might charge you for the
+VMs that are still running.  To stop all VMs and containers:
```
kklin: Should we add a note about how to tell if the VMs are actually stopped? (when `kelda show` doesn't show the machines anymore)

docs/_GettingStarted.md
luise: Oh, good catch! I'll update it to be the newest output, and match with the below output.

docs/_GettingStarted.md
luise: I'm glad!

docs/_GettingStarted.md
luise: Good point!

luise:
@kklin , I addressed your comments and also renamed the file (so the diff isn't clear anymore).

PR #1260, opened 2017-11-10
Expose VM properties to the user

kayousterhout:
The point of this PR is to expose information about the memory and CPUs on each worker so that blueprints can use it (see https://github.com/kelda/spark/pull/22 for an example).  There's one other commit here to fix an existing problem that I uncovered in the process of adding this (the other issue needed to be fixed to make the unit tests with the new change).


js/bindings/bindings.js
```diff
@@ -700,12 +700,12 @@ class Machine {
    *   size: '512mb',
    * });
    *
-   * @param {Object.<string, string>} [opts] - Optional arguments that
-   *   modify the machine.
-   * @param {string} [opts.provider] - The cloud provider that the machine
+   * @param {Object.<string, string>} opts - Arguments that
```
kklin: Nit - could "modify" be on the same line as "that"? The line break looks a little funny

js/bindings/bindings.js
```diff
@@ -866,15 +881,15 @@ class Machine {
         continue;
       }
 
-      if (bestSize === '' || description.Price < bestPrice) {
-        bestSize = description.Size;
+      if (bestMachine === undefined || description.Price < bestPrice) {
+        bestMachine = description;
         bestPrice = description.Price;
```
kklin: I have a small preference to get rid of the `bestPrice` variable and reference `bestMachine.Price`, what do you think?

js/bindings/bindings.js
```diff
@@ -114,8 +114,13 @@ class Infrastructure {
       containers.push(c.toKeldaRepresentation());
     });
 
+    const machines = [];
+    this.machines.forEach((m) => {
```
kklin: Might be cleaner to express this as `this.machines.map`? Don't feel strongly though

kayousterhout: Oh nice I didn't realize JS had a map!

js/bindings/bindings.js
kayousterhout: Oh that's much nicer - thanks!

PR #1262, opened 2017-11-11
Only allow one base infrastructure

luise:
It wasn't clear why users would create multiple base infrastructures, and the
ability to do so added complexity to both the docs and init command. With this
PR, users can only have a single base infrastructure at a time.


ejj:
yay!

js/bindings/bindings.js
```diff
@@ -194,26 +183,18 @@ function getBaseInfrastructure(infraPath) {
...
-  const infraPath = getInfraPath(name);
-  if (!fs.existsSync(infraPath)) {
-    throw new Error(`no infrastructure called ${name}. Use 'kelda init' ` +
-      'to create a new infrastructure.');
+function baseInfrastructure() {
+  if (!fs.existsSync(baseInfraLocation)) {
+    throw new Error('no base infrastructure. Use `kelda init` to create one.');
   }
-  return getBaseInfrastructure(infraPath);
+  return getBaseInfrastructure();
```
kklin: I don't think having this separate is very useful -- maybe we should just move the contents of that function into this function?

js/initializer/prompter.js
```diff
@@ -355,31 +328,31 @@ function machineCountPrompts() {
...
-          return machineCountPrompts();
-        })
-
-        .then((machineAnswers) => {
-          Object.assign(answers, machineAnswers);
-          return { provider, answers };
+  return overwritePrompt()
+    .then((shouldOverwrite) => {
+      if (shouldOverwrite[consts.infraOverwrite] === false) {
+        return new Promise(resolve => resolve(null));
```
kklin: The relationship between this and `run()` took me awhile to understand. I have two suggestions:
1. Do we have to return a `Promise`? It looks like on line 354 you're just returning an object directly?
2. Can we set an explicit flag rather than returning null? Then the `run()` function would read more easily I think. So, this would be `return { shouldAbort: true }`, and in `run()` the test would read `if (result.shouldAbort)` which is clearer IMO.

js/bindings/bindings.js
luise: Did you see the comment for the function? It's a separate function because it simplifies testing to be able to mock it out.

kklin: Ah I missed that, sorry! Sounds fine then

js/initializer/prompter.js
luise: Ah, yeah, I like both points. I'll update, thanks!

luise:
@kklin , addressed your comments!

kklin:
Oh, if you think that this deserves to be in the changelog, can you add an entry there as well?

luise:
Good point, @kklin. I just updated the changelog.

PR #1263, opened 2017-11-11
minion: Build custom Docker images in parallel

aegamesi:
Previously, custom Docker images would be built sequentially by the
master, potentially not making full use of the machine. This change
allows up to 8 images to be built simultaneously.


minion/registry/registry.go
```diff
@@ -46,29 +47,54 @@ func syncImages(conn db.Conn, dk docker.Client) {
...
+	}
 
-		img.DockerID = id
-		img.Status = db.Built
+	maxBuilders := 8
+	if len(toBuild) < maxBuilders {
+		maxBuilders = len(toBuild)
+	}
+	images := make(chan db.Image)
+	for i := 0; i < maxBuilders; i++ {
```
ejj: Two broad points on this PR.  1) Why limit the number of builders?  Does some sort of error occur if too many happen at once?

2) Assuming that you need to limit the number of builders, the concurrency code would be a lot simpler if you used a sempahore (http://www.golangpatterns.info/concurrency/semaphores).  I.e. you can loop over the images, make a go routine per image, and then use a semaphore to limit the number of active go routines to whatever you'd like using the semaphore.  This approach would allow you to avoid the complexity of creating all of these builder threads and sending images to them

aegamesi: I'm not sure what the common use case is for custom Docker images is, but if you have too many CPU or RAM bound Dockerfile builds at once, the system slowly grinds to a halt and breaks. I was testing with a very lightweight Dockerfile, but as I increased the number of parallel builds I did see those issues. It seems like 8 is a high enough number that if you have a reasonable number of different images that need to be built, they'll all be parallelized, but low enough that if you have an absurd number of different images that need to be built, the parallelism is limited enough that you don't break the system.

aegamesi: 2. Oh yeah, I didn't realize you could use channels as semaphores. This does simplify the code quite a bit.

minion/registry/registry.go
```diff
@@ -46,29 +47,46 @@ func syncImages(conn db.Conn, dk docker.Client) {
 	var toBuild []db.Image
 	conn.Txn(db.ImageTable).Run(func(view db.Database) error {
 		toBuild = view.SelectFromImage(func(img db.Image) bool {
-			return img.Status != db.Built
+			return !img.BuildQueued
```
ejj: I don't quite see the point of this BuildQueued thing. Nothing uses it outside of this function ...  Also, the way you've written it here, it won't work.  You can only write to an image in a database transaction.

aegamesi: The BuildQueued marks images that *will* be built, but are not yet building. Since syncImages may be called multiple times, this prevents several goroutines from being spawned in separate invocations of syncImages that all attempt to build the same image.

The new image is being saved to the database below: https://github.com/aegamesi/kelda/blob/24da744d2c880c1d457a2344753534e4bbc90dca/minion/registry/registry.go#L80

aegamesi:
@ejj

integration-tester/tests/build-dockerfile/custom_images_test.go
```diff
@@ -62,6 +63,11 @@ func TestCustomImages(t *testing.T) {
 	if countErr != nil {
 		t.Error(countErr)
 	}
+
+	parallelErr := checkBuildParallelized(machines, containers)
```
ejj: I would just call this err I think, it's more idiomatic with the rest of the code.  Also, if you have space, I'd pull the if statement into this line.

PR #1264, opened 2017-11-13
digitalocean: Refactor SetACL to join godo.InboundRules

kklin:


cloud/digitalocean/digitalocean.go
```diff
@@ -425,9 +413,6 @@ func toRules(acls []acl.ACL) (rules []godo.InboundRule) {
 			portRange := fmt.Sprintf("%d-%d", acl.MinPort, acl.MaxPort)
 			if acl.MinPort == acl.MaxPort {
 				portRange = fmt.Sprintf("%d", acl.MinPort)
-				if acl.MinPort == 0 {
-					portRange = "all"
-				}
```
aegamesi: Why is this removed? syncACLs converts the ACLs into InboundRules, but if the ACL specifies 'all' (with the range [0, 0]), won't the call to the DigitalOcean API end up trying to add a rule with port range "0" (which the DO API doesn't look)?

aegamesi: Ah, nevermind, I see -- I forgot that the only reason there would be an ACL with port range [0, 0] is if it were created by converting from a DO rule to a Kelda ACL -- and that situation wouldn't happen in this patch.

PR #1265, opened 2017-11-13
docs: Use new figure for How Kelda Works

luise:
The old diagram didn't really fit the visual style of our docs, and it
additionally said `quilt` instead of `kelda`.


luise:
@kayousterhout , changed the font size

kayousterhout:
One tiny nit here: do you think the dashed line to the cluster should come specifically from the daemon, rather than from the machine that's running both of them?

luise:
Hmm, I don't feel strongly. Happy to update it though, if you think that would be better

luise:
Went ahead and did it. I guess it is more accurate, so why not!

kayousterhout:
Cool -- over time I think we'll need to be more precise about what exactly the daemon is doing, although we've also had so many conversations about re-structuring the relationship between the daemon and the master that I don't think we should invest too much effort in clarifying this right now

PR #1266, opened 2017-11-14
docs: Update instructions for GCE credentials

kayousterhout:
The instructions for GCE credentials assumed that the user would
run `kelda init`. However, since the getting started guide
assumes AWS, it seems like the most likely path for users to reach
the GCE instructions will be when the user has already set up a
base infrastructure with a different provider. Based on that
assumption, this commit changes the instructions for how to configure
the GCE credentials to tell the user how to install the file themselves
rather than using `kelda init`.


luise:
The Quick Start doesn't actually assume that the user uses Amazon -- it just refers users to the cloud provider config section for their desired provider (so Google users will end up in the section you are editing in this commit).
If you think we need descriptions of what to do without init (which I agree might be useful), I would be in favor of having both instructions for init and for manual setup -- e.g., step 5 could say "There are two ways to install provider credentials in the location Kelda expects" and then have sub bullets 5a) Using kelda init, and 5b) Manually. Or something like that.
Regardless of which option you choose, I think it would be good to update DigitalOcean and AWS to have similar information, for consistency.

kayousterhout:
Ah I see I didn't realize the earlier quick start guide didn't have cloud provider setup instructions anymore.

For Amazon, it makes sense to me to recommend `init`, because it's annoying to format the file (looks like DO has a somewhat similar issue?) - but for GCE, since Google already gives you the file, moving the file seems just as easy to me (/ easier) than going through the whole `kelda init` command (since with `kelda init`, you still need to figure out the path to the downloaded file, and you're forced to do a bunch of other setup too, like decide a particular instance type).  Also, moving the file still works if you want to run `kelda init`, IIUC, because it will detect that the file is there and leave it be (so seems like things still work fine for the getting started guide, but makes things much easier for folks who don't need a base infra -- like me in this case).  Thoughts?

kayousterhout:
Actually maybe it would be better to update Digital Ocean too (since that format also looks super simple?) and then we don't need the bolded / scary instructions to "come back here before running `kelda init`".

luise:
Hmm, yeah, I think we should decide if we want kelda init to set credentials up at all then. I.e., it doesn't make sense to do, if users already did it manually. But then it seems silly to still ask for AWS -- or at least inconsistent. 
We've talked about splitting up the init command, anyways. Maybe we should just do that now and make a credentials setup command?

kayousterhout:
Ok I'll close this

PR #1267, opened 2017-11-14
changelog: Prepare for release 0.6.0

kklin:


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ Kelda Change Log
...
 
+Release 0.6.0
+-------------
+
+Release 0.6.0 simplifies the `kelda ssh` command and allows containers to reference
+their public IP.
+
 - The Kelda command line utility now accepts container names in addition to
 IDs.  For example, `kelda ssh spark-master` can be used instead of
 `kelda ssh 6cf081531346`.
```
luise: This is not from this commit, but in the below point, I think "its" should be "their" since "containers" is plural.

PR #1268, opened 2017-11-14
docs: Remove fake links from 'Developing the Minion'

aegamesi:
For some reason, the documentation previously had references to specific files
in Kelda be actual, clickable links. This was very misleading, because the
links wouldn't actually end up going anywhere and the documentation wouldn't
display the full path to the referenced file.

This commit fixes those links, and corrects some slightly incorrect Markdown
link syntax in another section.


docs/_Dev.md
```diff
@@ -364,13 +364,13 @@ and not the default Kelda minion image, follow these steps:
 
 1. Create a new empty repository on your favorite registry -
 [docker hub](https://hub.docker.com/) for example.
-2. Modify `keldaImage` in [cfg.go](../cloud/cfg/cfg.go) to
+2. Modify `keldaImage` in `/cloud/cfg/cfg.go` to
```
kayousterhout: I'd remove the leading '/' here since this is a relative path from the root directory

PR #1269, opened 2017-11-14
js: Throw an error when overwriting a previously defined Infrastructure

aegamesi:
Previously, if a user called `new Infrastructure(...)` twice, the second
call would silently overwrite the first infrastructure. Since that's
probably a mistake, this commit makes the Infrastructure constructor
explicitly throw an error informing the user of that fact.

---

I'm a bit iffy on the error message "an Infrastructure has already been created" -- I was also considering something like "another Infrastructure has already been defined", etc.
Suggestions are welcome :)


js/bindings/bindings.js
```diff
@@ -59,6 +59,9 @@ class Infrastructure {
     this.containers = new Set();
     this.loadBalancers = [];
 
+    if (_keldaInfrastructure !== undefined) {
+      throw new Error('an Infrastructure has already been created');
```
kayousterhout: I'd make this very very specific, to make it as obvious as possible to users how to resolve it.  How about something like "The Infrastructure constructor has already been called once (each Kelda blueprint can only define one Infrastructure)."

js/bindings/bindingsTest.js
```diff
@@ -13,11 +13,13 @@ const { expect } = chai;
...
-  beforeEach(() => {
-    b.resetGlobals();
+
+  const createBasicInfra = function createBasicInfra() {
     const machine = new b.Machine({ provider: 'Amazon' });
     infra = new b.Infrastructure(machine, machine);
-  });
+  };
+
+  beforeEach(b.resetGlobals);
```
kayousterhout: would it make sense to just call createBasicInfra here? Seems like not quite all of the tests need it - but enough that it would make sense to call it here rather than duplicating the beforeEach for many of the sub-tests

js/bindings/bindingsTest.js
```diff
@@ -1037,6 +1057,7 @@ describe('Bindings', () => {
       });
       expect(() => new b.Infrastructure(['not a Machine'], [machine]))
         .to.throw('not an array of Machines; item at index 0 ("not a Machine") is not a Machine');
+      b.resetGlobals();
```
kayousterhout: Does it make sense to just split this second part into its own test?

js/bindings/bindingsTest.js
aegamesi: Yeah -- that's how it was before, and that would probably be ideal. However, many (probably most) of the tests create their own Infrastructure within the test, and so if 'createBasicInfra' is called before *every* test, this new error will cause the majority of tests to fail (because they'll be creating an Infrastructure when one already exists). Kevin and I discussed this for a little bit and this seems to maybe be the cleanest way?

kayousterhout: Ohhhh yuck.  I see.  Ok I agree with you both!

js/bindings/bindings.js
aegamesi: Works for me!

js/bindings/bindingsTest.js
```diff
@@ -976,6 +986,16 @@ describe('Bindings', () => {
       infra = new b.Infrastructure([machine], [machine], { namespace });
       expect(infra.toKeldaRepresentation().namespace).to.equal(namespace);
     });
+    it('errors when it would overwrite a previously defined Infrastructure', () => {
```
kayousterhout: can you add "constructor" at the beginning to the string?

js/bindings/bindingsTest.js
aegamesi: Actually, I changed the constructor so that it doesn't set itself to be *the* Infrastructure if it encounters an error before it reaches the end of the constructor. That avoids the need for this line, and I think it's more correct anyway.

PR #1273, opened 2017-11-15
ssh: Demote most SSH key errors to DEBUG

kayousterhout:
Kelda tries many locations for SSH keys; for most of those locations,
we don't expect to find a key that Kelda can use. This commit message
demotes all log messages associated with SSH keys to debug unless
they're for the default Kelda ssh key path.


kayousterhout:
FYI @aegamesi I fixed this issue that you mentioned recently

cli/ssh/native.go
```diff
@@ -73,12 +73,21 @@ func defaultSigners() []ssh.Signer {
...
 		if err != nil {
-			if os.IsNotExist(err) {
+			// We look for SSH keys in many places; for most of these places,
+			// we don't typically expect to find an SSH key that Kelda can
+			// use, so log a debug message. We _do_ expect to, in most cases,
+			// find a key at cliPath.DefaultSSHKeyPath, because this is the
+			// path where Kelda automatically generates an SSH key, so log
+			// a warning if no key is found there.
+			if identityPath == cliPath.DefaultSSHKeyPath {
+				log.WithError(err).WithField("path", identityPath).Warn(
```
kayousterhout: @kklin is there a way to test log messages? Do we have any tests that mock this out to test? Do you think it would be worthwhile?

kklin: It's possible to test, although I don't think we do it anywhere. https://github.com/sirupsen/logrus#testing

I think it'd be reasonable to test this. It's not pressing or anything, but it might be worth putting in the time (like <10 mins) to set a precedent.

kayousterhout:
@kklin I updated this - would be helpful if you could sanity check the test since I'm still learning about Go and our unit testing style.

cli/ssh/native_test.go
```diff
@@ -54,6 +56,70 @@ func TestDefaultKeys(t *testing.T) {
...
+		if err := writeRandomKey(filepath.Join(sshDir, key), true); err != nil {
+			t.Errorf("Failed to write key: %q", err.Error())
+			return
+		}
+	}
+
+	signers := defaultSigners()
+	assert.Len(t, signers, 0)
+
+	defaultErrorLogged := false
```
kklin: I think this would be easier to read if the tests were in a table format:
```
tests := []struct{
  message string
  level logrus.Level 
}{
  {"Unable to load default identity file", logrus.WarnLevel},
  {"Unable to load identity file", logrus.DebugLevel},
  {"Key does not exist", logrus.DebugLevel},
}

Outer:
for _, test := range tests {
  for _, entry := range hook.AllEntries() {
    if test.message = entry.Message {
      assert.Equal(t, test.level, entry.Level)
      continue Outer
    }
  }
  t.Errorf("Failed to find log message: %s", test.message)
}
```

It also has the benefit of not failing if additional log messages are added.

cli/ssh/native_test.go
```diff
@@ -54,6 +56,70 @@ func TestDefaultKeys(t *testing.T) {
...
+	}
+
+	sshDir := filepath.Join(dir, ".ssh")
+	if err := util.AppFs.MkdirAll(sshDir, 0600); err != nil {
+		t.Errorf("Failed to create SSH directory: %q", err.Error())
+		return
+	}
+
+	// Write one password-protected key.
+	for _, key := range []string{"id_rsa"} {
```
kklin: Since you're the slice only has length 1, I think you can get rid of the loop?

cli/ssh/native_test.go
```diff
@@ -54,6 +56,70 @@ func TestDefaultKeys(t *testing.T) {
...
+	hook := logrusTest.NewGlobal()
+
+	util.AppFs = afero.NewMemMapFs()
+
+	// Don't pull in keys from the host OS. Setting this environment variable
+	// is safe because it won't affect the parent shell.
+	os.Setenv("SSH_AUTH_SOCK", "")
+
+	dir, err := homedir.Dir()
+	if err != nil {
```
kklin: I personally prefer to use `assert.NoError`, I think it's more readable:
```
assert.NoError(t, err, "failed to get homedir")
```

The one caveat is that `NoError` doesn't exit if the test fails. So if this fails to get the homedir, a `.ssh` folder will still be created. Usually I don't mind though, since if the test fails I don't really care what happens after. Or maybe there's a flag or something to make `NoError` exit.

cli/ssh/native_test.go
kklin: You could turn `hook.AllEntries` into a map rather than the double for loop if you wanted of course. Would probably be cleaner

kayousterhout: I left this as-is because, as discussed offline, the problem with the map is that it may obscure two messages that are the same but at different log levels, and the problem with the table is that it makes it harder to check that there aren't any _extra_ messages as > debug level

cli/ssh/native_test.go
kayousterhout: I added a new commit before mine to update this in the other tests in this file too (I'd just copied this test from another one)

cli/ssh/native_test.go
kayousterhout: Derp thanks

PR #1274, opened 2017-11-15
docs: Add network debugging instructions

kayousterhout:
This commit adds a new "How To" section that describes one
useful way to debug network connectivity problems in a
blueprint.


docs/_HowTo.md
```diff
@@ -292,3 +292,119 @@ onto the filesystem.
...
+# lsof -i -P -n
+COMMAND PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
+java     47 root  115u  IPv4 117041      0t0  TCP 10.181.11.2:60088->10.32.155.142:5432 (SYN_SENT)
+```
+
+In this case, the fact that the destination IP address begins with 10 signifies
+that the destination is another container in the deployment, because IP
+addresses beginning with 10 are private IP addresses. This problem is somewhat
+harder to debug because `kelda show` doesn't show each container's private IP
+address, so you may need to log in to the cloud provider API (e.g., the AWS API)
```
kklin: Won't the IP be of a container, and not a machine? (so the IP address won't show up in the AWS console?)

docs/_HowTo.md
```diff
@@ -292,3 +292,119 @@ onto the filesystem.
...
+postgresContainer.allowFrom(buggyContainer, 5432);
+```
+
+If you're curious, `lsof` lists all open files, which includes network
+connections because Linux treats network connections as file handles.  The `-i`
+argument tells `lsof` to only show IP files -- i.e., to only show network
+connections.  The `-P` argument specifies to show port numbers rather than port
+names, which is more useful here because Kelda relies on port numbers, not
+names, to enable network connections.  Finally, the `-n` argument specifies to
+show IP addresses rather than hostnames (the hostnames output by `lsof` are
```
kklin: I suspect we could get this to work by properly setting the container's hostname when we boot them. I bet the hostname you saw was `<dockerID>.kelda`?

This is probably worth doing because the `HOSTNAME` environment variable is wrong too. You can see this if you ssh into a container:
```
# echo $HOSTNAME
e0db08383164
```
I believe one of Michael's undergrads has reported that there was an application that relies on publishing `$HOSTNAME` into some database so that other nodes could find it. I wrote a fix for him, but never got it merged into our codebase.

Thanks for writing this btw! I'm sure it'll be really useful, and I didn't know about it before either!

kayousterhout: Ah interesting the hostname I was getting was $HOSTNAME (it didn't include .kelda -- e.g., the one I got was f54f90b9ce81).

docs/_HowTo.md
kayousterhout: Yeah I think you're right (and I did not check this!) - the hostname looked so random I incorrectly assumed it was AWS-assigned.  I'll fix this.

docs/_HowTo.md
kklin: So I spent more time looking into this than I should have.

I think the IP is being resolved into a hostname using reverse DNS. There's two problems here:
- The first is that our DNS server doesn't respond to PTR requests. This is easy to fix though.
- The other problem is that the way DNS works in Docker is that Docker has an "embedded DNS server" that first tries to respond to any queries that it knows about -- if that fails, it passes the query onto the user supplied server (i.e. our DNS implementation). The problem is that if the container IP that's being resolved is on the same host as the container requesting it, Docker will think it knows its DNS name because we give Docker the container IPs when we start them. Therefore, Docker responds with the container's Docker ID (which is what's shown in `lsof`), and our DNS server never gets a chance to send a different response. You can't turn off the embedded DNS server as of now.

tldr reverse DNS doesn't work in Kelda, and it'd be non-trivial to get it working. I don't know how common it is for applications to use reverse DNS, but I do know that kerberos uses it (although it can be turned off).

docs/_HowTo.md
```diff
@@ -292,3 +292,127 @@ onto the filesystem.
...
+COMMAND PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
+java      8 root   34u  IPv4  88572      0t0  TCP 10.14.9.118:56902->151.101.41.128:443 (SYN_SENT)
+```
+
+The useful output is the `NAME` column, which shows the source network address
+(in this case, port 56902 on address `10.14.9.118`) and the destination address
+(in this case, port 443 on address `151.101.41.128`). The `COMMAND` column may
+also be useful; in this case, it shows that the network connection was initiated
+by a Java process.  `SYN_SENT` means that the container tried to initiate a
+connection, but the machine at 151.101.41.128 never replied (in this case
```
kklin: Put `151.101.41.128` in backticks for consistency with the other IPs

docs/_HowTo.md
```diff
@@ -292,3 +292,127 @@ onto the filesystem.
...
+COMMAND PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
+java     47 root  115u  IPv4 117041      0t0  TCP 10.181.11.2:60088->10.32.155.142:5432 (SYN_SENT)
+```
+
+In this case, the fact that the destination IP address begins with 10 signifies
+that the destination is another container in the deployment, because IP
+addresses beginning with 10 are private IP addresses (in the context of Kelda,
+these are typically container IP addresses). This problem is somewhat
+harder to debug because `kelda show` doesn't show each container's private IP
+address, but it's often possible to determine which other container
```
kklin: FWIW you could also figure out the IP to hostname mapping by doing searching the master minion logs. It's a bit hard to explain though and I think what you have here is plenty helpful so fine to leave out

docs/_HowTo.md
```diff
@@ -292,3 +292,127 @@ onto the filesystem.
...
+`buggyContainer` is trying to connect to based on the application logs or
+the port that `buggyContainer` is trying to connect to.  To see the application
+logs:
+
+```console
+$ kelda logs buggyContainer
+```
+
+In this case, `buggyContainer` needs access to port `5432` on a container that
```
kklin: Do you want to put port numbers in backticks? I don't care either way but there's some inconsistencies (I think everywhere else doesn't have backticks)

kayousterhout:
@kklin I updated this - let me know what you think of the new text around private IPs

docs/_HowTo.md
kayousterhout: Ah that is useful to know but I agree probably too complicated, at least for now

docs/_HowTo.md
kayousterhout: I updated this and also the IP addresses, which were similarly inconsistent

PR #1276, opened 2017-11-15
amazon: Always return machine size when listing

kklin:
Amazon spot requests can be listed before their associated instance has
actually been fulfilled. This resulted in a period of time right after
boot where a cloud list would show spot requests without any sizes --
this would cause the cloud join to destroy the machines without sizes,
and boot new ones.

Before, the machines appeared to have no size because the previous code
pulled the size information from the instance associated with the spot
request. However, for that period of time where there was no associated
instance, there was no source for that information. This patch fixes
this by populating the size information from the spot request when there
is not yet an associated instance running.


HongjianZhang:
Why do we allow the spot requests to be listed when the associated instance has not been fulfilled?

kklin:
@HongjianZhang it's because of the way the logic that waits for machines to boot is pulled out of the provider (https://github.com/kelda/kelda/blob/master/cloud/cloud.go#L217). The way that code works is that it calls the provider's boot function, which returns the IDs that will exist once the cloud has finished booting, and then repeatedly calls `List` until those IDs show up. So we can't just not return any IDs because then that code won't block at all -- the cloud join will then quickly get called, and it will boot even more machines since it doesn't think there are any machines running.

Alternatively, the cloud provider could not return until the spot requests have an instance associated with them (i.e. pull the wait logic in `cloud.go` into the provider). This will slow things down though since then the cloud code can't wait for all the cloud operations in parallel (this was one of the benefits of pulling out the wait). The other ugly thing is that the wait logic would be duplicated in both the cloud and the provider.

PR #1277, opened 2017-11-15
Enable passing command line arguments to blueprints

luise:
Now, users can pass command line arguments to their blueprint when executing
`kelda run` and `kelda inspect`. For instance, if they execute
`kelda run ./blueprint.js arg0 arg1`, blueprint.js can access "arg0"
and "arg1" as process.argv[2] and process.argv[3]. This is exactly how they
would do it if they had run `node blueprint.js arg0 arg1`.

This will allow users to customize their deployments without opening the blueprint. For instance, in the Quick Start guide, we can now modify the blueprint to take the git repository from the command line, so they can deploy their own application just with `kelda run ./nodeExample.js https://github.com/me/myApp.git` instead of editing a variable in the blueprint.

For reference, process.argv[0] is the path to the node binary, and process.argv[1] is the absolute path to the node script.


integration-tester/tests/command-line-args/command_line_args_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/kelda/kelda/blueprint"
+	"github.com/kelda/kelda/integration-tester/util"
+	"github.com/stretchr/testify/assert"
+)
+
+const testBlueprint = `const kelda = require('kelda');
+const infrastructure = require('../../config/infrastructure.js');
```
kklin: Ah sorry I forgot that referencing the infrastructure will require a relative path, which means that this won't work for the same reason we couldn't just put a `blueprint.js` in the test suite, and exec `kelda run ./blueprint.js <args>` in the test (because the test is actually invoked from a parent, so the command is `./tests/command-line-args/command_line_args.test`, and the relative path `./blueprint.js` won't be `./tests/command-line-args/blueprint.js`).

I think the next best thing we can do is to have a `dep/blueprint.js` in the test suite, and figure out the path to pass to `kelda run` by figuring out the path of the test executable (https://stackoverflow.com/questions/18537257/how-to-get-the-directory-of-the-currently-running-file). Note that the blueprint will have to be in a subdirectory of the test suite because if it's in the same directory as the test, the integration-tester will get confused and possibly try to run it.

Let me know if you want to talk about this in person, it'd probably be easier. Sorry for telling you the wrong approach!

luise: Thanks for your help, @kklin ! I updated the test to use absolute paths as in the link you posted.

luise:
@aegamesi, just so you know, @kklin didn't take over the review -- he just left a comment about something we talked about earlier. No rush to review though!

cli/command/inspect/inspect.go
```diff
@@ -43,10 +46,11 @@ func (iCmd *Inspect) Parse(args []string) error {
 	iCmd.outputType = args[1]
 	switch iCmd.outputType {
```
aegamesi: Is this (kind of weird looking) switch statement the idiomatic Go way of doing this?

luise: Yeah, I think it is a decent to check against three strings (see https://github.com/golang/go/wiki/Switch#multiple-cases). I just made some tweaks to make the logic clearer. Do you find it less ugly now?

aegamesi: Oh, yeah, I think that looks better.

blueprint/blueprint.go
```diff
@@ -116,6 +117,12 @@ var lookPath = exec.LookPath
 
 // FromFile gets a Blueprint handle from a file on disk.
 func FromFile(filename string) (Blueprint, error) {
+	return FromFileWithArgs(filename, []string{})
```
kklin: I think it's slightly cleaner to do `FromFileWithArgs(filename, nil)`, but that might just be personal preference

cli/command/run.go
```diff
@@ -58,18 +61,22 @@ func (rCmd *Run) Parse(args []string) error {
 			return errors.New("no blueprint specified")
 		}
 		rCmd.blueprint = args[0]
+		rCmd.blueprintArgs = args[1:]
+	} else {
+		// When the blueprint is passed as a flag value, rather than as an
+		// argument, all args are blueprint args.
+		rCmd.blueprintArgs = args[0:]
```
kklin: This could just be `args`

cli/command/run_test.go
```diff
@@ -223,13 +223,30 @@ func TestRunFlags(t *testing.T) {
 
 	expBlueprint := "blueprint"
 	checkRunParsing(t, []string{"-blueprint", expBlueprint},
-		Run{blueprint: expBlueprint}, nil)
-	checkRunParsing(t, []string{expBlueprint}, Run{blueprint: expBlueprint}, nil)
+		Run{blueprint: expBlueprint, blueprintArgs: []string{}}, nil)
```
kklin: You could omit `blueprintArgs` altogether since `nil` is the default value

integration-tester/tests/command-line-args/command_line_args_test.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		t.Fatal("couldn't get api client")
+	}
+	defer clnt.Close()
+
+	containers, err := clnt.QueryContainers()
+	if err != nil {
+		t.Fatal("couldn't query containers")
+	}
+	assert.Len(t, containers, 0)
```
kklin: I think there's an `assert.Empty`

cli/command/run_test.go
luise: It actually has to be set to an empty list rather than nil since the `Parse` function always sets `blueprintArgs` to be a list (even if it's empty).

PR #1278, opened 2017-11-16
Improve JS documentation

kayousterhout:


js/bindings/bindings.js
```diff
@@ -1370,8 +1370,11 @@ class RuntimeValue {
   }
 }
 
-// hostIP is the RuntimeValue for the public IP of the machine the container is
-// running on.
+/**
+ * hostIP is the {@link RuntimeValue} for the public IP of the machine the container is
+ * running on. If used (e.g., as the value of an environment variable), Kelda will
+ * substitute the host's public IP address at runtime.
```
luise: Maybe : "... Kelda will replace the value with the host's..."?

This might just be because I'm not a native speaker, but I think this could be misread as "Kelda will set the host's IP to be the value of this constant" (which doesn't make a lot of sense given that users can't set the value, but I was personally confused for a sec).

kayousterhout: I decided this would just be clearer as an example - what do you think?

luise: Yeah, that's a good idea! I think "at container runtime" is a little systems experty. Maybe "when the container has booted," or "when Kelda knows which host the container will run on"? (I'm not sure when the value is actually set)

kayousterhout: I just cut that part because I couldn't think of a good wording and there's already documentation for RuntimeValue that attempts to describe this.

luise:
Wait, actually, would it make sense to set the HOST_IP of two different containers in the example? Just to emphasize that the hostIP evaluates to different values?

kayousterhout:
I can't think of a good way of doing that because it's possible the two containers could be placed on the same machine, in which case the IP addresses could be the same (and including that whole explanation in the example seems like it would make it too hard to understand).

luise:
Hm, sure. The comment could just say "IP of the machine the container is running on -- which could be different for the two machines". 
You don't have to do it here, but I think we should definitely explain this somewhere. Maybe exactly because it's hard to understand.

kayousterhout:
Updated.

luise:
Looks good!

PR #1280, opened 2017-11-17
ci: Various changes to facilitate scale testing

kklin:
I've accumulated some changes while setting up scale testing. The scale test
setup isn't ready to be merged yet, but these changes are.


integration-tester/tests/30-mean/mean.js
```diff
@@ -6,9 +6,11 @@ const infrastructure = require('../../config/infrastructure.js');
 
 const infra = infrastructure.createTestInfrastructure();
 
-const mongo = new Mongo(3);
+// Mongo supports at most 7 voting members.
+const numMongo = Math.min(7, Math.floor(Math.floor(infrastructure.nWorker / 2)));
```
aegamesi: What is the purpose of the nested Math.floor? Also in integration-tester/tests/django/django.js .

kklin: Copy paste mistake, thanks!

PR #1282, opened 2017-11-20
Update several runtime dependencies

kklin:


aegamesi:
The "ci: Add test to check for outdated runtime dependencies" commit is somewhat huge and does multiple things: renames 100-logs to 100-meta, adds all of the 100-meta dependencies, and then finally adds "integration-tester/tests/100-meta/version_test.go".

I think it would be a lot easier to read and see what's going on if you added all of the dependencies in a separate commit -- at least separate that from actually adding the version test.

kklin:
@aegamesi done

integration-tester/tests/100-meta/version_test.go
```diff
@@ -0,0 +1,289 @@
...
+	assertNewestVersion(t, "docker", currVer, allVersions)
+}
+
+func TestOVSVersion(t *testing.T) {
+	skipIfNotDev(t)
+
+	aMaster, err := getMaster()
+	assert.NoError(t, err)
+
+	allVersions, err := parseReleaseLinks("http://openvswitch.org/releases/",
```
aegamesi: Is this the correct URL? I'm getting a 404 for http://openvswitch.org/releases/

integration-tester/tests/100-meta/version_test.go
```diff
@@ -0,0 +1,289 @@
...
+		t.Skip(`Skipping version check because Kelda version is not "dev"`)
+	}
+}
+
+func assertNewestVersion(t *testing.T, name, currVersionStr string,
+	allVersionStrs []string) {
+	currVersion, err := semver.NewVersion(toSemanticVersion(currVersionStr))
+	assert.NoError(t, err, "failed to parse current version for "+name)
+
+	assert.NotEmpty(t, allVersionStrs)
```
aegamesi: If the format of the download page changes and the tester is unable to retrieve *any* links, this assertion would fail. That's probably the only reason why the assertion would fail, actually -- so it could be useful to put some kind of message here indicating that as a reminder if/when the download page format changes and the integration tests start failing.

integration-tester/tests/100-meta/version_test.go
kklin: I swear that link used to work.. https://web.archive.org/web/20171110224842/http://openvswitch.org/releases

Anyways, I updated it to be http://openvswitch.org/download/

kklin:
@aegamesi ready for more review!

PR #1283, opened 2017-11-20
Properly fix namespace switching

kklin:
The previous fix made it so the containers weren't destroyed, but it didn't
address the underlying problem. Note that the previous solution of persisting
the etcd data to disk is still necessary for adding and removing master nodes.


cloud/foreman/foreman.go
```diff
@@ -173,19 +181,23 @@ func runOnce(conn db.Conn, cloudID string) (currConfig pb.MinionConfig, connecte
 	return
 }
 
-func makeConfig(machines []db.Machine, minionMachine db.Machine,
-	blueprint string) pb.MinionConfig {
+func makeConfig(waitForMachinesCutoff time.Time, machines []db.Machine,
```
ejj: I personally think this would be a bit cleaner if you actually separated the cutoff separation into its own function.  I.e. a `clusterReady(time.Time, []db.Machine) bool` that was called before makeConfig. I think the code would be a bit easier to follow this way.

kklin: Done, FYI I made `clusterReady` take just `[]db.Machine` and have the caller check the `time`. The `waitForMachinesCutoff` is a bit confusing, so the smaller its scope, the easier it is to document I think.

PR #1285, opened 2017-11-21
node: Try both node and nodejs binaries

luise:
Even though `node` is the blessed name for the Node.js binary, some package
managers (e.g. apt-get) install the binary as `nodejs`. Before, we only tried
the `node` binary, but with this commit, Kelda also tries the `nodejs` binary.

The `nodejs` name is used by package managers that already had a package named
`node`. To not accidentally execute the wrong `node` binary, we try the nodejs
binary first.


blueprint/blueprint.go
```diff
@@ -123,9 +125,9 @@ func FromFile(filename string) (Blueprint, error) {
 // FromFileWithArgs gets a Blueprint handle from a file on disk, passing the
 // given arguments to the node process.
 func FromFileWithArgs(filename string, cmdLineArgs []string) (Blueprint, error) {
-	if _, err := lookPath("node"); err != nil {
-		return Blueprint{}, errors.New(
-			"failed to locate Node.js. Is it installed and in your PATH?")
+	nodeBin, err := util.GetNodeBinary()
```
kklin: Can you just call this `nodeBinary`? `nodeBin` makes me think of a `/bin` folder that contains executables

cli/command/init.go
```diff
@@ -67,6 +62,18 @@ func (iCmd *Init) Run() int {
 		return 1
 	}
 
+	nodeBin, err := util.GetNodeBinary()
```
kklin: If you change the naming, you can do it here as well

util/util.go
```diff
@@ -209,3 +210,15 @@ func JoinNotifiers(a, b chan struct{}) chan struct{} {
 	}()
 	return c
 }
+
+// GetNodeBinary returns the name of the Node.js binary.
+func GetNodeBinary() (string, error) {
```
kklin: I think a comment here (basically your commit message would be awesome) about why this function is necessary, and why we try `node` first, would be helpful.

luise: Ah, yeah, good idea. I'll do that

PR #1288, opened 2017-11-21
client: Don't query unconnected machines for leader information

kklin:
Before this commit, we queried all machines for leader information,
regardless of whether they had connected. If a machine is not connected,
it probably won't respond, so querying it only increases the amount of
time required for the function to return. This resulted in `kelda ps`
being extremely slow when a machine had not yet connected (which can
happen when the cluster is first booting, or if a machine crashes).

Sometimes, the error "context deadline exceeded" would be printed
because the Query call to the daemon would timeout since obtaining the
leader client took so long.


ejj:
Well that's simple.  Doesn't fix every case, but not a bad idea =)

PR #1289, opened 2017-11-21
ci: Reapply commit 60cd6429 (Block container listing until both master and worker respond)

kklin:
Commit 69153285 (ci: Run the namespace switch test before the logs test)
accidentally reverted the commit while moving the file.


kklin:
The commit was actually properly applied, I just didn't look far enough down in the file..

PR #1291, opened 2017-11-21
cli: Make 'kelda stop' prompt the user for confirmation

aegamesi:
Previously, the stop command would immediately tell the daemon to halt
all of the machines or containers without confirming with the user that
this is the intended action. Now, the stop command warns the user if
running 'kelda stop' will terminate any containers or machines.

Some logic was slightly simplified as well, which caused 'stop' to not
bother deploying an empty blueprint if it was already empty, which had
the added effect of requiring some of the previously existing tests to
be slightly changed.


cli/command/stop.go
```diff
@@ -77,6 +84,36 @@ func (sCmd *Stop) Run() int {
...
+		if err != nil {
+			log.WithError(err).Error("Unable to diff deployments.")
+			return 1
+		}
+		if diff == "" {
+			fmt.Println("No change.")
+			return 0
+		}
+
+		fmt.Printf("This will stop %s running on %s.\n",
```
luise: `inflect` is neat, but I think it might be fine to just log `This will stop %d container(s) on %d machine(s)` which would simplify the code.

aegamesi: Yeah, I had that originally but talked to Ethan about it -- I think we didn't actually come to a specific conclusion about it though.

luise: Also, it's not clear that this will stop the machines also. It sounds like it will just stop the containers. It would be clearer to say something like `This will stop X containers and Y machines.`
Oh, also, remember to account for the fact where only the containers will be stopped (i.e. when `sCmd.onlyContainers` is true).

luise: Ah, ok. The message is obviously prettier when using `inflect`, but I just found it unintuitive before looking what `inflect` does.

aegamesi: If it were called 'pluralize', do you think it would be better?

luise: Yeah, that would be better!  It would also be nice to simplify the arguments then. I would probably just make the function dumb and only take the count and singular version of the word, and then just append an `s` if the count isn't 1. That should be fine for this case.

cli/command/stop.go
```diff
@@ -77,6 +84,36 @@ func (sCmd *Stop) Run() int {
...
+			inflect(len(currDepl.Containers), "container", "containers"),
+			inflect(len(currDepl.Machines), "machine", "machines"))
+	} else {
+		fmt.Println("This will stop an unknown number of machines and " +
+			"containers.")
+	}
+
+	if !sCmd.force {
+		shouldStop, err := confirm(os.Stdin, "Continue stopping namespace?")
```
luise: I would probably say "deployment" instead of "namespace" -- just because we assume (at least for now) that most users won't have to think about namespaces.

cli/command/stop.go
```diff
@@ -77,6 +84,36 @@ func (sCmd *Stop) Run() int {
...
 			newCluster.Machines = currDepl.Machines
 		}
+
+		diff, err := diffDeployment(currDepl.String(), newCluster.String())
+		if err != nil {
+			log.WithError(err).Error("Unable to diff deployments.")
+			return 1
+		}
+		if diff == "" {
+			fmt.Println("No change.")
```
luise: This is super subjective, and feel free to ignore it, but it might be clearer to say something like "Nothing to stop.

cli/command/stop_test.go
```diff
@@ -77,15 +83,56 @@ func TestStopFlags(t *testing.T) {
 	t.Parallel()
 
 	expNamespace := "namespace"
-	checkStopParsing(t, []string{"-namespace", expNamespace}, expNamespace, nil)
-	checkStopParsing(t, []string{expNamespace}, expNamespace, nil)
-	checkStopParsing(t, []string{}, "", nil)
+	checkStopParsing(t, []string{"-namespace", expNamespace},
```
luise: Maybe add a test where both `force` and a namespace are passed

cli/command/stop_test.go
```diff
@@ -77,15 +83,56 @@ func TestStopFlags(t *testing.T) {
...
+		c.On("QueryBlueprints").Return([]db.Blueprint{{
+			Blueprint: blueprint.Blueprint{
+				Namespace:  "ns",
+				Machines:   []blueprint.Machine{{}, {}},
+				Containers: []blueprint.Container{{}},
+			},
+		}}, nil)
+		c.On("Deploy", blueprint.Blueprint{Namespace: "ns"}.String()).Return(nil)
+
+		stopCmd := &Stop{
```
luise: I would probably just simplify this to: 
```go
stopCmd := NewStopCommand()
stopCmd.client = c
```

aegamesi:
@luise Thanks! I've addressed your comments.

cli/command/stop.go
```diff
@@ -77,6 +84,33 @@ func (sCmd *Stop) Run() int {
...
+			fmt.Println("Nothing to stop.")
+			return 0
+		}
+
+		containersDelta := len(currDepl.Containers) - len(newCluster.Containers)
+		machinesDelta := len(currDepl.Machines) - len(newCluster.Machines)
+		fmt.Printf("This will stop %s and %s.\n",
+			pluralize(containersDelta, "container"),
+			pluralize(machinesDelta, "machine"))
+	} else {
```
luise: Right now, `kelda stop` will say this even if I say `kelda stop myRunningNamespace`, where `myRunningNamespace` is the daemon's active namespace (i.e. when the given namespace is the namespace of `currDepl`). If the namespaces are equal, we should print the more informative message.

aegamesi:
@luise I've addressed your new feedback

PR #1293, opened 2017-11-25
Various digital ocean tweaks.

ejj:
Most of these are aimed at scaling to 64 nodes for spark.

### DIFF
--- a/cloud/amazon/amazon.go
+++ b/cloud/amazon/amazon.go
@@ -625,6 +625,29 @@ func logACLs(add bool, perms []*ec2.IpPermission) {
 	}
 }
 
+// Cleanup removes unnecessary detritus from this provider.  It's intended to be called
+// when there are no VMs running or expected to be running soon.
+func (prvdr *Provider) Cleanup() error {
+	groups, err := prvdr.DescribeSecurityGroup(prvdr.namespace)
+	if err != nil {
+		return err
+	}
+
+	for _, group := range groups {
+		log.WithFields(log.Fields{
+			"name":   *group.GroupName,
+			"id":     *group.GroupId,
+			"region": prvdr.region,
+		}).Debug("Amazon Delete Security Group")
+
+		if err := prvdr.DeleteSecurityGroup(*group.GroupId); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // blockDevice returns the block device we use for our AWS machines.
 func blockDevice(diskSize int) *ec2.BlockDeviceMapping {
 	return &ec2.BlockDeviceMapping{
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -28,6 +28,15 @@ type provider interface {
 	SetACLs([]acl.ACL) error
 
 	UpdateFloatingIPs([]db.Machine) error
+
+	// The Cleanup() function will be called occaisionally in those regions that have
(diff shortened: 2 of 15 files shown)


kklin:
I'll review this since I probably need it for scaling the integration tests

cloud/google/google.go
```diff
@@ -461,6 +461,15 @@ func (prvdr *Provider) UpdateFloatingIPs(machines []db.Machine) error {
 	return nil
 }
 
+// Cleanup removes unnecessary detritus from this provider.  It's intended to be called
+// when there are no VMS running or expected to be running soon.
```
kklin: VMS -> VMs

cloud/vagrant/vagrant.go
```diff
@@ -132,3 +132,9 @@ func (prvdr Provider) SetACLs(acls []acl.ACL) error {
 func (prvdr *Provider) UpdateFloatingIPs([]db.Machine) error {
 	return errors.New("vagrant provider does not support floating IPs")
 }
+
+// Cleanup removes unnecessary detritus from this provider.  It's intended to be called
+// when there are no VMS running or expected to be running soon.
```
kklin: VMS -> VMs

cloud/amazon/amazon.go
```diff
@@ -625,6 +625,29 @@ func logACLs(add bool, perms []*ec2.IpPermission) {
 	}
```
kklin: In the commit message, "cloud provides" -> "cloud providers"

cloud/cloud.go
```diff
@@ -28,6 +28,8 @@ type provider interface {
 	SetACLs([]acl.ACL) error
 
 	UpdateFloatingIPs([]db.Machine) error
+
+	Cleanup() error
```
kklin: I had to think a bit about the role of this function, and I think it'd be helpful if its constraints were documented more. At first I thought it was like a `Close` function where after it's been called, it won't be called again (but it'll actually be called every two minutes). It would also be helpful to document somewhere that `List` is called every two minutes, so if the provider implementation creates resources in that function for some reason, `Cleanup` won't behave properly.

Basically, I think this function imposes a very specific structure on the way cloud providers should be implemented, and that it'd be helpful for that to be documented somewhere.

Another example is that the Google provider only creates the Network in the `New` function, so if `Cleanup` removed the `Network` (which might be reasonable if it behaved like `Close`), then future `Boot` calls will fail since the network is gone.

cloud/google/google.go
```diff
@@ -461,6 +461,15 @@ func (prvdr *Provider) UpdateFloatingIPs(machines []db.Machine) error {
...
+func (prvdr *Provider) Cleanup() error {
+	// XXX: The only thing that could need cleaning up is firewalls, but those are
```
kklin: FWIW the network needs to be cleaned up as well, although I'm fine with not doing that in this commit.

kklin: It'll take some work since I think a single network is used for all regions, rather than a network per region.

ejj: Yeah google is a mess ....

ejj: Changed the comment to this:

```+       // XXX: There are several resources that could be cleaned up here.  Some are
+       // handled in a sort of ad hoc way in the other provider functions.  Some are not
+       // cleaned up at all.```

PR #1294, opened 2017-11-26
Ci: Grep daemon log output for errors

aegamesi:


integration-tester/tests/100-logs/logs_test.go
```diff
@@ -27,21 +31,47 @@ func TestMinionLogs(t *testing.T) {
...
+		for _, regex := range ignoreRegexes {
+			if regex.MatchString(line) {
+				fmt.Printf("Ignoring line: %s\n", line)
+				continue outer
+			}
+		}
+
 		// Errors pulling an image are completely out of our control.  They
 		// should still be logged as warnings to the user, but they shouldn't
 		// cause a test failure.
```
kklin: Can you make this part of `ignoreRegexes` as well?

integration-tester/tests/100-logs/logs_test.go
```diff
@@ -27,21 +31,47 @@ func TestMinionLogs(t *testing.T) {
...
+	daemonLogOutput, err := ioutil.ReadFile(daemonLoggerPath)
+	if err != nil {
+		t.Fatalf("couldn't read daemon output: %s", err.Error())
+	}
+
+	checkLogs(t, string(daemonLogOutput))
+}
+
+var ignoreRegexes = []*regexp.Regexp{
+	regexp.MustCompile("(?:ERROR|WARNING) \\[.*? 500 Server was " +
```
kklin: A comment for some context about these errors would be helpful. At least mentioning that they originate from the DigitalOcean provider. This way it'll be more obvious when it's safe to remove them.

integration-tester/tests/100-logs/logs_test.go
```diff
@@ -52,22 +82,22 @@ func checkString(t *testing.T, str string) {
...
+			t.Error("Logs have a stack trace")
 		}
 
 		// The trailing open bracket is necessary to filter out false positives
 		// in the log output. For example, as part of logging DNS requests, the
 		// status string NOERROR is printed.
 		if strings.Contains(line, "ERROR [") ||
-			strings.Contains(line, "WARN [") {
-			t.Errorf("Minion logs has error: %s", line)
+			strings.Contains(line, "WARNING [") {
```
kklin: Is it safe to remove `WARN`? Maybe logrus prints out both?

kklin: Either way, the change should be in a separate commit (the way I think about it is that if we need to revert the change, we wouldn't want to revert this entire PR)

integration-tester/tests/100-logs/logs_test.go
```diff
@@ -8,8 +8,9 @@ import (
```
kklin: I don't think the ignoring errors change should be in its own commit since the integration tests would be broken between this commit, and the commit that adds the test. It's borderline though, if you think the single commit would be too big then I'm fine having it separate.

integration-tester/tests/100-logs/logs_test.go
```diff
@@ -30,31 +31,47 @@ func TestMinionLogs(t *testing.T) {
 
 	for _, machine := range machines {
 		fmt.Println(machine)
-		logsOutput, err := exec.Command("kelda", "ssh", machine.CloudID,
```
kklin: This renaming stuff belongs better in the previous commit

integration-tester/tests/100-logs/logs_test.go
kklin: I looked through the logs and it looks like only `Warning` is printed! Good catch! (I wonder how many warnings we've missed because of this..)

aegamesi: Yep -- looks like Logrus can only print 'warning' anyway: https://github.com/sirupsen/logrus/blob/2c677e6a5e5ae8e1b967f1efcf2c3dbaaf625b2d/logrus.go#L23

aegamesi:
@kklin ready for another look!

kklin:
Sorry, looks like my last review didn't properly post. I said that the 'WARN' -> 'WARNING' change should be in a separate commit, and that this needs a rebase

aegamesi:
@kklin Like this?

kklin:
Yep, there's still a merge conflict though

aegamesi:
Weird -- Git was able to automatically rebase locally. Anyway, resolved @kklin .

integration-tester/tests/100-meta/logs_test.go
```diff
@@ -7,6 +7,10 @@ import (
 	"testing"
 
 	"github.com/kelda/kelda/integration-tester/util"
+	"io/ioutil"
```
kklin: Ah, I didn't notice this earlier sorry @aegamesi. These imports should be grouped with the other stdlib ones above.

aegamesi: Fixed!

PR #1295, opened 2017-11-26
Fix bug where blueprint would sometimes not get enforced when the cluster first boots

kklin:
This should fix the failed 10-network integration tests where the containers never boot.


db/db.go
```diff
@@ -103,7 +103,8 @@ func (tr Transaction) Run(do func(db Database) error) error {
```
ejj: "even" -> "event" in the commit message.

PR #1297, opened 2017-11-27
Fix assigning floating IPs in new cluster

kklin:
@ejj I'm assigning you since you're really familiar with the cloud code


cloud/join.go
```diff
@@ -127,9 +127,17 @@ func (cld cloud) planUpdates(view db.Database) joinResult {
...
-		if bpm.FloatingIP != dbm.FloatingIP {
-			dbm.FloatingIP = bpm.FloatingIP
-			res.updateIPs = append(res.updateIPs, dbm)
+		// Only update IPs once the roles are set. This way, we avoid assigning
+		// a floating IP to a machine that is going to connect with the other
+		// role, which will require reassigning the IP. This also prevents the
+		// cloud code from attempting to assign floating IPs while machines are
+		// still booting, which can be a problem for Amazon preemptible
+		// instances.
+		if bpm.Role != db.None && dbm.Role != db.None {
```
ejj: The blueprint role can't be db.None (I think)

ejj: Btw I assigned this PR to you: https://github.com/kelda/kelda/pull/1300

It's been sitting in a branch I've been working on, and this code made me think of it.

cloud/join.go
```diff
@@ -180,13 +180,21 @@ func machineScore(left, right interface{}) int {
 		return -1
```
ejj: Nice commit message!

integration-tester/tests/move-floating-ip/moveFloatingIP.js
```diff
@@ -0,0 +1,50 @@
...
+
+const infra = infrastructure.createTestInfrastructure();
+
+const workers = infra.machines.filter(m => m.role === 'Worker');
+if (workers.length === 0) {
+  throw new Error('Failed to find any worker machines');
+}
+
+let workerWithFloatingIp;
+for (let i = 0; i < workers.length; i += 1) {
```
ejj: I'm somewhat surprised this floating IP reassignment stuff works.  Won't it end up just assigning the floating IP to a machine that's otherwise exactly the same, but just appears in a different order in the list of machines in the json object? I.E. wouldn't you need the machines to have some distinguishable characteristic (i.e. size for example) for this test to be meaningful?

kklin: Ah yeah I think you're right

kklin: I changed the test so that it moves the floating IP from the worker to the master. I have two questions:
- Do you think that test is OK to merge while we know about the bug where machines might be terminated/restarted when moving floating IPs between roles? I'm leaning towards merging it now so that we can collect more data, but I also don't want to make Jenkins failures expected, which would decrease the response to other failures that we can do something about immediately.
- Do you want to review the test again?

PR #1299, opened 2017-11-28
Add better debug logging for Spark tests

kayousterhout:


integration-tester/tests/20-spark/spark_test.go
```diff
@@ -77,3 +79,37 @@ func TestRunShuffleJob(t *testing.T) {
 			outputAsNum)
 	}
 }
+
+// outputWorkerLogs prints the logs from all of the workers (Spark logs this output in a
+// special file, so it's not automatically included in the logs that Jenkins saves after
+// running each test).
+func outputWorkerLogs(t *testing.T) {
+	for _, name := range []string{"spark-worker", "spark-worker2"} {
```
kklin: Can you get these names by using `QueryContainers` so it works for all cluster sizes? This won't behave as expected when booting at larger scales

integration-tester/tests/20-spark/spark_test.go
```diff
@@ -77,3 +79,37 @@ func TestRunShuffleJob(t *testing.T) {
...
+		// Get the most recent Spark application that ran (this handles the case
+		// where a user runs the integration tests multiple times, using the
+		// same set of containers).
+		workerCmd := exec.Command("kelda", "ssh", name, "ls", "-t", "spark/work")
+		workerBytes, workerErr := workerCmd.CombinedOutput()
+		if workerErr != nil {
+			t.Fatalf("Failed to list logs in spark/work: %s\n%s",
+				workerBytes, workerErr)
```
kklin: does `workerBytes` need to be wrapped in a `string`?

integration-tester/tests/20-spark/spark_test.go
```diff
@@ -77,3 +79,37 @@ func TestRunShuffleJob(t *testing.T) {
...
+		}
+		appName := strings.Split(
+			strings.Trim(string(workerBytes), " \n\t"),
+			"\n")[0]
+		// Log all executors for the application (there may be multiple if
+		// some failed and the driver re-started them). This command needs to
+		// use bash because the default shell doesn't evaluate the wildcard
+		// correctly.
+		lsCmd := fmt.Sprintf("bash -c \"cat spark/work/%s/*/stderr\"",
```
kklin: Rather than escaping the double quotes you could use backticks. then you won't need to escape the double quotes inside the string

integration-tester/tests/20-spark/spark_test.go
```diff
@@ -77,3 +79,37 @@ func TestRunShuffleJob(t *testing.T) {
...
+			appName)
+		workerCmd = exec.Command("kelda", "ssh", name, lsCmd)
+		workerBytes, workerErr = workerCmd.CombinedOutput()
+		if workerErr != nil {
+			t.Fatalf("Failed to get worker logs: %s\n%s",
+				workerBytes,
```
kklin: Same here re wrapping in a `string`

integration-tester/tests/20-spark/spark_test.go
```diff
@@ -77,3 +79,37 @@ func TestRunShuffleJob(t *testing.T) {
...
+// running each test).
+func outputWorkerLogs(t *testing.T) {
+	for _, name := range []string{"spark-worker", "spark-worker2"} {
+		// Get the most recent Spark application that ran (this handles the case
+		// where a user runs the integration tests multiple times, using the
+		// same set of containers).
+		workerCmd := exec.Command("kelda", "ssh", name, "ls", "-t", "spark/work")
+		workerBytes, workerErr := workerCmd.CombinedOutput()
+		if workerErr != nil {
+			t.Fatalf("Failed to list logs in spark/work: %s\n%s",
```
kklin: Would it be helpful if these were `Error`s rather than `Fatal`s so that all of the logs will be collected even if we can't get them from the first container?

integration-tester/tests/20-spark/spark_test.go
kklin: Nit - I found the name `lsCmd` a bit confusing since I thought you would be listing files and iterating over them or something, but this is actually grabbing the file contents too.

kayousterhout: Oh oops thank for pointing this out - the command used to be `ls` instead of `cat` and I didn't notice when I updated

integration-tester/tests/20-spark/spark_test.go
kayousterhout: Nope %s works with byte arrays (just double checked this and the output is the same)

kklin: Cool I didn't know that!

kayousterhout:
Thanks for the review @kklin - updated!

PR #1304, opened 2017-12-01
cloud: Make provider initialization more robust

kayousterhout:
This commit makes two related changes to the way providers are
initialized:
(1) Continuously re-try to initialize providers when initialization
fails, so that the Daemon can connect to a provider if, for example, a
user adds credentials in the correct location after the daemon has
already started. This functionality used to exist
(https://github.com/kelda/kelda/commit/dd11b5b060b8af08d183a29f3f2ca16c412848d1)
but was removed when the cloud code was recently refactored.
(2) Log an error message (rather than a debug message) when a provider
that is active (i.e., that is referenced by a blueprint, or that has
running machines) cannot be initialized. This fixes a problem where,
unless the user had debug logging on, a user could run a blueprint
that references a provider that isn't properly configured, and the
daemon would silently fail.


cloud/cloud.go
```diff
@@ -162,7 +143,57 @@ func (cld cloud) run(stop <-chan struct{}) {
...
+	var numBPMachines int
+	txnErr := cld.conn.Txn(db.BlueprintTable, db.MachineTable).Run(
+		func(view db.Database) error {
+			numDBMachines = len(cld.selectMachines(view))
+
+			bp, bpErr := view.GetBlueprint()
+			if bpErr != nil {
+				return bpErr
+			}
+			numBPMachines = len(cld.desiredMachines(bp.Blueprint.Machines))
```
kklin: Do we have to convert the machines into the desiredMachines?

cloud/cloud.go
```diff
@@ -162,7 +143,57 @@ func (cld cloud) run(stop <-chan struct{}) {
...
+			return nil
+		})
+	if txnErr != nil {
+		log.WithFields(log.Fields{"provider": cld.String()}).Warnf(
+			"failed to initialize cloud provider (error: %s) and failed "+
```
kklin: I think in general logrus recommends logging errors using `Fields` rather than directly in the string since the `Fields` is easier for a machine to parse (e.g. logrus has an option to use a JSON logger).

That said, I'm fine with putting the error directly in the string if you think that's easier to read since this is an error that we want to make as easy to understand as possible.

cloud/cloud.go
```diff
@@ -162,7 +143,57 @@ func (cld cloud) run(stop <-chan struct{}) {
...
+// Returns whether the provider is active.
+func (cld *cloud) logInitializationErr(err error) bool {
+	// Check whether the provider is needed by the current blueprint, or
+	// if the database thinks that there are running machines for the
+	// provider.
+	var numDBMachines int
+	var numBPMachines int
+	txnErr := cld.conn.Txn(db.BlueprintTable, db.MachineTable).Run(
+		func(view db.Database) error {
+			numDBMachines = len(cld.selectMachines(view))
```
kklin: I'm not sure it's worth checking whether there are any machines in the database for the failed provider -- I don't think it can every be non-zero. If the provider was never initialized, it can't have listed machines and placed them in the db. And when there's a namespace change, we zero out the machines (https://github.com/kelda/kelda/blob/master/api/server/server.go#L236)

cloud/cloud.go
```diff
@@ -162,7 +143,57 @@ func (cld cloud) run(stop <-chan struct{}) {
...
+
+			bp, bpErr := view.GetBlueprint()
+			if bpErr != nil {
+				return bpErr
```
kklin: I don't think it's possible for the `GetBlueprint` call to fail, so you could panic (https://github.com/kelda/kelda/blob/master/cloud/join.go#L104), or treat the number of machines as zero. This does feel safer though, so I'm fine leaving it.

cloud/cloud.go
```diff
@@ -162,7 +143,57 @@ func (cld cloud) run(stop <-chan struct{}) {
...
+			}
+			numBPMachines = len(cld.desiredMachines(bp.Blueprint.Machines))
+			return nil
+		})
+	if txnErr != nil {
+		log.WithFields(log.Fields{"provider": cld.String()}).Warnf(
+			"failed to initialize cloud provider (error: %s) and failed "+
+				"to get information from the database about running "+
```
kklin: Technically, the txn can only fail if you fail to get the blueprint, not if you fail to get the running machines

cloud/cloud.go
kklin: If you decide to not consider the machines in the db, then I'd propose removing the `bool` return argument, and just having the caller check whether there are machines in the blueprint that warrant the provider to be considered "active". I find it a bit weird that this "log" function returns this seemingly unrelated `bool`.

cloud/cloud_test.go
```diff
@@ -128,7 +138,55 @@ func TestPanicBadProvider(t *testing.T) {
...
+		conn:         db.New(),
+		namespace:    "test",
+		region:       testRegion,
+		providerName: FakeAmazon,
+	}
+	cld.runOnce()
+	assert.Nil(t, cld.provider)
+
+	// Make sure that an error is logged.
+	assert.Equal(t, 1, len(hook.Entries))
```
kklin: Would it be hard to check the log entry level, and check that when there's relevant machines in the blueprint it's logged at Error, and otherwise it's logged at Debug?

cloud/cloud.go
kayousterhout: Why isn't it possible for the call to fail? (My understanding was that the reason for panicking in the spot you linked was because the blueprint had already been fetched in the txn: https://github.com/kelda/kelda/blob/master/cloud/join.go#L43, which isn't the case here)

cloud/cloud.go
kayousterhout: We don't have to do the whole conversion - but we do need to filter out only the machines that are relevant for this provider / region and that seemed like a convenient way of doing that.  Do you think it's better to do only the filtering here rather than using desiredMachines?

cloud/cloud.go
kayousterhout: Ah I was wondering about the standard way of doing then when I wrote the code. I think both are easy to read so I changed it.

cloud/cloud_test.go
kayousterhout: The problem is that this test actually hits the (normally obscure) database error (so the error gets logged at warn), and it seemed like more work than it was worth to mock the database (although admittedly I didn't spend any time looking into the difficulty of mocking the DB call - so let me know if this is something that's easy to do)

cloud/cloud.go
kayousterhout: (I did realize that if this stays using desiredMachines, I should move that function out of join.go and to this file)

cloud/cloud.go
kklin: Ah yeah. I was thinking that the call couldn't fail because the providers are only started based on running a blueprint, but it's a tenuous/non-obvious thread, so I think it's good to check the error.

cloud/cloud_test.go
kklin: Rather than mock the database call with mockery, we mock the database by initializing the db connection with `Insert`s so that the call won't fail. So in this case, it should be as simple as
```
cld.conn.Txn(db.BlueprintTable).Run(func(view db.Database) error {
  bp := view.InsertBlueprint()
  bp.Machines = foo
  view.Commit(bp)
})
```

cloud/cloud.go
kayousterhout: Given that - what do you think is the right thing to do re: checking if the blueprint has active machines? It seems ugly to duplicate all of this logic in the caller to determine if the cloud is active. One alternative is to move the isActive member from joinResult to cloud, the logging function would set it, and then the caller could check that.

(in the meantime I renamed the logInitError function to handleInitError to make the return value seem slightly less weird, but I agree with your sentiment that it's still weird)

cloud/cloud_test.go
kayousterhout: Oh awesome thanks I'll do that!

cloud/cloud.go
kklin: Ah I see I didn't realize that `desiredMachines` is also filtering machines. That makes sense!

cloud/cloud.go
kklin: Maybe you could give the cloud a helper method that returns whether it's needed by any of the machines in the blueprint? Then the caller and this code would just both call that.

cloud/cloud.go
```diff
@@ -210,6 +204,49 @@ func (cld cloud) runOnce() (active bool) {
 	return jr.isActive
 }
 
+// usedByCurrentBlueprint returns whether this cloud provider is used by machines
+// in the blueprint that is currently active.
+func (cld *cloud) usedByCurrentBlueprint() bool {
+	var used bool
+	cld.conn.Txn(db.BlueprintTable, db.MachineTable).Run(
```
kklin: I don't think this txn needs to lock the Machine table anymore

cloud/cloud.go
```diff
@@ -210,6 +204,49 @@ func (cld cloud) runOnce() (active bool) {
...
+			bp, err := view.GetBlueprint()
+			if err != nil {
+				log.WithFields(log.Fields{
+					"provider": cld.String(),
+					"error":    err,
+				}).Warnf("failed to determine whether cloud provider " +
+					"is used by the current blueprint")
+				// If we're not sure if the cloud provider is used,
+				// conservatively assume that it is.
+				used = true
```
kklin: This might be personal preference, but I find the layer of indirection provided by the Txn a bit hard to follow. i'd prefer to move as much logic out of it as possible:

```
var bp db.Blueprint
var err error
conn.Txn(db.BlueprintTable).Run(func(view db.Database) error {
  bp, err = view.GetBlueprint()
})

if err != nil {
  log.Warn("foo")
  return true
}
return len(cld.desiredMachines(bp.Machines)) > 0
```

kayousterhout:
@kklin updated!

cloud/cloud.go
```diff
@@ -210,6 +204,49 @@ func (cld cloud) runOnce() (active bool) {
...
+
+// logInitializationErr handles errors initializing the cloud provider: if the
+// provider is active, it logs an error, and if the provider is not active (i.e.,
+// if the user isn't trying to use the provider), it logs a debug message.
+func (cld *cloud) logInitializationErr(err error) {
+	if cld.usedByCurrentBlueprint() {
+		log.WithFields(log.Fields{
+			"provider": cld.String(),
+			"error":    err,
+		}).Errorf("failed to initialize cloud provider used by the current " +
```
kklin: No need for these functions to have the trailing `f` now (this could just be `Error`)

cloud/cloud.go
```diff
@@ -210,6 +204,49 @@ func (cld cloud) runOnce() (active bool) {
...
+			"blueprint (will keep retrying)")
+	} else {
+		log.WithFields(log.Fields{
+			"provider": cld.String(),
+			"error":    err,
+		}).Debugf("failed to initialize cloud provider (will keep retrying)")
```
kklin: Part of me wants to reduce the repetition in these two logs. Maybe something like
```
logFields := log.Fields{
  "provider": cld.String,
  "error": err,
  "referencedByBlueprint": cld.usedByCurrentBlueprint(),
}

logMsg := "Failed to initialize cloud provider (will keep retrying)"
if cld.usedByCurrentBlueprint() {
  log.WithFields(logFields).Error(logMsg)
} else {
  log.WithFields(logFields).Debug(logMsg)
}
```

My thinking is that if we can simplify the log enough, we could just put it in the caller. I don't think it's very important though.

kayousterhout: Good idea done

kayousterhout:
One more Q @kklin: is this something you'd put in the changelog, or do you typically not include things this small?

kklin:
I haven't figured out a good rule yet. Lately I've been putting a lot less stuff in (I probably wouldn't have put this change in), but that's partly out of laziness/forgetfulness. Up to you

PR #1305, opened 2017-12-01
google: Behave better when an invalid floating IP is supplied

kklin:
Before we would remove the existing IP, and then the machine would become
un-listable. Kelda then couldn't modify the cluster (e.g. boot or stop
machines).


aegamesi:
So everything looks good to me, except when I was testing it I tried the case where I assigned the same floating IP to two different machines (after booting one machine with the floating IP and one machine without a floating IP, then updating the config), and I just got the expected:

"ERROR [Dec  1 12:57:28.023] Failed to update cloud provider.         region=Google-us-east1-b-default-namespace error=35.227.53.241 is not ready to be assigned count=1 action=update floating IPs"

That error is pretty misleading (I mean, the issue is that I'm trying to use one floating IP on two machines). It think it would be reasonable to have an error like '35.227.53.241 is already assigned' (if the status of the floating IP is, well, ASSIGNED). The root cause (2 machines with the same floating IP) I suppose should be addressed somewhere else, but this issue could also happen if you have 2 deployments on the same account.

kklin:
Thanks for the thorough testing @aegamesi! I've finally updated the PR, it's ready for another review

PR #1306, opened 2017-12-01
bindings: Remove machine blueprint ID

HongjianZhang:
The machine blueprint ID is no longer needed it. This patch will
remove them from js.


kklin:
BTW I actually noticed today that the ID is present in `blueprint.go` as "ID"

PR #1307, opened 2017-12-03
Fix failing 100-meta logs test

kklin:


integration-tester/tests/00-infrastructure-setup/infrastructure_setup_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"time"
+
+	"github.com/kelda/kelda/db"
+	testerUtil "github.com/kelda/kelda/integration-tester/util"
+	"github.com/kelda/kelda/util"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestInfrastructureSetup blocks until the machines in the infrastructure have
```
kayousterhout: This comment is confusing - is the point of this to test that the machines come up properly? Or is the point that this test will run before any of the other tests, so it ensures that all of the machines have been deployed properly before the other tests run? I think the latter - but can you update the comment to clarify?

integration-tester/tests/00-infrastructure-setup/infrastructureRunner.js
```diff
@@ -1,14 +1,3 @@
...
-// Because each container listens for connections from the public internet they
-// will need to be placed on separate machines. Therefore, the integration-tester
-// will block until all the machines are ready, since every machine has to be
-// utilized in order for all the containers to boot.
-for (let i = 0; i < infrastructure.nWorker; i += 1) {
-  const container = new kelda.Container('ignoreme', 'google/pause');
-  container.allowFrom(kelda.publicInternet, 80);
-  container.deploy(infra);
-}
+infrastructure.createTestInfrastructure();
```
kayousterhout: "connects" -> "connections" in the commit msg?

integration-tester/tests/00-infrastructure-setup/infrastructure_setup_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+		for _, m := range machines {
+			if m.Status != db.Connected {
+				return false
+			}
+		}
+		return len(machines) == len(bp.Machines)
+	}
+	err = util.BackoffWaitFor(machinesReady, 30*time.Second, 10*time.Minute)
+	assert.NoError(t, err)
```
kayousterhout: does NoError allow you to add an extra message, and if so, can you add a big warning that all of the later tests are expected to fail given that this didn't work?

integration-tester/tests/00-infrastructure-setup/infrastructure_setup_test.go
kklin: Yeah it's the latter

kklin:
And yes I don't think the scheduler should warn if the machines are in the process of booting. It doesn't have that information right now, though, so it'd be part of a larger project to make the change

PR #1309, opened 2017-12-04
google: Refactor the network architecture

ejj:
This patch rewrites most of the Google network related code in service of
several goals.  First, it creates a Google network per region rather than
requiring all regions to share a single network.  This facilitates us,
second, cleaning up the Google network in regions that are unused.  Third,
it generally simplifies the ACL code such that it's more readable, less
redundant, and easier to test.  And finally, it adds unit tests for the
sections of the code that it updates.

### DIFF
--- a/cloud/google/google.go
+++ b/cloud/google/google.go
@@ -1,6 +1,8 @@
 package google
 
 import (
+	"crypto/rand"
+	"encoding/base32"
 	"errors"
 	"fmt"
 	"path"
@@ -16,7 +18,6 @@ import (
 	"github.com/kelda/kelda/join"
 	"github.com/kelda/kelda/util"
 
-	"github.com/satori/go.uuid"
 	log "github.com/sirupsen/logrus"
 	compute "google.golang.org/api/compute/v1"
 )
@@ -144,7 +145,7 @@ func (prvdr *Provider) Boot(bootSet []db.Machine) ([]string, error) {
 			return nil, errors.New("preemptible vms are not implemented")
 		}
 
-		name := "kelda-" + uuid.NewV4().String()
+		name := randName()
 		names = append(names, name)
 
 		go func(m db.Machine) {
@@ -625,6 +626,16 @@ func (prvdr *Provider) createInternalFirewall() error {
 
 func networkURL(networkName string) string {
 	return fmt.Sprintf("global/networks/%s", networkName)
+}
+
+var randName = randNameImpl
+
+func randNameImpl() string {
+	b := make([]byte, 10)
+	if _, err := rand.Read(b); err != nil {
+		panic(err) // This really shouldn't ever happen.
+	}
+	return fmt.Sprintf("k%s", strings.ToLower(base32.StdEncoding.EncodeToString(b)))
 }
 
 func groupACLsByPorts(acls []acl.ACL) map[acl.ACL][]string {
--- a/cloud/google/google.go
+++ b/cloud/google/google.go
@@ -202,7 +202,7 @@ func (prvdr *Provider) operationWait(ops ...*compute.Operation) (err error) {
 			}
 		}
 		return true
-	}, 30*time.Second, 3*time.Minute)
+	}, 10*time.Second, 3*time.Minute)
 }
 
 // Create new GCE instance.
--- a/cloud/google/google.go
+++ b/cloud/google/google.go
@@ -44,11 +44,10 @@ const ipv4Range string = "172.16.0.0/12"
 type Provider struct {
 	client.Client
 
-	networkName string // gce identifier for the network
-	intFW       string // gce internal firewall name
-	zone        string // gce boot region
-
-	ns string // client namespace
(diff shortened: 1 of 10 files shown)


cloud/google/google_test.go
```diff
@@ -111,62 +83,236 @@ func TestListBadNetworkInterface(t *testing.T) {
...
-				{Ports: []string{"1-65535"}},
-			},
-			SourceRanges: []string{"foo"},
-		},
+
+	_, err := gce.parseACL(&compute.Firewall{})
+	assert.EqualError(t, err, "malformed firewall")
+
+	// Missing ports.
+	gce.parseACL(&compute.Firewall{
```
kklin: Did you mean to check the return values? Seems like there's a couple other places in this test where you call `gce.parseACL`, but then check the error returned by `gce.parseACL(&compute.Firewall{})`

cloud/google/google_test.go
```diff
@@ -238,10 +384,53 @@ func TestInstanceConfig(t *testing.T) {
...
-		},
 	}
 
 	assert.Equal(t, exp, res)
 }
+
+func TestCleanup(t *testing.T) {
+	mc, gce := getProvider()
+
+	mc.On("ListNetworks", "network").Return(nil, errors.New("err")).Once()
```
kklin: I think it'd be a bit better to reference `gce.network` rather than "network". If you search for `"network"` there's a lot of reference to the string directly.

cloud/google/google_test.go
```diff
@@ -238,10 +384,53 @@ func TestInstanceConfig(t *testing.T) {
...
+	assert.NoError(t, gce.createNetwork())
+
+	mc.On("ListNetworks", "network").Return(&compute.NetworkList{
+		Items: []*compute.Network{}}, nil)
+	mc.On("InsertNetwork", mock.Anything).Return(nil, errors.New("err")).Once()
+	assert.EqualError(t, gce.createNetwork(), "err")
+
+	mc.On("InsertNetwork",
+		&compute.Network{Name: "network", IPv4Range: ipv4Range}).Return(nil, nil)
+	assert.NoError(t, gce.createNetwork())
```
kklin: I think a `mc.AssertExpectations` would be appropriate for this case

cloud/google/google.go
```diff
@@ -230,163 +228,136 @@ func (prvdr Provider) instanceConfig(name, size, cloudConfig string) *compute.In
...
+	var gacls []gACL
+	for _, a := range acls {
+		ip := strings.Replace(a.CidrIP, ".", "-", -1)
+		ip = strings.Replace(ip, "/", "-", -1)
+		name := fmt.Sprintf("%s-%s-%d-%d", prvdr.network, ip,
+			a.MinPort, a.MaxPort)
+		gacls = append(gacls, gACL{name: name, ACL: a})
+	}
+
+	_, adds, removes := join.HashJoin(aclSlice(gacls), aclSlice(cloudACLs), nil, nil)
```
kklin: Did you consider converting the desired ACLs into `compute.Firewall`s, and joining that with `cloudACLs`? Then you wouldn't need to add the `gACL` struct, and the join would naturally deal with firewall rules that are in a format it doesn't understand (the rules wouldn't match any of the desired rules, so they would get removed)

cloud/google/google.go
```diff
@@ -230,163 +228,136 @@ func (prvdr Provider) instanceConfig(name, size, cloudConfig string) *compute.In
...
-				"%d-%d", acl.MinPort, acl.MaxPort)).
-				Debug("Google: Deleting firewall")
-			op, err = prvdr.DeleteFirewall(fw.Name)
-			if err != nil {
-				return err
-			}
+	var cloudACLs []gACL
+	for _, fw := range cloudFWs {
+		if acl, err := prvdr.parseACL(fw); err != nil {
+			remove = append(remove, fw.Name)
```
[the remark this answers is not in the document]
ejj: good idea

cloud/google/google.go
ejj: I did consider it but I decided against it.  compute.Firewalls are this really complicated nested structure in which a given gACL could have multiple different representations.  Doing so feels a bit fragile to me.

cloud/google/google_test.go
ejj: eh, that's a messy error.  Copy and pasted wrong.  Good catch

cloud/google/google.go
kklin: I see. So for my own understanding, the benefit of doing it the current way is that you're very explicit about the expected representation of the firewalls? (i.e. the parseACL function)

cloud/google/google_test.go
ejj: 👍

cloud/google/google.go
ejj: Yep!  Another way to put it, is I'm confident this way works, while the other way there would have been enough edge cases that it would have been less obviously correct.

PR #1311, opened 2017-12-05
foreman: Fix frequent tick logic

kklin:
The foreman has two separate tickers for interacting with the minion:
when the minion is not connected, it tries frequently, and when it is
connected, it tries less often. This code, however, had a bug where it
created a new `connected` variable within the scope of the outer
`connected` variable, rather than assigning the value to the outer
variable. Therefore, the frequent ticker was always used, even if the
minion was already connected.


kayousterhout:
Is it hard to write a unit test for this?

kklin:
It's possible but I'm not sure the best way. The problem is mocking out these tickers https://github.com/kelda/kelda/blob/master/cloud/foreman/foreman.go#L91. The only options I can think of is passing in the tickers as arguments to the function, or adding a function that gets the ticker, and mocking that out.

PR #1312, opened 2017-12-05
dns: Don't log DNS requests and responses

kklin:
The logs are too noisy with all the request and response logs,
especially at scale.


kklin:
@ejj requesting a review from you since you wrote this initially

PR #1314, opened 2017-12-06
ci: Add Spark integration test that reads from S3

kayousterhout:
This commit adds a new integration test for Spark that reads
data from S3. The test needs S3 access keys, which are added
using Kelda secrets.


integration-tester/tests/21-spark/spark_test.go
```diff
@@ -81,6 +74,40 @@ func TestRunShuffleJob(t *testing.T) {
...
+	}
+	defer clnt.Close()
+
+	// Run a job that reads a file from S3.
+	cmd := exec.Command("kelda", "ssh", "spark-driver", "run-example",
+		fmt.Sprintf("HdfsTest s3a://kelda-spark-test/sonnets.txt"))
+
+	outBytes, err := cmd.CombinedOutput()
+	fmt.Println(string(outBytes))
+	if err != nil {
```
kklin: Is there an easy check we could do on the output to make sure the test ran as expected?

kayousterhout: As discussed in-person there's not a good way to check this

PR #1315, opened 2017-12-06
server: Daemon errors if user runs blueprint with invalid regions

HongjianZhang:
Before the patch, if a user ran a blueprint with unsupported regions,
there wasn't any feedback, and the daemon would do nothing.
This patch checks that the region specified in the blueprint is supported
for a given provider, and errors if it is not.


api/server/server.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/kelda/kelda/api/client"
 	"github.com/kelda/kelda/api/pb"
 	"github.com/kelda/kelda/blueprint"
+	"github.com/kelda/kelda/cloud"
```
kayousterhout: Usually we make the commit message fit in 60 chars (this one looks like it will line wrap) - how about "server: Error for invalid regions"?

api/server/server.go
```diff
@@ -228,6 +238,13 @@ func (s server) Deploy(cts context.Context, deployReq *pb.DeployRequest) (
 		}
 	}
 
+	for _, m := range newBlueprint.Machines {
```
kayousterhout: Since the Javascript code (https://github.com/kelda/kelda/blob/master/js/bindings/bindings.js#L311) ensures that all machines have the same region / provider, I'd write a comment here saying that, and then just check the first machine.  Then, since the code is simpler, I'd inline the checkRegion method.

api/server/server_test.go
```diff
@@ -217,6 +219,29 @@ func TestDeploy(t *testing.T) {
...
+	}, {"Provider":"Amazon",
+		"Role":"Worker",
+		"Size":"m4.large",
+		"Region":"FakeRegion"
+	}]}`
+
+	_, err := s.Deploy(context.Background(),
+		&pb.DeployRequest{Deployment: createMachineDeployment})
+
+	assert.EqualError(t, err, "region: FakeRegion is not supported"+
```
kayousterhout: supernit: for consistency with other places in the code, can you put the space on the first line, after "supported", rather than before "for"?

api/server/server.go
```diff
@@ -228,6 +229,24 @@ func (s server) Deploy(cts context.Context, deployReq *pb.DeployRequest) (
 		}
 	}
 
+	// Since the Javascript code ensures that all machines have the same region
```
kayousterhout: Nit here about the comment: can you instead add a comment here that says "Ensure that the region is valid", and then move your existing comment to below the if-statement?  (That helps some users to skip all of the code in the if-statement if they don't care about the valid region stuff, and also moves the bit about checking the first machine to right where the first machine is grabbed)

kayousterhout:
Merged this - thanks @HongjianZhang!

PR #1316, opened 2017-12-07
bindings: Add support for t2.small instances

kayousterhout:
This commit does not fix the broader problem that we should
support all of the t2 instance types.


kayousterhout:
Thanks for the quick review! I'll submit the broader PR soon but I'm guessing that will require more discussion before merging.

PR #1317, opened 2017-12-07
Update support for t2 instances

kayousterhout:
These commits update Kelda to support all of the t2 instance types that Amazon types.

Because t2 instances have unusual semantics where a CPU is not fully allocated to the instance (instead, the VM is limited to using the CPU during a fixed number of bursts), this PR also changes Kelda to never select T2 instances for a user (when a user just gives CPU/RAM requirements).  This means that, by default, if a user doesn't specify any machine size properties, Kelda will use m3.medium rather than t2.micro instances.  I'm not sure if we should add documentation specifically describing to use t2.micro instances.  I hesitated to do that because Amazon changes things so often that I'm worried that if we describe exactly what's in the free tier, it will change and our documentation will become out of date.


js/bindings/bindings.js
```diff
@@ -732,21 +731,15 @@ class Machine {
...
    * @param {Range} cpu - The desired number of CPUs.
    * @param {Range} ram - The desired amount of RAM in GiB.
    * @returns {void}
    */
   chooseSize(cpu, ram) {
-    if (this.preemptible && this.size !== ''
-        && nonPreemptibleSizes[this.provider].includes(this.size)) {
-      throw new Error(`Requested size ${this.size} can not be preemptible.` +
-        ' Please choose a different size or set preemptible to be False.');
-    }
```
HongjianZhang: What if a user specifies `Preemptible: true` in the blueprint, but specifies a non-preemptible size, `t2.micro` for example. Would kelda still launch the instance?

kayousterhout: Did you see the commit message? (if you click on "Commits", you can look through each of the commits in the PR)

Amazon now allows preemptible t2.micro instances, which is the reason for this change.

HongjianZhang: Oh sorry I missed that detail. What about the other providers? I am not sure if Google or DigitalOcean have non-preemtible sizes.

kayousterhout: This isn't an issue because we don't currently support preemptible instances on either Google or DigitalOcean, but we we do run into the issue with one of those providers (or if/when Amazon changes to add other non-preemptible sizes), we'd add this code back in at that point.

kayousterhout:
Thanks for the review @HongjianZhang!

PR #1318, opened 2017-12-08
docker: Update to latest version of the Docker engine

kklin:


kayousterhout:
I assume there's no way to use a latest tag here?

kklin:
We could install Docker without a pinned version, but I'd prefer not to introduce the failure mode of some incompatible Docker upgrade.

PR #1319, opened 2017-12-11
debug-logs: Collect counter information

kklin:


kklin:
I know @aegamesi has finals so unassigning him

cli/command/counter.go
```diff
@@ -85,10 +88,11 @@ func queryCounters(c client.Client, tgt string) ([]pb.Counter, error) {
 	return c.QueryMinionCounters(ip)
 }
 
-func printCounters(out io.Writer, counters []pb.Counter) {
+func printCounters(out io.Writer, target string, counters []pb.Counter) {
 	w := tabwriter.NewWriter(out, 0, 0, 2, ' ', 0)
 	defer w.Flush()
 
+	fmt.Fprintln(out, target)
```
kayousterhout: Why does this go to `out` rather than `w`?  Is this because you don't need the tab-writing?

cli/command/counter.go
```diff
@@ -30,13 +30,16 @@ counters for a machine, use the machine's ID.`, daemonTarget)
...
 	targets []string
+	all     bool
 
 	connectionHelper
 }
 
 // InstallFlags sets up parsing for command line flags.
 func (cmd *Counters) InstallFlags(flags *flag.FlagSet) {
 	cmd.connectionHelper.InstallFlags(flags)
+	flags.BoolVar(&cmd.all, "all", false,
```
kayousterhout: Why use a flag rather than using `all` as the `target`?

cli/command/counter.go
```diff
@@ -14,20 +14,22 @@ import (
...
 	"github.com/kelda/kelda/util"
+
+	log "github.com/sirupsen/logrus"
 )
 
 const daemonTarget = "daemon"
 
-var counterCommands = "kelda counters [OPTIONS] TARGET"
+var counterCommands = "kelda counters [OPTIONS] TARGETS"
 var counterExplanation = fmt.Sprintf(`Display internal counters tracked for
```
kayousterhout: I was confused by this at first - as long as you're changing things in here how about changing this to
```
Display internal counters tracked for developer debugging purposes.  Most users will not need this command.  TARGETS specifies the targets that counters should be fetched from, and can be a machine's ID or %s for the daemon (the command fetches all counters tracked on the given target).
```

cli/command/counter.go
kklin: I think it's a bit simpler, you don't have to explain cases like `kelda debug-logs daemon all`. Also, `kelda debug-logs` also has an `-all` flag.

cli/command/counter.go
kklin: Thanks, that's much easier to understand!

cli/command/counter.go
kklin: I think it's better to have it go to `w`! I didn't read the existing code very closely.

PR #1320, opened 2017-12-12
ci: Remove load balancer from network test

kklin:
The HPing test sometimes fails against the load balancer because the
return packet's source address does not get rewritten to the load
balancer's IP, so the kernel drops the packet.

This commit disables testing load balancers with HPing because we have
no immediate plans to fix this, and the load-balancer test still covers
the case where there's an actual endpoint behind the load balancer
(the HPing relies on the RST packet generated by the Kernel, and thus
doesn't actually go through the entire TCP protocol).


kklin:
Yeah, I don't see any reason to remove it right now

PR #1321, opened 2017-12-12
cloud: Implement preemptible instances on GCE

HongjianZhang:
Before this patch, preemtible instances were only supported
on AWS. This patch allows preemptible instances to be created
on GCE if the `Preemptible` flag is set to be true in the
blueprint.


kklin:
Can you enable preemptible instances for the integration tests as well? (i.e. modify this https://github.com/kelda/kelda/blob/master/integration-tester/config/infrastructure.js#L47)

cloud/google/google.go
```diff
@@ -113,13 +113,14 @@ func (prvdr *Provider) List() ([]db.Machine, error) {
 		}
 
 		machines = append(machines, db.Machine{
```
aegamesi: nitpick: preemptible is spelled as 'preemtible' in the commit message

integration-tester/config/infrastructure.js
```diff
@@ -44,8 +44,9 @@ function createTestInfrastructure() {
   const workers = [];
   for (let i = 0; i < nWorker; i += 1) {
     const worker = baseMachine.clone();
-    // Preemptible instances are currently only implemented on Amazon.
-    if (baseMachine.provider === 'Amazon') {
+    // Preemptible instances are currently only implemented on
+    // Amazon and Google.
+    if (baseMachine.provider in ['Amazon', 'Google']) {
```
aegamesi: This doesn't actually work in JavaScript -- the 'in' operator checks if a property exists in an object. For arrays, that's also array indices, so like:
`0 in ['Amazon', 'Google'] === true`
`1 in ['Amazon', 'Google'] === true`
`2 in ['Amazon', 'Google'] === false`
`'Amazon' in ['Amazon', 'Google'] === false`
So this will always return false

kklin: Fun fact, Javascript properties are always strings. If you use a non-string key it gets automatically converted into a string. So `'0' in ['Amazon', 'Google']` is also `true`.

HongjianZhang: Interesting. Seems like people are using `indexOf` to check if something is in array.

integration-tester/config/infrastructure.js
```diff
@@ -44,8 +45,9 @@ function createTestInfrastructure() {
   const workers = [];
   for (let i = 0; i < nWorker; i += 1) {
     const worker = baseMachine.clone();
-    // Preemptible instances are currently only implemented on Amazon.
-    if (baseMachine.provider === 'Amazon') {
+    // Preemptible instances are currently only implemented on
+    // Amazon and Google.
+    if (preemptibleProviders.indexOf(baseMachine.provider) > -1) {
```
aegamesi: I think this comment should be in a different place -- but why bother doing the whole array thing? I think it would just be simpler for now to do

`baseMachine.provider === 'Amazon' || baseMachine.provider === 'Google'`

kklin:
Closing this for now, I'll open a new PR addressing @aegamesi's comments

PR #1330, opened 2017-12-13
ci: Write JUnit results in realtime

kklin:
This commit modifies the integration test runner to write the test results
as individual files into a directory, rather than a single large file
containing all the tests. This way, the test result can be written right
after the test has finished, allowing users to view test results before the
entire build has completed.


kklin:
@ejj please review https://github.com/kelda/tester/pull/67 as well. I know you're probably busy -- there's no particular rush for this PR

PR #1331, opened 2017-12-14
Small improvements to the README

luise:


README.md
```diff
@@ -34,6 +37,10 @@ $ npm install -g @kelda/install
 ```
 Check out more in our [Quick Start tutorial](http://docs.kelda.io/#quick-start).
 
+## Example
+For a great example of how Kelda simplifies running complex applications, see
+[how to easily run and manage Spark clusters with Kelda](https://github.com/kelda/spark).
```
kklin: Maybe "see how easy it is to run and manage Spark clusters"? I think it sounds a bit friendlier, but maybe it's too assertive.

luise: No, that's a good idea! I'll update it

PR #1332, opened 2017-12-14
cloud: Run less aggressively in inactive regions

ejj:
This patch gives inactive regions their own database trigger which
ignores the machine table.  Thus, changes to the machine table in
active regions, won't cause cloud lists in all regions of every
provider.  This should hopefully prevent us from bumping into request
limits in digital ocean.


cloud/cloud.go
```diff
@@ -100,17 +100,11 @@ func startClouds(conn db.Conn, ns string, stop chan struct{}) {
 func (cld *cloud) run(stop <-chan struct{}) {
 	log.Debugf("Start Cloud %s", cld)
 
-	// Note that we need to be fairly conservative about running due to request
-	// limits enforced by the cloud providers.  On unused regions we run very rarely
-	// as a result.
```
kayousterhout: Why remove this comment?

PR #1333, opened 2017-12-15
ci: Add integration test for HDFS

kayousterhout:
The new integration test writes a file to HDFS, reads it back,
and makes sure that the result is the same as the original file.


integration-tester/tests/hdfs/hdfs.js
```diff
@@ -0,0 +1,8 @@
+const hadoop = require('@kelda/hadoop');
+const infrastructure = require('../../config/infrastructure.js');
+
+const hdfs = new hadoop.HDFS(infrastructure.nWorker - 1);
+
+const infra = infrastructure.createTestInfrastructure();
+hdfs.deploy(infra);
+
```
kklin: Extra newline

integration-tester/tests/hdfs/hdfs_test.go
```diff
@@ -0,0 +1,50 @@
...
+	// Location in HDFS to write to.
+	hdfsFilepath := "/README.txt"
+
+	// Location of the file to write.  Use the Hadoop readme because it is a
+	// text file that conveniently already exists on the container.
+	readmeLocalFilepath := "/hadoop/README.txt"
+
+	// Write the Hadoop readme to a location in HDFS.
+	writeCmd := exec.Command("kelda", "ssh", "hdfs-namenode", "hdfs",
+		"dfs", "-copyFromLocal", "/hadoop/README.txt", hdfsFilepath)
```
kklin: Did you mean to use `readmeLocalFilepath` here?

integration-tester/tests/hdfs/hdfs_test.go
```diff
@@ -0,0 +1,50 @@
...
+	outBytes, err := writeCmd.CombinedOutput()
+	fmt.Println(string(outBytes))
```
kklin: since you're printing two outputs, would it make the output easier to read if we logged something like "Writing file to HDFS...". Or is the command output sufficient context?

integration-tester/tests/hdfs/hdfs.js
kayousterhout: Ahhh my signature :)

kayousterhout:
Thanks for the quick review @kklin!

PR #1337, opened 2017-12-26
ci: Add Let's Encrypt integration test

aegamesi:
This commit adds a new integration test that fetches TLS certificates
from the Let's Encrypt staging server using the haproxy_auto_https
blueprint.

This requires six subdomains to be set up (according to `integration-tester/config/infrastructure.js`):
* ci-amazon-a.kelda.io: 13.57.99.49
* ci-amazon-b.kelda.io: 13.57.99.49
* ci-google-a.kelda.io: 104.196.11.66
* ci-google-b.kelda.io: 104.196.11.66
* ci-digitalocean-a.kelda.io: 138.68.203.188
* ci-digitalocean-b.kelda.io: 138.68.203.188

This also, naturally, requires the haproxy_auto_https blueprint PR to be merged (https://github.com/kelda/haproxy_auto_https/pull/1), as well as the associated Docker image to be pushed to Dockerhub.


integration-tester/tests/lets-encrypt/lets_encrypt_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Determine the correct domain names to use based on the provider.
+	provider := getProvider(t, machines)
+	domainA := "ci-" + provider + "-a.kelda.io"
+	domainB := "ci-" + provider + "-b.kelda.io"
+
+	// Create a custom HTTP client that supports the Let's Encrypt Staging CA.
+	certpool := x509.NewCertPool()
+	certpool.AppendCertsFromPEM([]byte(fakeLERootCAPem))
+	tlsConfig := &tls.Config{
```
kklin: If I'm understanding correctly, the test doesn't actually ensure that the domain is using a HTTPS certificate generated by Let's Encrypt. Can you either add a comment saying why this test is sufficient to make sure the ACME generation is working (e.g. because the URL uses the `https` protocol, the Go HTTP client is definitely going through the TLS handshake, so the server certificate is either signed by a certificate trusted by the system, or the Let's Encrypt CA), or test directly that the server's certificate is signed by ACME?

BTW I'm not actually sure whether my example comment is true 😄

aegamesi: Yeah -- that's how it works. It's actually stronger than that though -- the only root CA allowed is the Let's Encrypt Staging CA (not even regularly trusted ones).

kklin: Awesome

aegamesi:
I also made the test more to the point -- I got rid of the whole 15 requests thing to check if all of the expected responses were returned -- that kind of behavior is tested in the regular haproxy integration test (and that part of the haproxy config is identical). 

The blueprint takes a couple of seconds (maybe 15 at most if everything is working correctly) to fetch the certificate, even after the containers are up. I'm assuming this would probably be covered by the waiting that the integration tester does anyway, but if not, it would fail.

I'm not sure what the best way to handle this is. Maybe keep making requests (as long as, say, 60 seconds have not passed since the test began), and only fail if *none* of them succeed?

kklin:
The integration tester waits until the containers are up, then sleeps an additional amount of time before starting the test (90 seconds I think?). If you're worried about that being an insufficient amount of time, then I think your suggest of trying multiple times is good. But it sounds like for now we can just add that later if it ends up failing in our test runs

kklin:
@aegamesi How did you generate the floating IP for digitalocean? I don't see it under https://cloud.digitalocean.com/networking/floating_ips?i=c48e98

For the Amazon and Google floating IP, I'll need to generate an IP for the CI accounts since floating IPs are scoped to users and projects, respectively.

kklin:
I allocated 35.202.229.231 on Google, and 35.202.229.231 on Amazon.

aegamesi:
I think you have to press the link that says 'To reserve a floating IP for a specific datacenter, *click here*'.

Why did you reserve additional floating IPs though? They should just be the ones defined in the testing infrastructure.

kklin:
As discussed on Slack, I was confused 😄

kklin:
Added the DNS entries!

Issue #1327, opened 2017-12-13
Make Kelda's DNS server case insensitive


luise:
Kelda's DNS server should be case insensitive, because many applications assume this is the case -- for instance, Node.js lowercases hostnames before attempting to resolve them in DNS. This means that any hostnames with capital letters will fail to resolve when given to a Node app. Kelda's DNS server code is in `minion/network/dns.go`.

kklin:
I've written a simple blueprint that shows this problem:

```
// This blueprint deploys a container that prints "Pass" if it successfully
// resolves a Kelda hostname with a mutated case.
const kelda = require('kelda');

const machine = new kelda.Machine({ provider: 'Amazon' });
const infra = new kelda.Infrastructure(machine, machine);

const container = new kelda.Container('exampleHostnameWithCaps', 'alpine');

const lowerCaseHostname = container.getHostname().toLowerCase();
if (container.getHostname() === lowerCaseHostname) {
  console.error("The test hostname is already lowercase.")
  process.exit(1);
}

container.command = [
  'sh', '-c',
    `if getent hosts ${lowerCaseHostname}; then ` +
      'echo Pass; ' +
    'else ' +
      'echo Fail; ' +
    'fi;' +
    'tail -f /dev/null',
];
container.deploy(infra);
```

Right now, running the blueprint then running `kelda logs exampleHostnameWithCaps` (to get the output of the test container) shows "Fail". If this issue is fixed, the logs should have "Pass".

What the blueprint is doing is creating a container whose hostname has capital letters, but then resolving the hostname to an IP with an all-lowercase version of the hostname.

Ch3ck:
I'll clone and code and try to build then I'll look at the files with the logs.

Ch3ck:
@kklin You can close this now.

PR #1340, opened 2018-01-04
minion/network:	Make Kelda's DNS server case insensitive.

Ch3ck:
* Kelda's DNS server should be case insensitive, converting hostnames to lower cases before atempting to resolve them in DNS.
* Fixes issue #1327


minion/network/dns.go
```diff
@@ -28,6 +28,8 @@ var table *dnsTable
 
 var dnsC = counter.New("Network Dns")
 
+var lookupHost = net.LookupHost
```
kklin: Moving this line of code isn't really relevant to the patch, and will change the Git history unnecessarily. Can you leave it as it was before?

kklin: Also, in the same vein of making the Git history cleaner, can you:
- Squash your three commits into a single commit. (`git log` is much easier to read if every commit is isolated, and it's easier to find a single commit that introduces a bug)
- This is really nitpicky, but I would make a couple small changes to your commit message for consistency with the rest of our messages. Again, this just makes `git log` a bit nicer to read:
    - Change "Made" to "Make". ([This](https://chris.beams.io/posts/git-commit/#imperative) post has a good explanation on why we prefer the imperative mood)
    - Remove the trailing period in the commit title (after "insensitive")
    - "converting hostnames to _lower cases_" should be "converting hostnames to _lowercase_"

minion/network/dns.go
```diff
@@ -246,6 +249,7 @@ func hostnamesToDNS(hostnames []db.Hostname) map[string]net.IP {
 	records := map[string]net.IP{}
 	for _, hn := range hostnames {
 		if ip := net.ParseIP(hn.IP); ip != nil {
+			hn.Hostname = strings.ToLower(hn.Hostname)
```
kklin: Can you add a unit test for this PR? I think adding a case to the `TestLookupA` test would be sufficient

minion/network/dns.go
Ch3ck: Let me make the necessary fixes.

minion/network/dns.go
Ch3ck: Let me do that now.

Ch3ck:
@kklin Issues fixed

integration-tester/tests/10-network/dns.js
```diff
@@ -0,0 +1,25 @@
+// This blueprint deploys a container that prints "Pass" if it successfully
```
kklin: Sorry, I meant adding a Go unit test here https://github.com/kelda/kelda/blob/master/minion/network/dns_test.go#L89. The integration tests require more setup, and I don't think it's worth it for this minor change.

Ch3ck: Yeah, that's true, let me do that real quick

Ch3ck: @kklin Done!

kklin:
Thanks @Ch3ck!

luise:
Thanks @Ch3ck !

Ch3ck:
Onto the next one :-)

PR #1354, opened 2018-01-24
bindings: Use allowTraffic instead of allowFrom/allow

aegamesi:
This change updates this blueprint to use the new allowTraffic method
instead of the now deprecated allowFrom and allow methods.


kklin:
I think there's an outdated use of the API in the README that needs to be updated

luise:
It would also be awesome if you could update the examples in the docs!

kklin:
This PR should fix the part of the build that's failing because of Django https://github.com/kelda/kelda/pull/1358

aegamesi:
@kklin I've fixed all of the non-code references to allowFrom

kklin:
@aegamesi build is still failing because you omitted the dest here  https://github.com/kelda/kelda/pull/1354/files#diff-cd4719279a050dcc6fbd6f4afe3cb8a9R51

aegamesi:
oof, fixed @kklin

PR #1359, opened 2018-01-25
Improve DigitalOcean default region handling

kklin:
Booting machines in DigitalOcean without specifying a region is not working
currently.


docs/_FAQ.md
```diff
@@ -53,3 +53,14 @@ container on a Kelda worker:
...
 [VPC section of the online AWS console](http://console.aws.amazon.com/vpc).
+
+### My DigitalOcean machines are failing to boot with the error "422 Region is not available"
+DigitalOcean sometimes temporarily disables Droplet creations. Trying a
+different region will most likely resolve this problem.
+
+To confirm that the error is because Droplet creations are disabled for your
+region, go the [Droplet creation
+page](https://cloud.digitalocean.com/droplets/new) on the DigitalOcean UI. If
+DigitalOcean has disabled your desired region, hovering over the region under
```
luise: This is a nit, but I wonder if it's better to be less specific, so our docs aren't outdated when DO e.g change their message. We could e.g. say something like "... on the DigitalOcean UI. Under "Choose a datacenter region", you can see if any of the regions are unavailable".

kklin: That sounds good, it flows better too!

kklin:
@luise addressed your comments, and fixed the failing tests

PR #1362, opened 2018-01-26
mock_docker: Check if image exists when pushing

kklin:
The previous mock Docker implementation did not require that images were
built before pushing.


kklin:
OK @luise, my testing is done. This PR is _officially_ assigned to you 😄

luise:
@quilt-bot (which should totally be renamed?!) never doubted it was my PR 😀

PR #1367, opened 2018-01-30
bindings: Restrict allowable hostnames

kklin:
Before, we did no validation on hostnames. A bad hostname can introduce
all sorts of bugs: for example, if uppercase hostnames were allowed, two
hostnames that only differed in capitalizations would lead to
non-determinism in DNS lookups.

This commit adopts the same hostname definition used by Kubernetes.
Specifically, hostnames must not be longer than 253 characters, and can
only container lowercase letters, numbers, and hyphens.


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Kelda Change Log
 Up Next
```
aegamesi: Commit message:

'container' should be 'contains', and I personally think 'differ only in capitalizations' should drop the 's' on capitalization

js/bindings/bindings.js
```diff
@@ -1460,6 +1463,23 @@ function getInfrastructure() {
...
 
+/**
+ * validateHostname checks whether the given hostname is a valid hostname.
+ * If the hostname is invalid, it throws an error.
+ *
+ * @param {string} hostname - The hostname to validate.
+ * @returns {void}
+ */
+function validateHostname(hostname) {
+  const regexp = new RegExp('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$');
```
aegamesi: Is not starting with/ending with hyphens also a restriction? If so, I think that should be documented somewhere. Is `a------b` a valid hostname? This regex seems a little strange.

CHANGELOG.md
kklin: Ha.. I actually make the "contain" -> "container" typo a lot. I'm really used to typing "container"

js/bindings/bindings.js
kklin: I added a note to the error message about not starting/ending with hyphens.

The regex comes from Kubernetes, so I'm inclined to just leave it the same

kklin:
@aegamesi do you want to take another look at this?

PR #1368, opened 2018-01-30
bindings: Remove deprecated methods allowFrom and allow

aegamesi:
These methods were replaced with allowTraffic and deprecated in an
earlier release. This change removes them and their associated tests.


kklin:
Our kelda-corp deployment should be updated before this is merged. I'll handle it.

kklin:
@aegamesi looks like there's some merge conflicts now

aegamesi:
@kklin fixed merge conflicts

PR #1374, opened 2018-02-02
dns: Remove .q TLD from Kelda hostnames

kklin:


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ this feature should query their public IP at runtime using a service such as
 checkip.amazon.com, or pass containers a floating IP (or DNS name backed by a
 floating IP) and use placement rules to ensure the container is scheduled at
 that floating IP.
+- Remove the `.q` TLD for Kelda hostnames. Kubernetes doesn't support hostnames
```
luise: This will break the HAProxy unit test sthat checks the config: https://github.com/kelda/haproxy/blob/master/test/haproxy-test.js and https://github.com/kelda/haproxy_auto_https/blob/master/test/haproxy-test.js. Not sure if it should work with master or latest release.

luise: The Spark master URL in this example will also be out of date: https://github.com/kelda/spark#run-jobs.

kklin: Good catch! I'll submit PRs for them now, but they shouldn't be merged until the next release is made.

How did you find these references? Wondering what else I might have missed..

luise: Awesome! I just thought of the HAProxy one, which made me look around a little more. I ran a `grep` that wasn't ideal, but helped narrow things down (fortunately there aren't many other words in our code that start with q!):

```
grep -rIF --exclude-dir=node_modules --exclude-dir=vendor --exclude-dir=integration-tester/node_modules --exclude-dir=docs ".q" 
```

I ran something similar in my blueprints directory, but that's not bullet proof since some of my local clones are out of date with upstream -- did help me catch the spark one, though!

js/bindings/bindings.js
```diff
@@ -343,7 +343,7 @@ class LoadBalancer {
    * @returns {string} The Kelda hostname that represents the entire load balancer.
    */
   hostname() {
```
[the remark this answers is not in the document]
kklin: I personally prefer using functions since it makes the documentation easier to read IMO. Functions are more obviously public-facing than member variables (I don't think our documentation even includes the member variables right now), but that could just be personal preference.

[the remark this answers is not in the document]
luise: That's fine with me too. I thought I remembered you saying that you prefer accessing the properties directly as opposed to through methods (but after looking up that conversation, you were talking about setting properties, not getting them).

I'm fine with keeping them, but we should definitely make sure that they're called the same. It's confusing that one is `getHostname()` and the other is `hostname()`. Unless there is a good reason?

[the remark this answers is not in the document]
kklin: No, I agree that their names should be consistent

CHANGELOG.md
kklin: Nice! I didn't know about the --exclude-dir flag.

PR #1381, opened 2018-02-05
docker: Upgrade engine to latest stable release

kklin:
This will hopefully fix a problem with `docker exec` hanging in the
Kubernetes rewrite.


kklin:
Yep, it passes the network and mean integration tests. Don't know if it'll fix the exec bug though since it's a bit hard to repro.

PR #1383, opened 2018-02-06
debug-logs: Don't exit prematurely if we fail to query cluster

kklin:
Before, if we were unable to fetch the running machines or containers,
we exited without collecting any logs at all. However, there are many
times when there might be useful logs to be collected. For example, we
might be unable to query the current containers because the master
minion crashed, in which case the machine logs would be very useful.


aegamesi:
Any particular reason you moved the
`ipMap := map[string]string{}`
stuff from below fetching the machines to below fetching the containers?

kklin:
I thought it was a bit easier to follow to group the `numQueryErrors` logic close together, and that having the `ipMap` stuff close wasn't adding much.

What do you think?

aegamesi:
Yeah, I think that's reasonable.

PR #1387, opened 2018-02-10
Allow constructors to take a single object with all arguments

luise:
This PR lets the Container, Infrastructure, Image, and LoadBalancer constructors take all arguments as a single object rather than individual args, e.g.: 

```javascript
const container = new Container({
  name: 'my-app',
  image: 'nginx',
  filepathToContent: { '/etc/myconf': 'foo' },
});
```

rather than
```javascript
const container = new Container(
  'my-app', 'nginx', {filepathToContent: {'/etc/myconf': 'foo'}});
```

The main goal of this is to improve readability. 

For backwards compatibility, the constructors also allow the old way of passing arguments. When I've updated all the blueprints to use the new constructor style from this PR, I will make another PR to remove the ability to pass individual arguments (@kklin will make the appropriate releases).


kklin:
I'll review this since I want it in ASAP

luise:
Cool, thanks! I’ll submit the updated blueprints sometime over the weekend. Probably also the PR that removes the ability to pass individual atguments.

kklin:
What's the thing about individual arguments?

kklin:
Oh, never mind.

js/bindings/bindings.js
```diff
@@ -51,17 +70,26 @@ class Infrastructure {
...
+    // If the masters is an object (but neither a Machine nor an array of Machines),
+    // we assume this object contains all the user's arguments.
+    if ((typeof masters === 'object') && !(masters instanceof Machine) && !(Array.isArray(masters))) {
+      Object.assign(args, masters);
+    } else {
+      Object.assign(args, opts, { masters, workers });
+    }
 
-    this.machines = [];
+    this.adminACL = getStringArray('adminACL', args.adminACL);
```
kklin: I learned this from @simplygreatwork -- he basically assigned the args object into `this`. This way, the names would be guaranteed to be consistent, and it's just less boilerplate. But it doesn't quite translate to our code since we have extra validation like `checkExtraKeys`. Just mentioning because I thought it was cool -- maybe something to do once this is merged.

js/bindings/bindings.js
```diff
@@ -77,9 +105,14 @@ class Infrastructure {
...
     boxedMasters.forEach(master =>
-      this.machines.push(machineWithRole(master, 'Master')));
+      this.masters.push(machineWithRole(master, 'Master')));
     boxedWorkers.forEach(worker =>
-      this.machines.push(machineWithRole(worker, 'Worker')));
+      this.workers.push(machineWithRole(worker, 'Worker')));
+    // We set the masters and workers as separate properties of the Infrastructure
+    // to satisfy `checkExtraKeys`. It has the additional benefit that the user
+    // can easily access the worker machines in their blueprint, which is often
+    // done to determine how many workers are available.
```
kklin: So this will behave unexpectedly now.. Not sure if we care though.
```
const baseMachine = new Machine({ provider: 'Amazon' });
const infra = new Infrastructure({ masters: baseMachine, workers: baseMachine });

// Add another worker.
infra.workers.push(baseMachine);
```

js/bindings/bindings.js
luise: Ah, yeah, that's nice! As you say, we'd have to update the `checkExtraKeys`, and we would also have to do some validation on the arguments. But definitely worth thinking about for the future!

js/bindings/bindings.js
luise: That's a good point. I think that is likely to confuse someone. I'll fix it!

js/bindings/bindings.js
```diff
@@ -983,37 +991,44 @@ class Container {
    *   the container using the new files.  Files are installed with permissions
    *   0644 and parent directories are automatically created.
    */
-  constructor(hostnamePrefix, image, opts = {}) {
+  constructor(name, image, opts = {}) {
```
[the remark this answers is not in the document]
luise: Ah, yes, that's a great point. That will ve very helpful.

js/bindings/bindingsTest.js
```diff
@@ -453,13 +453,36 @@ describe('Bindings', () => {
...
+    it('errors when passed invalid optional arguments - object', () => {
+      expect(() => new b.Container({ name: 'host', image: 'image', badArg: 'foo' })).to
+        .throw('Unrecognized keys passed to Container constructor: badArg');
+      expect(() => new b.Container({
+        name: 'host',
+        image: 'image',
+        badArg: 'foo',
+        command: [],
+        alsoBad: 'bar' }))
+        .to.throw('Unrecognized keys passed to Container constructor: ');
```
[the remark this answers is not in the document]
luise: I think they were left out originally because the order of the bad arguments in the error message was arbitrary.  When I run the tests now though, the order seems to be deterministic, so I'll add them back in.

luise:
@kklin , ready for another review

js/bindings/bindings.js
```diff
@@ -70,16 +98,14 @@ class Infrastructure {
...
-      copy.role = role;
-      return copy;
-    };
-
+    // It is now redundant to set the roles and to concatenate the lists here
+    // since we do that in `toKeldaRepresentation`. However, for backwards
+    // compatibility, we keep this functionality until the next release.
     boxedMasters.forEach(master =>
-      this.machines.push(machineWithRole(master, 'Master')));
+      this.masters.push(machineWithRole(master, 'Master')));
```
kklin: I don't think this will work now, although I think it'll work once you remove the old style (i.e. remove this code):
```
const baseMachine = new kelda.Machine({ provider: 'Amazon' });
const infra = new kelda.Infrastructure({ masters: baseMachine, workers: baseMachine });
baseMachine.size = 'm3.medium'
```

It might be confusing if it _did_ work though. Dunno. Just wanted to point it out for you to think about since it's some subtle behavior.

luise: Thank for pointing that out. But that actually also doesn't work with the current constructors, so I'd say it's fine. As you say, it will work in the next version, which might be confusing in some cases, but since it's just a consequence of objects and passing by reference, I'd say it's fine -- we're not really causing the confusion.

kklin:
The commit you just pushed LGTM

luise:
Oh, that was fast! I was just about to message you :)

kklin:
GitHub emailed me about it :)

kklin:
BTW, just realized that some of this stuff could go in the CHANGELOG. Up to you what format you want to do it in (e.g. add a new entry for every commit, or tack on a single commit that summarizes all the changes).

luise:
Ah, good point. I'll add it!

kklin:
LGTM

luise:
Awesome, thanks for the quick review! I'll merge when the build passes.

PR #1389, opened 2018-02-12
ci: Create one SSH connection per machine

kklin:
Before, the parallel SSH utility made a pool of SSH TCP connections.
However, the SSH protocol itself has a concept of multiplexing: a single
TCP connection can have open multiple sessions. This commit modifies our
SSH utility so that it only creates a single connection to each machine,
reusing that connection to create sessions as needed for callers.

This approach reduces the overhead of creating multiple SSH connections. It
also better handles callers that don't fully utilize the connections.
Before, SSH connections were made even if they weren't used -- this was a
bit wasteful, and caused scary errors when running locally since it
stressed the local SSH agent.

Note that this blocks merging the Kubernetes rewrite.


integration-tester/util/ssh.go
```diff
@@ -21,44 +21,38 @@ const maxSSHSessions = 512
 // machines and containers in order to figure out the container's public IP
 // address.
 //
-// 3. It rate limits the number of parallel connections and session creations
-// to avoid overloading the host.
+// 3. It rate limits the number of parallel sessions to avoid overloading the
+// remote host.
 type SSHUtil struct {
```
ejj: I think it would be a bit cleaner to make an `SSHClient` struct that contains an `ssh.Client` and a `chan struct{}`, and then have `type SSHUtil map[string]SSHClient`.

kklin: Ooh nice

PR #1392, opened 2018-02-14
[RFC] Run on top of Kubernetes

kklin:
For the RFC, all commits but the `kube: Kubernetes WIP` commit can be ignored.

A lot of code has changed, but basically the only important stuff is in the `minion/kubernetes` package. I'm looking for feedback on the implementation there. I'll leave comments in the PR about the parts I'm most unsure about.

I haven't done my own final read through of the code, so there might be smaller style/cleanliness things I've yet to fix up (so feel free to ignore issues that seem like I'll fix in my pass). I'm opening it now, though, so that the overall architecture can marinate.


minion/kubernetes/deployment.go
```diff
@@ -0,0 +1,340 @@
...
+func updateDeployments(conn db.Conn, deploymentsClient appsclient.DeploymentInterface,
+	secretClient SecretClient) {
+
+	currentDeployments, err := deploymentsClient.List(metav1.ListOptions{})
+	if err != nil {
+		log.WithError(err).Error("Failed to list current deployments")
+		return
+	}
+
+	key := func(intf interface{}) interface{} {
```
kklin: I'm conflicted on whether we should call `deploymentsClient.Update` for every pair, and let kube decide whether the "Update" actually reflects a change, or if we should decide for ourselves if the deployments need changing, and call `Create` and `Delete` when necessary.

The upside of just always calling `Update` is that this code is simpler. We would have to account for a lot of attributes if we did the join ourselves (e.g. affinities, volumes, container attributes..).

The downside is that we have to be very careful that the deployments we pass to `Update` is consistent. For example, if the order of environment variables change between calls, Kube might treat that as a reason to restart the containers.

kklin:
BTW, the diff is big enough that GitHub is collapsing some non-trivial diffs (i.e. not auto-generated code).

blueprint/blueprint.go
```diff
@@ -186,6 +186,24 @@ func NewString(str string) ContainerValue {
 	return ContainerValue{str}
 }
 
+// SortContainerValues divides a map of ContainerValues to two maps -- one of
```
ejj: "into two maps"

ejj: Also sort isn't a great name for this.  Makes me think of quicksort.

minion/etcd/container.go
```diff
@@ -91,7 +82,6 @@ func joinContainers(view db.Database, etcdDBCs []db.Container) {
 			IP                string
 			BlueprintID       string
 			Image             string
-			ImageID           string
```
ejj: Sigh, it's really too bad we need the containers in Etcd still at all ... I hope we can get rid of it once the networking stuff is improved.

minion/kubernetes/affinity.go
```diff
@@ -0,0 +1,166 @@
...
+	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/retry"
+)
+
+const providerKey = "kelda.io/host.provider"
+const regionKey = "kelda.io/host.region"
+const sizeKey = "kelda.io/host.size"
+const floatingIPKey = "kelda.io/host.floatingIP"
+
+func toAffinities(placements []db.Placement) map[string]*corev1.Affinity {
```
ejj: I think this function could benefit from a high level comment on why transitioning between placements and affinity is so hard.  I.E. how are the concepts different?

ejj: Also, should we be shifting our concept of placement to look more like Kubernetes concept of affinity?  It's not used very much at the moment, so soon would be the time to do it...

minion/kubernetes/affinity.go
```diff
@@ -0,0 +1,166 @@
...
+	// Sort the placements so the generated affinities will be consistent.
+	// Otherwise, Kubernetes might treat a difference in affinity term
+	// orderings as a different deployment spec, and restart the existing
+	// deployment unnecessarily.
+	sort.Sort(db.PlacementSlice(placements))
+
+	result := map[string]*corev1.Affinity{}
+	for _, plcm := range placements {
```
ejj: I might make a function that converts one placement to one affinity just to break this up a bit and make it easier to read (and test likely).

minion/kubernetes/affinity.go
```diff
@@ -0,0 +1,166 @@
...
+							hostnameKey: plcm.OtherContainer,
+						},
+					},
+					TopologyKey: "kubernetes.io/hostname",
+				}
+				podAffinities := &affinity.PodAntiAffinity.
+					RequiredDuringSchedulingIgnoredDuringExecution
+				*podAffinities = append(*podAffinities, req)
+			} else {
+				// XXX: This is not difficult to implement with Kubernetes, but
```
ejj: I think this is important enough that we should file an issue on it.  Spark, for example, could take advantage of it (I'm pretty sure the job scheduler thing needs to be near the master or something).

Btw, the only reason it wasn't done before, is cause it was hard.

minion/kubernetes/affinity.go
```diff
@@ -0,0 +1,166 @@
...
+			matchExpressions := &affinity.NodeAffinity.
+				RequiredDuringSchedulingIgnoredDuringExecution.
+				NodeSelectorTerms[0].MatchExpressions
+			*matchExpressions = append(*matchExpressions, req)
+		}
+	}
+	return result
+}
+
+func updateNodeLabels(nodes []db.Minion, nodesClient clientv1.NodeInterface) {
```
ejj: Comment on this one too

PR #1395, opened 2018-02-18
docs: Update README

luise:
This commit updates the README to better reflect the current state of
the project, and to have less text and more code.


README.md
```diff
@@ -1,119 +1,179 @@
...
+# What is Kelda?
 
-To take this knowledge into production, simply `kelda run` the JavaScript blueprint of
-your application. Kelda will set up virtual machines, configure a secure network, install
-containers, and whatever else is needed to get up and running smoothly on your favorite
-cloud provider.
+Kelda is a simple way to configure, run, and manage containerized applications in the cloud.
 
-*Kelda is currently in beta.*
+No more YAML, templating, complicated CLIs, and new domain specific languages. Kelda’s intuitive JavaScript API makes it easy to describe the systems you want to run in the cloud — that is, the virtual machines, containers, network policies, and more.
```
ejj: "describe the systems", I almost want "describe what" instead.  Feels slightly less jargony to me.

README.md
```diff
@@ -1,119 +1,179 @@
...
 
-## Deploy Quickly on...
+Just pass this high-level blueprint to Kelda’s Kubernetes-based deployment platform using the `kelda` CLI, and we take care of the nitty gritty details needed to get your system up and running in your cloud of choice.
```
ejj: "Just pass this" -> "Just pass a"

ejj: Also this sentence feels just a tad clunky.  Wonder if it could be broken up into a couple of shorter sentences.

README.md
```diff
@@ -1,119 +1,179 @@
...
 
-![providers](./docs/source/images/providers.png)
+In short: *Describe with JavaScript. Deploy from the command line.*
```
ejj: I dunno, you really like this tag line, I don't think it adds much ...  Feels like it's struggling to sneak in here.

README.md
```diff
@@ -1,119 +1,179 @@
...
-Run any container.
+![providers](./docs/source/images/providers.png)
 
-[//]: # (b1)
-<!-- const {Container, LoadBalancer, Machine, allowTraffic, publicInternet} = require('kelda'); -->
-```javascript
-let web = new Container('web', 'someNodejsImage');
-```
+# How Do I Use It?
+1. **Describe your application using Kelda's JavaScript API.**
```
ejj: Not suggesting this in this version, but random thought:

It would be awesome if we had a list of blueprints, and could say something like "Or use one of these pre-made ones".

This kind of makes it sound like you've got to write the blueprint from scratch, which may be a bit intimidating.

README.md
```diff
@@ -1,119 +1,179 @@
...
+    // A Docker container running your application.
+    const app = new kelda.Container({
+      name: 'my-app',
+      image: 'myDockerImage',
+    });
+
+    // Expose the application to the public internet on port 80.
+    kelda.allowTraffic(kelda.publicInternet, app, 80);
+
+    ...
```
ejj: More that I think about it, I dunno that the ellipsis in these examples are necessary.  I think people generally expect that code snippets are fragments of a larger file.

README.md
```diff
@@ -1,119 +1,179 @@
...
+
+    CONTAINER       MACHINE         COMMAND          HOSTNAME    STATUS     CREATED           PUBLIC IP
+    bce2362f3a1f    i-0b6ef239c7                     my-app      running    20 seconds ago    54.218.19.34:80
+    ...
+    ```
+
+    Get the logs of a container or machine (here, the `my-app` container).
+    ```bash
+    $ kelda logs my-app
```
ejj: Before this I would show `kelda ssh` too.  I think that's way cooler than logs.

README.md
```diff
@@ -1,119 +1,179 @@
...
+  provider: 'Google', // 'Amazon', 'DigitalOcean'.
+  size: 'n1-standard-2', // 'm3.medium', '1gb'.
+});
 ```
+See the full [API reference](http://docs.kelda.io/#kelda-js-api-documentation) in our docs.
 
-This is just a small sample of the Kelda CLI. [Check out more handy commands](http://docs.kelda.io/#kelda-cli) for managing your deployment.
-
-## Get Started
+# Examples
```
ejj: Might say something like "Reference Blueprints" instead.  Examples feels like they're toys we aren't intending people actually use.

README.md
luise: I'm just paranoid of being another cryptic project that makes it impossible to understand what it actually does.  But I think the above description and code steps below should make it clear. I'll remove this!

README.md
luise: Yeah, that's a good point. I'll probably just show `ssh` instead of `logs`.

README.md
luise: Hmm, yeah, I guess that makes sense. Wasn't sure if it was clear, but I guess it is (especially because `app` is referenced in both snippets).

luise:
@ejj , I updated it. Do you want to take another look?

PR #1396, opened 2018-02-18
Fully migrate to object literal constructor arguments

luise:
This PR fully migrates to only allowing object literal notation for passing arguments to the Container, LoadBalancer, Infrastructure, and Image constructors.


js/bindings/bindings.js
```diff
@@ -504,6 +504,10 @@ function assertArrayOfType(name, array, type) {
  * @returns {void}
  */
 function checkRequiredArguments(name, object, keys) {
+  if (!(typeof object === 'object') || Array.isArray(object) || (object === null)) {
```
kklin: Could `!(typeof object === 'object')` just be `typeof object != 'object'`?

js/bindings/bindings.js
```diff
@@ -93,14 +89,10 @@ class Infrastructure {
...
 
-    // It is now redundant to set the roles and to concatenate the lists here
-    // since we do that in `toKeldaRepresentation`. However, for backwards
-    // compatibility, we keep this functionality until the next release.
-    boxedMasters.forEach(master =>
-      this.masters.push(machineWithRole(master, 'Master')));
-    boxedWorkers.forEach(worker =>
-      this.workers.push(machineWithRole(worker, 'Worker')));
-    this.machines = this.masters.concat(this.workers);
+    this.masters = boxedMasters;
```
kklin: It'd be a bit shorter to just assigned to `this.masters` rather than `boxedMasters`

js/bindings/bindings.js
luise: Definitely!

PR #1400, opened 2018-02-24
kelda: Replace scheduler with Kubernetes

kklin:
Before, we had a custom scheduler for ensuring the desired containers
were running. While it is not too difficult to implement a basic
scheduler, moving to Kubernetes will allow us to iterate more quickly on
features such as volumes since Kubernetes already implements much of the
logic for actually creating and managing volumes -- we will be able to
just focus on how to express volumes in the blueprint language.

Furthermore, this change allows us to potentially, in the future, run on
top of an existing Kubernetes cluster. This will be helpful given the
general movement in industry towards using Kubernetes.

At a high level, this commit encompasses four major changes:
1) It modifies the deployment engine to boot a Kubernetes cluster on the
Kelda machines.
2) It replaces the custom scheduler with code that translates the
containers specified by the user into deployments to be enforced by the
Kubernetes cluster.
3) It replaces the Docker network plugin with a CNI network plugin
compatible with Kubernetes.
4) It replaces Vault in favor of the built-in Kubernetes solution for
secret management. It's simpler to just use the Kubernetes solution, and
we weren't making use of any of the more advanced Vault features.


kklin:
@ejj I left a couple small documentation TODOs for you. You can find them by searching for "TODO Ethan". There were also some TODOs that you left from your CNI work that I don't know if you want to address.

ejj:
Are there any dependencies we can remove from the vendor directory?

kklin:
FYI @ejj this will change a bit to support the volumes/privileged stuff that was merged yesterday. Up to you how you want to review it

cli/command/debug.go
```diff
@@ -170,33 +175,38 @@ func (dCmd Debug) Run() int {
 	// we might be unable to query the current containers because the master
 	// minion crashed, in which case the machine logs would be very useful.
 	var numQueryErrors int
```
ejj: Perhaps I'm reading the code wrong, but this variable doesn't appear to be used at all (other than incremented).  Not the patch to fix it of course ...

kklin: It's used in the return statement

ejj:
FWIW, I think I would have liked to see the fuzzy lookup change be in it's own PR.  Seems worthwhile, but I don't see what it has to do with the Kube migration.

kklin:
I assume you meant the fuzzy lookup change to remove the argument for the public IP of the return thing. It was needed before because SSHing into a container required SSHing into the public IP of the worker, which differed according to the container. With Kube, we just always go through the leader, so that's why it's removed in this PR. I guess I could have left it unused in the Kube commit, and then removed it entirely in a later one.

ejj:
Ah fair enough.

minion/kubernetes/affinity.go
```diff
@@ -0,0 +1,190 @@
...
+		if !ok {
+			affinity = &corev1.Affinity{}
+			targetToAffinity[plcm.TargetContainer] = affinity
+		}
+
+		if plcm.OtherContainer != "" {
+			handlePodAffinity(affinity, plcm.OtherContainer, plcm.Exclusive)
+		}
+
+		nodeConstraints := []struct{ key, val string }{
```
ejj: Why not use a `map[string]string`?

minion/kubernetes/affinity.go
```diff
@@ -0,0 +1,190 @@
...
+		var needsUpdate bool
+		for key, exp := range labels {
+			actual, ok := node.Labels[key]
+			if !ok || exp != actual {
+				node.Labels[key] = exp
+				needsUpdate = true
+			}
+		}
+
+		if needsUpdate {
```
ejj: I think I would `if !needsUpdate { continue }`.  Doesn't matter much though

minion/kubernetes/configmap.go
```diff
@@ -0,0 +1,122 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+// updateConfigMaps creates a config map for each container's
+// FilepathToContent. It ignores any Secret ContainerValues.
+// These config maps are referenced by the updateDeployment function to mount
+// the container's filepathToContent into the container.
+func updateConfigMaps(conn db.Conn, configMapsClient clientv1.ConfigMapInterface) (
+	noErrors bool) {
```
ejj: This seems odd, why not just return an error?

minion/kubernetes/configmap.go
```diff
@@ -0,0 +1,122 @@
+package kubernetes
```
ejj: I think this file would benefit from an overview comment.

minion/kubernetes/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+		if err != nil {
+			log.WithError(err).WithField("deployment", deployment.Name).
+				Error("Failed to update deployment")
+		}
+	}
+
+	for _, intf := range toCreate {
+		deployment := intf.(appsv1.Deployment)
+		log.WithField("deployment", deployment.Name).
+			Info("Creating deployment")
```
ejj: FWIW, this fits on the above line.

minion/kubernetes/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+		if _, err := deploymentsClient.Create(&deployment); err != nil {
+			log.WithError(err).WithField("deployment", deployment.Name).
+				Error("Failed to create deployment")
+		}
+	}
+
+	for _, intf := range toDelete {
+		deployment := intf.(appsv1.Deployment)
+		log.WithField("deployment", deployment.Name).
+			Info("Deleting deployment")
```
ejj: Same here

minion/kubernetes/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+	for _, dbc := range containers {
+		deployment, ok := builder.makeDeploymentForContainer(dbc)
+		if ok {
+			deployments = append(deployments, deployment)
+		}
+	}
+	return deployments
+}
+
+type deploymentBuilder struct {
```
ejj: Looking at this, I don't think it's buying us very much.  I think I'd be inclined just to use a function rather than having this config object with methods.  Passing the config isn't so big a deal.

minion/kubernetes/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+			Dockerfile: img.Dockerfile,
+		}] = img
+	}
+
+	return deploymentBuilder{imageMap, toAffinities(placements), secretClient}
+}
+
+func (builder deploymentBuilder) makeDeploymentForContainer(dbc db.Container) (
+	appsv1.Deployment, bool) {
+	pod, ok := builder.makePodForContainer(dbc)
```
ejj: Btw, one thing that you could do to make it a bit easier to unit test, is pass the pod into makeDeploymentForContainer rather than generating it within.  Not sure if it will clean things up much or not, just a thought.

minion/kubernetes/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+		imageMap[db.Image{
+			Name:       img.Name,
+			Dockerfile: img.Dockerfile,
+		}] = img
+	}
+
+	return deploymentBuilder{imageMap, toAffinities(placements), secretClient}
+}
+
+func (builder deploymentBuilder) makeDeploymentForContainer(dbc db.Container) (
```
ejj: Also, I think `makeDeployment` and `makePod` are fine names.  The ForContainer isn't really giving any additional information.  Particular since it takes a container as an argument.

minion/kubernetes/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+					},
+					Key: subpath,
+				},
+			},
+		})
+	}
+	return envVars
+}
+
+func toPodVolumes(filepathToContent map[string]blueprint.ContainerValue) (
```
ejj: I might call these `toPod` functions `makePod` instead.  Would be more consistent with the `makePod()` function above.

minion/kubernetes/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+			Name:      filesVolumeName,
+			ReadOnly:  true,
+			MountPath: path,
+			SubPath:   configMapKey(path),
+		})
+	}
+	return
+}
+
+func hashStr(toHash string) string {
```
ejj: Btw, how securely does kubernetes treat environment variables?  Leaking sha1s of secrets isn't ideal ...  Not sure what to do about it, but thought I'd mention it for the record.

minion/kubernetes/status.go
```diff
@@ -0,0 +1,176 @@
...
+		EnvHash               string
+		FilepathToContentHash string
+		DockerfileHash        string
+	}
+	dbcKey := func(intf interface{}) interface{} {
+		dbc := intf.(db.Container)
+		return joinKey{
+			Hostname: dbc.Hostname,
+			Image:    dbc.Image,
+			Command:  fmt.Sprintf("%+v", dbc.Command),
```
ejj: It's a bit odd that this one uses `%+v` and the one below simply uses `%v`.  I don't think it should matter, but was that intentional?

minion/network/openflow/run.go
```diff
@@ -0,0 +1,95 @@
+package openflow
+
+import (
+	"github.com/kelda/kelda/blueprint"
+	"github.com/kelda/kelda/db"
+	"github.com/kelda/kelda/minion/ipdef"
+	log "github.com/sirupsen/logrus"
+)
+
+// Run occaisionally updates the OpenFlow tables based on the containers in the database.
```
ejj: occasionally is misspelled

kklin: Haha, that's actually from your PR. But I'll fix

minion/kubernetes/affinity.go
kklin: The order needs to be consistent. I'll add a comment

minion/kubernetes/configmap.go
kklin: It's to avoid the complexity of aggregating multiple errors. The caller wouldn't make use of the error anyways, it would just check whether it was `== nil`. What do you think?

kklin:
@ejj I addressed your comments, and implemented the volume/privileged container stuff. The commits with new code are:
- minion: Use blueprint row in database 
- kube: Squash me
- vendor: Remove unused dependencies

ejj:
@kklin Looked at the new patches, all good.  Merge when ready.

PR #1403, opened 2018-02-27
engine: Don't modify placement rules if blueprint is reprocessed

kklin:
Before, if the leader reprocessed the same blueprint (as happens when
the daemon restarts) the engine would sometimes modify the placement
rules in the database. This was because the `portPlacements` function
made use of the fact that the order of TargetContainer and
OtherContainer doesn't matter:  if there was an exclusive placement, it
would create one rule, using the first referenced container in the given
connections as the TargetContainer. Therefore, if the order of
connections passed to `portPlacements` changed, the resulting placement
rules might have swapped TargetContainers and OtherContainers. Note that
it's expected for the order to change since the connections are read
from the database, which returns rows in a non-deterministic order.

This commit addresses this problem by creating two placement rules for
each exclusive placement:  one rule for each affected container, with
it as the TargetContainer.

Although this placement swapping didn't result in incorrect behavior in
the current scheduler implementation, it will be problematic in the
Kubernetes implementation. This is because the placement information is
part of the deployment spec we pass to Kubernetes, so if the placement
rules move from one container to another, their deployment specs will
change, and Kubernetes will restart the changed deployments.

Furthermore, defining placement rules for both affected containers will
make the Kubernetes deployment specs easier to debug. Looking at the
placement rules for a given deployment will be sufficient to understand
its placement constraints:  there's no possibility that another
deployment has a placement rule that references the container.


ejj:
Bleh.  I'll be really glad when we can get rid of our funky placement code and just use Kube's.

minion/engine.go
```diff
@@ -45,8 +45,11 @@ func updatePolicy(view db.Database, bp string) {
 	updatePlacements(view, compiled)
 }
 
-// `portPlacements` creates exclusive placement rules such that no two containers
-// listening on the same public port get placed on the same machine.
+// `portPlacements` creates exclusive placement rules such that no two
+// containers listening on the same public port get placed on the same machine.
+// It produces the same placement rules regardless of the order of connections.
```
luise: I'd say more explicitly that the important part is that we create two placement rules per placement. I could imagine reading this comment in the future and not realizing immediately that's the important part.

PR #1404, opened 2018-03-01
Support privileged containers and hostPath volumes

kklin:
Closes https://github.com/kelda/kelda/issues/1402


js/bindings/bindingsTest.js
```diff
@@ -641,6 +641,7 @@ describe('Bindings', () => {
     const hostname = 'host';
     const image = new b.Image({ name: 'image' });
     const command = ['arg1', 'arg2'];
+    const privileged = true;
```
luise: Maybe also add a test that checks that this is false if not set? Just because it would be really bad if it isn't... :)

kklin: Done!

integration-tester/tests/volume/volume.js
```diff
@@ -0,0 +1,31 @@
...
+  image: 'ubuntu',
+  command: ['tail', '-f', '/dev/null'],
+};
+
+const hasMountArgs = Object.assign({}, noMountArgs, {
+  name: 'has-mount',
+  volumeMounts: [
+    new kelda.VolumeMount({
+      volume,
+      mountPath: '/docker.sock',
```
kklin: I wonder if this should just be called "path". Kubernetes calls it `mountPath`, but the inconsistency with the `kelda.Volume` signature might be confusing.

The downside to calling it `path` is that Kubernetes has a `subPath` field that we might eventually support (it lets you selectively pick what you want to mount out of the Volume).

js/bindings/bindings.js
```diff
@@ -1047,6 +1047,9 @@ class Container {
    *   Docker Hub) that the container should boot.
    * @param {string[]} [args.command] - The command to use when starting
    *   the container.
+   * @param {bool} [args.privileged] - Whether the container should be run in
```
ejj: I might explicitly say "root" somewhere in this paragraph.  Most people know what that means and it would be a bit more clear.

I also think that it would be worth adding a sentence that warns that you really shouldn't be using this unless you know what you're doing.

integration-tester/tests/volume/volume.js
ejj: I'd say leave it.  Though I'm not sure why I feel that way =)

integration-tester/tests/fuse/fuse.js
```diff
@@ -0,0 +1,93 @@
...
+// The SSH key used to authenticate the SSH connection. Although it's ugly to
+// save them here in plaintext, it's safe to do so as all traffic is in the
+// private network, and the SSH server isn't hosting anything of importance.
+const sshPubKey = 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCsxTbw33l7enItiAptl' +
+  'OkIj+N6Mg1HZxXXrX/V/PcODNEeVvDxokhfM1vJnPhiaYu+qLIvDJeixo9SO0xIbKcJ9p74af' +
+  'GsfEShfw4MPCSaaJk/gni/tqrNZxBf7AgTkO/96uPToALeXkAFVw7AtwlV16RqCB7+kUMiW62' +
+  'o9hmLCbjUuI7myJCMjCffQaLC7Oz7cCphxEG1lNb4E7dz/6CEJQF+84EuRfLjuq6sJchIHs2j' +
+  'NriV/vL6PXB9a8DAzTCvzrHlf3A4gESn2CkEScBq+QaC3EN4sM2C8LqlQAfFFSi5N5siHmDBb' +
+  'zHYlPDLnLj2iL0IXveTno5RjbjZbYd1';
+const sshPrivKey = `-----BEGIN RSA PRIVATE KEY-----
```
ejj: I wonder if github is going to freak out that we're pushing a private key ..... will be interesting to find out =)

PR #1408, opened 2018-03-03
digitalocean: Cache List() calls

ejj:
Digital Ocean has a rather aggressive API request limit that we're
hitting in integration tests.  Part of the problem, is that the API is
global, serving all regions, rather than segregated per region as it
is in other cloud providers.  This means that instead of making a
single List() call that could request all droplets, we were making a
list call per region, artificially inflating our requests per second.
This patch solves the issue by caching the result of a List() call for
1 second, allowing it to be reused by other regions without making API
requests.

### DIFF
--- a/cloud/digitalocean/digitalocean.go
+++ b/cloud/digitalocean/digitalocean.go
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/digitalocean/godo"
 
@@ -79,7 +81,7 @@ func New(namespace, region string) (*Provider, error) {
 		return prvdr, err
 	}
 
-	_, _, err = prvdr.ListDroplets(&godo.ListOptions{}, prvdr.getTag())
+	_, _, err = prvdr.ListDroplets(&godo.ListOptions{})
 	return prvdr, err
 }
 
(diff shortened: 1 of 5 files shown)


cloud/digitalocean/digitalocean.go
```diff
@@ -6,7 +6,10 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+
```
kklin: Why this newline?

cloud/digitalocean/digitalocean.go
```diff
@@ -104,8 +107,47 @@ var newDigitalOcean = func(namespace, region string) (*Provider, error) {
...
+var listAllMutex sync.Mutex
+var listAllTimeout time.Time
+var listAllMachines []taggedMachine
+var listAllErr error
+
 // List will fetch all droplets that have the same name as the cluster namespace.
 func (prvdr Provider) List() (machines []db.Machine, err error) {
+	listAllMutex.Lock()
+	if time.Now().After(listAllTimeout) {
+		c.Inc("List Cache Hit")
```
kklin: The counter should be for a cache miss

cloud/digitalocean/digitalocean_test.go
```diff
@@ -54,26 +54,32 @@ func init() {
...
 
 	dropLast := []godo.Droplet{
 		{
 			ID:        125,
 			Networks:  network,
 			SizeSlug:  "size",
 			VolumeIDs: []string{"foo"},
-			Region:    godoRegion}}
+			Region:    godoRegion,
+			Tags:      []string{tag},
```
kklin: Can you test that the machines in different namespaces aren't returned?

Also, should you test the cache hit case?

PR #1409, opened 2018-03-04
aws: add us-east-1 as a supported region

cure:


kklin:
Thanks for the PR @cure!

It's unfortunate, but we depend on a kernel module for our networking tunnels, so the AMI image needs to have a specific kernel version. (If the kernel version is wrong then we compile the module at runtime, which is slow).

I found an AMI for the right kernel version, could you try that? ami-f0768de6

Also, a super small nit, but by convention we capitalize the first letter after the colon in our commit titles (so it should be `aws: Add us-east-1 ...`). Could you tweak that when you update the image?

cure:
Hi @kklin - no problem; I've tested with the AMI you provided and it worked fine. I've also updated the commit message as you suggested.

kklin:
Thanks!

PR #1410, opened 2018-03-07
docs: Document instance properties

luise:
This commit documents class properties, so they will appear in the docs.
To avoid confusion, we only document properties users should have to care about.

Example of a propteries table in the docs:
screen shot 2018-03-06 at 4 24 16 pm


aegamesi:
Is getConnectableName intentionally left public for the Connectable interface? ( http://docs.kelda.io/#Connectable ) -- I mean, that's the entire point of the interface, just want to make sure that's intentional

js/bindings/bindings.js
```diff
@@ -1099,9 +1118,23 @@ class Container {
...
+   * @param {VolumeMount[]} [args.volumeMounts] - A list of volumes to mount
    *   within the container. Referenced volumes are automatically created by
    *   Kelda.
+   *
+   * We only document properties users should care about.
+   * @property {Image} image The image of the container.
+   * @property {string[]} command The command to run when the container starts.
+   * @property {Object.<string, string|Secret>} env An object containing the
+   *   environment variables to set in the container. The key is the name of the
+   *   variable and value is the variable's value.
```
aegamesi: would sound better as 'and *the* value is the variable's value'.

js/bindings/bindings.js
```diff
@@ -99,6 +100,18 @@ class Infrastructure {
...
    *   to allow access from 1.2.3.4, set adminACL to ["1.2.3.4/32"]. To allow access
    *   from all IP addresses, set adminACL to ["0.0.0.0/0"].
+   *
+   * We only document properties users should care about.
+   * @property {Container[]} containers All containers that have been registered
+   *   to run on this infrastructure.
+   * @property {LoadBalancer[]} loadBalancers All load balancers that have been
+   *   registered to run on this infrastructure.
+   * @property {Machine[]} masters The Master machines of this infrastructure.
+   * @property {Machine[]} workers The Worker machines of this infrastructure.
```
aegamesi: are 'Worker' and 'Master' intentionally capitalized?

luise: I guess I did it because we always capitalize the actual role in the JS code -- but there is no reason to do it here. I'll fix it!

luise:
@aegamesi , yeah, I left `getConnectableName()` of the interface public because it still seems relevant there.

luise:
@aegamesi , I addressed your comments.

PR #1411, opened 2018-03-08
cloud: Prefer to leave a floating IP in place when possible

ejj:
Before this patch, the cloud preferred to join machines with known
roles even if that meant reassigning a floating IP away from a machine
with an unknown role to a different worker with a known role.  This
meant that on boot (or namespace change), a perfectly health machine
could have its floating IP taken away from it, simply because one of
its peers declared its role to the daemon before it got a chance to.

There's really no particular reason for this preference, so this patch
fixes the issue by reversing it.  All things being equal, we will
prefer to match a machine with the correct floating IP, even if that
means its role is unknown.

### DIFF
--- a/cloud/join_test.go
+++ b/cloud/join_test.go
@@ -233,7 +233,7 @@ func TestMachineScore(t *testing.T) {
 	// Wrong ID, but an assigned role.
 	m1 = m
 	m1.CloudID = "5"
-	assert.Equal(t, 8, machineScore(m, m2))
+	assert.Equal(t, 9, machineScore(m, m2))
 
 	// Wrong ID, but no Role.
 	m1 = m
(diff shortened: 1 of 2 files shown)


cloud/join_test.go
```diff
@@ -287,7 +287,7 @@ func TestMachineScore(t *testing.T) {
 	workerWithBadIP.Role = db.Worker
 	workerWithBadIP.FloatingIP = "wrong"
 
-	assert.True(t, machineScore(desired, workerWithBadIP) <
+	assert.True(t, machineScore(desired, workerWithBadIP) >
```
kklin: Isn't it bad that it's possible to match a worker with a master? I think it can have a cascading effect where pairing one worker with a master forces a blueprint master to try to be paired with a worker (which will fail). Then, that worker will be destroyed, and another master will be booted.

ejj: Let's talk about this in person, but I'm not sure this is a real problem.  The code will only attach a worker to a machine with an assigned floating IP, if that machine's role is None.  In this case we've got a machine in the cloud that has at some previous time had a floating Ip assigned to it, which almost certainly means its a worker as we don't have a mechanism to assign floating IPs to masters.

At any rate, the issue you're suggesting could have happened in the previous code.  We're just slightly changing the preferences here so that previously assigned floating IPs take precedence over being absolutely certain about the role assignment.   In the case where you've got a bunch of unknown roles floating around, it's still possible to have a certain amount of ambiguity.  I think the only way to solve that would be to wait to do the join until we've figured out everybody's role.  That may be the right thing to do, but not sure it's worth it yet.

ejj:
TODO add a comment

PR #1413, opened 2018-03-09
README: Add logo

luise:


luise:
That's a fair point. I'll play around with it

luise:
Couldn't come up with anything better, so I'll just stick with this, and we can always change it later.

kklin:
Sounds good!

PR #1422, opened 2018-03-15
connection: Improve error message when unix socket already exists

kklin:
It's somewhat common for users to accidentally run two daemons on the
same machine. Because this error is especially common with
first-time users, this commit updates the error message to have
instructions for fixing the error.

Closes https://github.com/kelda/kelda/issues/1416


connection/connection.go
```diff
@@ -38,6 +39,22 @@ func Client(proto, addr string, opts []grpc.DialOption) (*grpc.ClientConn, error
 // to open the socket, it tries again until it succeeds.
 func Server(proto, addr string, opts []grpc.ServerOption) (net.Listener, *grpc.Server) {
 	for {
+		if proto == "unix" {
+			// Check if the address already exists so that we can give a more
+			// helpful error message.
+			if _, err := os.Stat(addr); err == nil {
+				msg := "The socket path already exists. Another daemon is " +
+					"probably already running, in which case it can be stopped " +
+					"with the `kill` command. If the socket is leftover " +
```
kklin: I wasn't sure if there was a better way to say that the other daemon should be killed. The `kill` command might not be familiar to some users.

The following command should also work to find the process and kill it, but it looks a bit ugly:
```
lsof -F p /tmp/kelda.sock | grep p | tr -d p | xargs kill
```

connection/connection.go
```diff
@@ -34,10 +35,25 @@ func Client(proto, addr string, opts []grpc.DialOption) (*grpc.ClientConn, error
 		grpc.WithBlock(), grpc.WithTimeout(timeout))...)
 }
 
+const duplicateDaemonMsg = "The socket path already exists. Another daemon is " +
+	"probably already running, in which case it can be stopped with the `kill` " +
```
luise: I think it's fine to refer to the `kill` command - it's simpler, and people can always google it. I think we should suggest a signal number (2 or whatever) as well, so we can clean up correctly (otherwise the problem won't be fixed). Might just say "... stopped with the command: `kill -2 <PID_of_daemon_process>`..."

connection/connection.go
```diff
@@ -34,10 +35,25 @@ func Client(proto, addr string, opts []grpc.DialOption) (*grpc.ClientConn, error
...
+	"command. If the socket is leftover from a previous daemon that didn't exit " +
+	"cleanly, the path can be manually deleted from the filesystem."
```
luise: I might be even more to-the-point here and for instance say "If no other daemon is running, try removing the socket manually with `rm /tmp/kelda.sock`".

luise:
This is going to be nice to have!

kklin:
@luise ready for more review!

PR #1426, opened 2018-03-19
supervisor: Properly restart containers when they exit

kklin:
Before, if a system container exited, the supervisor would attempt to start a
new one in its place, but the creation would fail because Docker doesn't
allow having two containers with the same name. This commit gets around
this by renaming the old container so that the new container can use the
friendly name.


minion/docker/docker_test.go
```diff
@@ -228,9 +212,9 @@ func TestRunEnv(t *testing.T) {
 		},
 	}
 
-	for _, env := range testEnvs {
+	for i, env := range testEnvs {
 		id, err := dk.Run(RunOptions{
-			Name: "name",
+			Name: fmt.Sprintf("name%d", i),
```
aegamesi: Why is this change a part of this patch?

minion/supervisor/supervisor.go
```diff
@@ -135,6 +136,39 @@ func joinContainers(desiredContainers []docker.RunOptions) {
...
+		containersWithName, err := dk.List(map[string][]string{
+			"name": {ro.Name},
+		}, true)
+		if err != nil {
+			log.WithError(err).WithField("name", ro.Name).Error(
+				"Failed to check for containers with same name")
+			continue
+		}
+
+		if len(containersWithName) != 0 {
```
aegamesi: It feels weird to have a list but only look at the first element. I mean, it makes sense if Docker doesn't allow multiple containers with the same name, so there could only ever be one element in this list, I suppose.

minion/supervisor/supervisor_test.go
```diff
@@ -53,11 +57,55 @@ type fakeDocker struct {
...
+		if strings.HasPrefix(c.Name, ro.Name) && !c.Running &&
+			c.ID == exitedContainerID {
+			foundStoppedContainer = true
+		}
+	}
+
+	assert.True(t, foundNewContainer,
+		"a new container should be booted with the container name")
+	assert.True(t, foundStoppedContainer,
+		"the stopped container should still exist, but be renamed")
```
aegamesi: The test seems to only implicitly check that the container has been renamed (by virtue of Docker not allowing multiple containers with the same name).

minion/docker/docker_test.go
kklin: Because they're all created with the same name, running the second container would throw an error now that I've modified the mock to error when there's two containers with the same name.

minion/supervisor/supervisor.go
kklin: Yeah, that was the logic. Do you have an idea for a different way to write it?

aegamesi: I guess looping over the containers with that name would seem more correct.

kklin: Sure, i'll do that

minion/supervisor/supervisor_test.go
kklin: I changed the test to check for the "_stopped" part too

PR #1430, opened 2018-03-21
Split up the `kelda init` command

luise:
This PR splits `kelda init` into two commands: `kelda configure-provider` and `kelda base-infrastructure`. The former helps set up the correct cloud provider credentials files and the latter will create a base infrastructure.


kklin:
Man, I thought this would be a quick review, didn't realize the diff was so big 😁

luise:
Haha, yes, it’s also bigger than I expected! It’s a lot of bits and pieces from the old code though, but the diff just doesn’t pick it up very well.

js/base-infrastructure/prompter.js
```diff
@@ -0,0 +1,265 @@
...
+      type: 'input',
+      name: consts.masterCount,
+      message: 'How many master VMs?',
+      default: 1,
+      validate(input) {
+        return isNumber(input);
+      },
+    },
+  ];
+  console.log('Master machines are in charge of keeping your application running. ' +
```
luise: Regarding our earlier conversation about whether it's safe to log before the asynchronous prompts, I think it's fine because the logs are inside the chain of `then()`s. So the next log and inquiry won't execute before the prior is resolved.

js/configure-provider/prompter.js
```diff
@@ -0,0 +1,135 @@
...
+  const keys = provider.getCredsKeys();
+  const keyNames = Object.keys(keys);
+
+  keyNames.forEach((keyName) => {
+    questions.push(
+      {
+        type: 'input',
+        name: keyName,
+        message: `${keys[keyName]}:`,
+        // Ask this questions for all credential inputs that are not paths.
```
kklin: questions -> question

js/install/packageTemplate.json
```diff
@@ -13,6 +13,7 @@
   "bin": {
     "kelda": "kelda",
     "kelda-initializer.js": "./initializer/initializer.js",
```
kklin: Can this be removed?

kklin:
There's still a reference to `kelda init` in an error message in `bindings.js`

kklin:
Oh, you should also update the `.gitignore` in the `js/install` directory

kklin:
The `js/install/makePackage.js` script is also out of date -- you can repro by running `make prep-install` (it's not done by travis until after the code is merged)

js/install/packageTemplate.json
luise: It can indeed!

luise:
@kklin , thanks for catching all the things around the installation! It's ready for another review!

luise:
Oh, I'll update the CHANGELOG

PR #1433, opened 2018-03-28
js/install: Fix path of the base infrastructure script

kklin:
`kelda base-infrastructure` shells out to `kelda-base-infra-init.js`,
but we installed the script as `kelda-base-infra.js` instead. This
caused the command to fail since it couldn't find the script.


luise:
Thanks for fixing this!

kklin:
No problem, I missed it in my review as well

